PR #23, opened 2023-10-31
Disabled docker correctly for no docker tests

OliverLok:
Disabled docker then restored the docker_disable config to the original value for when we test grading with no docker.


eriq-augustine:
lgtm!

PR #39, opened 2023-12-07
Added the remove submission endpoint

OliverLok:
Added a new API endpoint to remove submissions. Let me know if there are any changes I need to make.


db/disk/submission.go
```diff
@@ -294,3 +294,17 @@ func (this *backend) getMostRecentSubmissionID(assignment *model.Assignment, ema
...
     return dirents[len(dirents) - 1].Name(), nil;
 }
+
+func (this *backend) RemoveSubmission(assignment *model.Assignment, email string, shortSubmissionID string) error {
+    submissionDir := this.getSubmissionDirFromAssignment(assignment, email, shortSubmissionID);
+    if (!util.PathExists(submissionDir)) {
+        return fmt.Errorf("No submission found");
+    }
+
+    err := os.RemoveAll(submissionDir)
```
eriq-augustine, line 304: Prefer util.RemoveDirent().

db/disk/submission.go
```diff
@@ -294,3 +294,17 @@ func (this *backend) getMostRecentSubmissionID(assignment *model.Assignment, ema
...
+    if err != nil {
+        return fmt.Errorf("Error removing submission")
```
eriq-augustine, line 306: NEVER* ignore an error.

*99.9% of the time.

eriq-augustine (changes requested):
Pretty good start, most of the core is there.

api/submission/remove_submission.go
```diff
@@ -0,0 +1,36 @@
...
+    "github.com/eriq-augustine/autograder/api/core"
+    "github.com/eriq-augustine/autograder/db"
+)
+
+type RemoveSubmissionRequest struct {
+    core.APIRequestAssignmentContext
+    core.MinRoleGrader
+
+    TargetUser core.TargetUserSelfOrGrader `json:"target-email"`
+    TargetSubmission string `json:"target-submission"`
```
eriq-augustine, line 13: This typing (being a string that can be empty) indicates that if no ID is supplied the most recent submission is used.
I think that this is the correct functionality, but the rest of your code assumes that a submission ID is always provided.

OliverLok: Doesn't the autograder-py interface only send a request to the server if the targetsubmission is not empty, meaning if there was a request there will always be a submissionID since it got through that check in autograder-py?

eriq-augustine: What about when people don't use the autograder-py library to send requests?
Or they use the library, but override/bypass our code?

Regardless, I recommended changing the Python side as well.
This is also a really strange statement, since you are also making the Python side right now.
Even if the other side was not currently in development, we should always strive to do the right thing, even if it means changing standards.

api/submission/remove_submission.go
```diff
@@ -0,0 +1,40 @@
...
+    response := RemoveSubmissionResponse{};
+
+    if (!request.TargetUser.Found) {
+        return &response, nil;
+    }
+
+    response.FoundUser = true;
+
+    doesExist, err := db.RemoveSubmission(request.Assignment, request.TargetUser.Email, request.TargetSubmission);
+    if ( (err != nil) || (doesExist == false)){
```
eriq-augustine, line 31: Spacing is off here.
Extra space after first parent, missing space after last paren.

api/submission/remove_submission.go
```diff
@@ -0,0 +1,40 @@
...
+        response.FoundSubmission = false;
+        return nil, core.NewInternalError("-418", &request.APIRequestCourseUserContext, "Failed to remove the submission.").
```
eriq-augustine, line 33: NEVER* group error and non-error conditions.
If you look at endpoints like peek and even your own response structure,
it is not an error to try and remove a submission that doesn't exist (like peeking an error that doesn't exist).

We want errors from our server to be as rare as possible, this means that we only create API errors when there is an actual error that we cannot recover from.
An unknown submission is not an error we can't recover from, we just tell them it doesn't exist.

**IMPORTANT**
I may be getting a bit too deep in my analysis here,
but I think it is important that you slow down a bit and really think about this paragraph.
Every piece of code should be intentional and make as much sense as possible.
Here you invoke an "InternalError" when a submission was not found.
But the submission ID is provided by the user, so the cause of that error cannot be internal, it's external (the user sent it to us).
Therefore, an "internal" error does not make sense.
I want to make sure you are thinking in ways like this when working with the codebase.

*99.9% of the time.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,71 @@
...
+                    test.Errorf("Case %d: Incorrect error returned on permissions error. Expcted '%s', found '%s'.",
+                            i, expectedLocator, response.Locator);
+                }
+            } else {
+                test.Errorf("Case %d: Response is not a success when it should be: '%v'.", i, response);
+            }
+            
+            continue;
+        }
+
```
eriq-augustine, line 68: Extra newlines here and later.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,71 @@
...
+        // Grader, self, recent.
+        {model.RoleGrader, "grader@test.com", "", true, false, false},
+        {model.RoleGrader, "grader@test.com", "", true, false, false},
+
+        // Grader, self, missing.
+        {model.RoleGrader, "",                "ZZZ", true, false, false},
+        {model.RoleGrader, "grader@test.com", "ZZZ", true, false, false},
+
+        // Grader, other, recent.
+        {model.RoleGrader, "student@test.com", "",true, true, false},
```
eriq-augustine, line 25: Missing space.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+        // Grader, other, specific.
+        {model.RoleGrader, "student@test.com", "1697406256", true, true, false, },
```
eriq-augustine, line 28: Extra space (and comma).

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,71 @@
...
+        {model.RoleGrader, "student@test.com", "1697406265", true, true, false},
+        {model.RoleGrader, "student@test.com", "1697406272", true, true, false},
+
+        // Grader, other, specific (full ID).
+        {model.RoleGrader, "student@test.com", "course101::hw0::student@test.com::1697406256", true, true, false},
+        {model.RoleGrader, "student@test.com", "course101::hw0::student@test.com::1697406265", true, true, false},
+        {model.RoleGrader, "student@test.com", "course101::hw0::student@test.com::1697406272", true, true, false},
+
+        // Grader, other, missing.
+        {model.RoleGrader, "student@test.com", "ZZZ", true, false, true},
```
eriq-augustine, line 38: This is not a perm error.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+        // Grader, missing, recent.
+        {model.RoleGrader, "ZZZ@test.com", "", false, false, true},
+    };
```
eriq-augustine, line 42: Add in a set of cases that sweep through the other roles (users),
making sure all below grader get a perm (permissions) error.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,71 @@
...
+                if (response.Locator != expectedLocator) {
+                    test.Errorf("Case %d: Incorrect error returned on permissions error. Expcted '%s', found '%s'.",
+                            i, expectedLocator, response.Locator);
+                }
+            } else {
+                test.Errorf("Case %d: Response is not a success when it should be: '%v'.", i, response);
+            }
+            
+            continue;
+        }
```
eriq-augustine, line 67: You don't check the response (foundUser and foundSubmission).

db/database.go
```diff
@@ -71,6 +71,10 @@ type Backend interface {
     // Remove a user.
     // Do nothing and return nil if the user does not exist.
     RemoveUser(course *model.Course, email string) error;
+    
+    // Remove a submission.
+    // Return a bool indicating whether the submission exists or not and an error if there is one
```
eriq-augustine, line 76: Missing period.

db/disk/submission.go
```diff
@@ -214,7 +214,7 @@ func (this *backend) GetRecentSubmissionSurvey(assignment *model.Assignment, fil
 
 func (this *backend) GetSubmissionContents(assignment *model.Assignment, email string, shortSubmissionID string) (*model.GradingResult, error) {
     var err error;
-
+
```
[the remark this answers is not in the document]
eriq-augustine: (Unresolving comment.)

This has not been resolved.
Be really careful to miss something that you say is resolved (reviews hate that).

db/disk/submission.go
```diff
@@ -294,3 +294,30 @@ func (this *backend) getMostRecentSubmissionID(assignment *model.Assignment, ema
 
     return dirents[len(dirents) - 1].Name(), nil;
 }
+
+func (this *backend) RemoveSubmission(assignment *model.Assignment, email string, shortSubmissionID string) (bool, error) {
+    var err error;
+
+    if (shortSubmissionID == "") {
+        shortSubmissionID, err = this.getMostRecentSubmissionID(assignment, email);
+        if (err != nil){
```
eriq-augustine, line 303: Missing space.

(There are more of these, but I will stop making spacing comments. (You still have to fix them.))

db/disk/submission.go
```diff
@@ -294,3 +294,30 @@ func (this *backend) getMostRecentSubmissionID(assignment *model.Assignment, ema
...
+
+    if (shortSubmissionID == ""){
+        return false, nil;
+    }
+
+    submissionDir := this.getSubmissionDirFromAssignment(assignment, email, shortSubmissionID);
+
+    if (!util.PathExists(submissionDir)) {
+        return false, fmt.Errorf("Submission does not exist.");
+    }
```
eriq-augustine, line 316: Missing a newline after this block.

If this was intentional, then you will want to make sure to use vertical whitespace to separate blocks into chunks of code that does one discrete thing.

db/disk/submission.go
```diff
@@ -294,3 +294,30 @@ func (this *backend) getMostRecentSubmissionID(assignment *model.Assignment, ema
...
+
+    if (shortSubmissionID == ""){
+        return false, nil;
+    }
+
+    submissionDir := this.getSubmissionDirFromAssignment(assignment, email, shortSubmissionID);
+
+    if (!util.PathExists(submissionDir)) {
+        return false, fmt.Errorf("Submission does not exist.");
```
eriq-augustine, line 315: This is not an error condition.
The whole reason to have the first return value (the boolean) is so that we can tell when a submission actually existed.
So there is no reason to also create an error in this case.

Also, think about this from the "recent submission" case (where the submission ID is empty).
It seems mean to punish someone with an error for removing an ID that they didn't pass in.
(They could have tried to get the submission first, but it still seems like we are putting a big onus on the caller.)

eriq-augustine (changes requested):
Your CI is failing.
In the future, make sure that CI passes before the next PR round.
Remember, you want to do whatever you can to make it easier for a reviewer.

My big advice for this round would be to slow down and go over tings again.
A lot of your changes were great, but there were also a lot of mistakes, e.g.:
 - CI failing.
 - Incomplete tests (e.g. missing checks for foundUser and foundSubmission that you already setup for and just didn't do).
 - Several spacing issues.

You are also missing the DB test we talked about.
It should be pretty simple (even simpler that your API test, since you don't have to worry about users or permissions).
Just a set of test cases like in your other test (recent, specific, and missing)
and ensure that if the DB says it was removed, it was removed.

I also want to reiterate (since it may seem like there are a lot of negative),
the core work here is good and you are not too far.
But you probably need to be a bit more meticulous.

db/disk/submission.go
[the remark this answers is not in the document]
OliverLok: Sorry, I mistook stray spacing for an extra line. Will not happen again.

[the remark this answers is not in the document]
eriq-augustine: np.
Make sure to read PR comments very matter-of-factly (all denotation, no connotation),
as they are generally written as just stating facts and not emotionally.
This applies to pretty much all PRs, not just mine.

(I just want to make sure, since it can be easy to accidentally read them and pull out emotion that is not there. And that can make PRs really traumatizing.)

Also, now you are in a peculiar spot where you commented on a thread, but also need to resolve this thread
(because the review will have nothing to do with it).
In general you want to avoid comments like this, and just resolve it.
If you really need to apologize for something (which is rare in code), it's better to do it in person, via some other communication, or in the general comments for the PR (on the main PR page).

If you do need to make small comments like this, then make the comment but still resolve it so the reviewer doesn't see it as an action item.
(Many reviewers will still look at all the resolved comments from last round to page in the PR.)

api/submission/remove_submission.go
```diff
@@ -0,0 +1,44 @@
...
+func HandleRemoveSubmission(request *RemoveSubmissionRequest) (*RemoveSubmissionResponse, *core.APIError){
+    response := RemoveSubmissionResponse{};
+
+    if (!request.TargetUser.Found) {
+        return &response, nil;
+    }
+
+    response.FoundUser = true;
+
+    doesExist, err := db.RemoveSubmission(request.Assignment, request.TargetUser.Email, request.TargetSubmission);
```
eriq-augustine, line 30: Always* check errors first.
The value for doesExist is a garbage value (semantically) if (err != nil).

*99% of the time.

api/submission/remove_submission.go
```diff
@@ -0,0 +1,44 @@
...
+    
+    if (doesExist) {
+        response.FoundSubmission = true;
+    } else { 
+        response.FoundSubmission = false;
+    }
```
eriq-augustine, line 36: Just a nit, but you could just do:
```
response.FoundSubmission = doesExist;
```

If would make sense to use an `if` if there were other things you need to do for each case.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,102 @@
+package submission
+
+import (
+    "testing"
+
+    "github.com/eriq-augustine/autograder/api/core"
+    "github.com/eriq-augustine/autograder/db"
+    "github.com/eriq-augustine/autograder/model"
+    "github.com/eriq-augustine/autograder/util"
+
```
eriq-augustine, line 10: Extra newline.

api/submission/remove_submission_test.go
eriq-augustine: (Unresolving)

This has not been resolved.
Look at a test like api/user/remove:
https://github.com/eriq-augustine/autograder-server/blob/main/api/user/remove_test.go#L23

All (but one) of the student tests you added are redundant, since they will all perm error.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,102 @@
...
+            test.Errorf("Case %d: Found user does not match. Expected: '%v', actual: '%v'.", i, testCase.foundUser, responseContent.FoundUser);
+            continue;
+        }
+
+        if (testCase.foundSubmission != responseContent.FoundSubmission) {
+            test.Errorf("Case %d: Found submission does not match. Expected: '%v', actual: '%v'.", i, testCase.foundSubmission, responseContent.FoundSubmission);
+            continue;
+        }
+
+        if (!responseContent.FoundSubmission) {
```
eriq-augustine, line 98: This last one is not necessary.

db/disk/submission.go
```diff
@@ -294,3 +294,31 @@ func (this *backend) getMostRecentSubmissionID(assignment *model.Assignment, ema
...
+
+    submissionDir := this.getSubmissionDirFromAssignment(assignment, email, shortSubmissionID);
+
+    if (!util.PathExists(submissionDir)) {
+        return false, nil;
+    }
+
+    err = util.RemoveDirent(submissionDir);
+    if (err != nil) {
+        return false, err;
```
eriq-augustine, line 320: Wrap this error and include the submission ID (short), e.g.,
"Failed to remove submission ('%s'): '%w'.".

db/submission_test.go
```diff
@@ -0,0 +1,58 @@
+package db
+
+import (
+    "testing"
+)
+
+func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
+    defer ResetForTesting();
+
+    testCases := []struct{targetEmail string; targetSubmission string; isRemoved bool}{
```
eriq-augustine, line 10: Missing space.

db/submission_test.go
```diff
@@ -0,0 +1,58 @@
...
+
```
eriq-augustine, line 11: Extra newline.

db/submission_test.go
```diff
@@ -0,0 +1,58 @@
...
+        // Student, specific.
```
eriq-augustine, line 12: In these errors, there is no longer the concept of "Student".
The DB does not care about roles, so you should clean up the comments a bit to not mention them,

db/submission_test.go
```diff
@@ -0,0 +1,58 @@
...
+        {"ZZZ@test.com", "1697406256", false},
+        {"ZZZ@test.com", "1697406265", false},
+        {"ZZZ@test.com", "1697406272", false},
+    };
+
+    for i, testCase := range testCases {
+        // Reload the test course every time.
+        ResetForTesting();
+        
+        assignment := MustGetAssignment(TEST_COURSE_ID, "hw0");
```
eriq-augustine, line 44: You can also use `MustGetTestAssignment()`.

db/submission_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+        isRemoved, err := RemoveSubmission(assignment, testCase.targetEmail, testCase.targetSubmission);
+
+        if (err != nil) {
+            test.Errorf("Case %d: Submission removal failed: '%v", i, err);
```
eriq-augustine, line 49: Missing period at the end of the string.

eriq-augustine (changes requested):
Looking pretty good, mainly small things.

api/submission/remove_submission_test.go
OliverLok: If the grader role can pass all tests, can we assume admin and owner will too or does that also require a set of test cases, and if so, how thorough should the tests be?

eriq-augustine: Your assumption is not wrong, but why rely on an assumption when it is trivial to test?
(Only five copy-paste lines.)

This is actually a really important lesson to learn about when to just trust assumptions and when to test things.
Like I mentioned above, in this case the cost of testing is so low compared to the safety it buys is that there is no reason not to do it.

OliverLok: Ok so as a follow up, there's 13 grader-role tests that test all scenarios. To be extra safe, wouldn't it make sense to use the same 13 test cases for Admin and Owner then? Also I'm a bit confused on the unknown role as it doesn't return a bad permission error (-306), but it is lower than a grader.

eriq-augustine: No, not necessary.
There is no functionality difference between any of the allowed roles (it's the same for an admin and owner).
So, we just need to make sure that those roles are not denied.
We typically choose to run tests at the lowest allowed permission level, but aside from that the choice of primary user is not too important.

Regarding the unknown role.
No user should have this role, so it will probably be caught even earlier in the system than a permissions error.
If no error is returned, then it is a problem.
An unknown role is like a nil role.

db/submission_test.go
OliverLok: I don't think that function is implemented anywhere unless I'm looking in the wrong spot for it.

eriq-augustine (changes requested):
Just some small things.

api/submission/remove_submission_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+        fields := map[string]any{
+            "target-email": testCase.targetEmail,
+            "target-submission": testCase.targetSubmission,
+        };
+
+        response := core.SendTestAPIRequestFull(test, core.NewEndpoint(`submission/remove/submission`), fields, nil, testCase.role);
+        
+        if (!response.Success) {
+            if (testCase.role == model.RoleUnknown) {
```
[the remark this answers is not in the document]
OliverLok: Ok, I guess I'm just wondering when to draw the line between when to add tests to be extra safe vs when tests are not necessary.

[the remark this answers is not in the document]
eriq-augustine: That can be a hard thing to figure out, and it can depend on the organization, project, and timeline.

But in this case it is a little different.
Since RoleUnknown is different than all the other roles, it does not need to be tested like the other roles.
It is handled at a much earlier stage (in the API's core), so anyone further down (like an endpoint) doesn't need to worry about it.

eriq-augustine (approved):
LGTM!

PR #51, opened 2024-01-17
Fetch Submission History API Endpoint

cfbeard:
Added an API endpoint to fetch all submissions to an assignment for a specific user.  Role is limited to grader and above.  

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206342160001609


api/submission/fetch_submission_history.go
```diff
@@ -0,0 +1,49 @@
...
+    "github.com/eriq-augustine/autograder/api/core"
+    "github.com/eriq-augustine/autograder/db"
+    "github.com/eriq-augustine/autograder/model"
+)
+
+type FetchSubmissionHistoryRequest struct {
+    core.APIRequestAssignmentContext
+    core.MinRoleGrader
+
+    TargetUser       core.TargetUserSelfOrGrader `json:"target-email"`
```
eriq-augustine, line 13: Strange spacing.

api/submission/fetch_submission_history.go
```diff
@@ -0,0 +1,49 @@
...
+
+    if !request.TargetUser.Found {
+        return &response, nil;
+    }
+
+    response.FoundUser = true;
+
+    gradingResults, err := db.GetSubmissionResultHistory(request.Assignment, request.TargetUser.Email);
+    if err != nil {
+        return nil, core.NewInternalError("-604", &request.APIRequestCourseUserContext, "Failed to get submissions.").
```
eriq-augustine, line 37: Wrong locator.

Be real careful here, either this was just a slip (which happens), or you didn't fully understand how the locators worked and guessed.
For something like this it can be hard to catch if missed, so make sure to ask.
Every locator should be unique so we can trace exactly where an error came from (e.g. a student has an error, and we only need this number to figure out what happened).
Each package gets allocated 100 numbers, choose the next available one for this package.

api/submission/fetch_submission_history.go
```diff
@@ -0,0 +1,49 @@
...
+
+    if gradingResults == nil {
+        return &response, nil;
+    }
+
+    response.FoundSubmissions = true;
+    response.GradingResults = gradingResults;
+
+    return &response, nil;
+}
```
eriq-augustine, line 49: Looks like some files are missing newlines at the end.
I would check your editor settings.

api/submission/fetch_submission_history_test.go
```diff
@@ -0,0 +1,122 @@
+package submission
+
+import (
```
eriq-augustine, line 3: Import groupings are off.
It should be:
 - stdlib
 - third party
 - this project

api/submission/fetch_submission_history_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+        if (!reflect.DeepEqual(testCase.result, responseContent.GradingResults)) {
+            test.Errorf("Case %d: Unexpected submission result. Expected: '%s', actual: '%s'.", i,
+                util.MustToJSONIndent(testCase.result), util.MustToJSONIndent(responseContent.GradingResults));
+            continue;
+        }
+    }
+}
+
+func getTestSubmissionsResultPath(shortID string) string {
```
eriq-augustine, line 120: Just use the existing function from fetch_submission_test.go anything in the package can use unexported functions.

api/submission/routes.go
```diff
@@ -12,10 +12,11 @@ var routes []*core.Route = []*core.Route{
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/scores`), HandleFetchScores),
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/submission`), HandleFetchSubmission),
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/submissions`), HandleFetchSubmissions),
+    core.NewAPIRoute(core.NewEndpoint(`submission/fetch/history`), HandleFetchSubmissionHistory),
```
eriq-augustine, line 15: I don't like this endpoint, too close to submission/history.

Let's think of an alternate name to replace "history". results? records? versions? attempts?
Or, I could maybe be persuaded for submission/fetch/submission/history.

(And of course, renaming here implies renaming files/functions.)

db/database.go
```diff
@@ -94,6 +94,10 @@ type Backend interface {
     // Can return nil if the submission does not exist.
     GetSubmissionResult(assignment *model.Assignment, email string, shortSubmissionID string) (*model.GradingInfo, error);
 
+    // Get the result history for a specific user.
+    // An assignment must be specified.
```
eriq-augustine, line 98: No need to specify the part about the assignment.
All these methods may take pointers (for efficiency and consistency), but none (I think) allow courses or assignments to be nil/optional.
Putting that comment would actually make the rest of the documentation more confusing (since it would put the nil-ability of the rest of the arguments in doubt).

api/submission/fetch_submission_history.go
```diff
@@ -0,0 +1,49 @@
...
+type FetchSubmissionHistoryResponse struct {
+    FoundUser        bool `json:"found-user"`
+    FoundSubmissions bool `json:"found-submissions"`
+    GradingResults []*model.GradingResult `json:"grading-results"`
+}
+
+func HandleFetchSubmissionHistory(request *FetchSubmissionHistoryRequest) (*FetchSubmissionHistoryResponse, *core.APIError) {
+    response := FetchSubmissionHistoryResponse{
+        FoundUser:        false,
+        FoundSubmissions: false,
```
eriq-augustine, line 25: The semantics around FoundSubmissions are a bit strange and rely on knowledge of the DB state.

If the user has submission(s), the this will be true and the results will be non-empty.
(Makes sense.)

If the user has never made a submission, then this will be true and the results will be nil.
(Pretty strange.)

If the user made a submission that was removed (the top-level submission dir exists with no contents), then this will be true and the results will be an empty list.
(Also pretty strange.)

I think the thing to do is remove FoundSubmissions and return an empty slice from the DB when there are no submissions, but no errors.
(You see this behavior a lot with containers: return nil on error, and empty on missing cases.)

api/submission/fetch_submission_history_test.go
```diff
@@ -0,0 +1,122 @@
...
+    "github.com/eriq-augustine/autograder/api/core"
+    "github.com/eriq-augustine/autograder/config"
+    "github.com/eriq-augustine/autograder/model"
+    "github.com/eriq-augustine/autograder/util"
+    "path/filepath"
+    "reflect"
+    "testing"
+)
+
+func TestFetchSubmissionHistory(test *testing.T) {
```
eriq-augustine, line 13: I'm not a fan of this test.
The testing procedure and mechanisms are great (which I should hope so, since I wrote TestFetchSubmission).
But I don't think we need an almost duplicate test effectively testing disk.GetSubmissionContents.

Instead, try to make a test that tests some of the semantics specific to this endpoint.
E.g., testing the semantics I discussed in the comment about FoundSubmissions would be great.

db/disk/submission.go
```diff
@@ -323,3 +323,34 @@ func (this *backend) RemoveSubmission(assignment *model.Assignment, email string
...
+    if err != nil {
+        return nil, fmt.Errorf("Unable to read user submissions dir '%s': '%w'.", submissionsDir, err);
+    }
+
+    if len(dirents) == 0 {
+        return nil, nil;
+    }
+
+    for _, dirent := range dirents {
+        var submission *model.GradingResult;
```
eriq-augustine, line 345: I don't think you need this declaration, just use `:=` below.

eriq-augustine (changes requested):
Good start!

There are some tricky semantic issues, but we should be able to work those out pretty easily.

api/submission/fetch_submission_history.go
cfbeard: Agreed!  Resolved this now.

api/submission/fetch_submission_history.go
cfbeard: I committed a fix to this.  I did not fully understand, but I do now.  Thank you.  I will make sure to ask in the future.

api/submission/fetch_submission_history.go
cfbeard: Committing a fix for this.

api/submission/fetch_submission_history_test.go
cfbeard: Fixed and noted!

api/submission/routes.go
cfbeard: Hmm...I like attempts and I like submission/fetch/submission/history.  I think they both communicate their intention.  Attempts make sense to me from a student standpoint.  My argument for the latter would be that both fetch/submission and fetch/submissions return grading results.  As such, I think that submission/fetch/submission/history communicates that we are fetching a history of grading results.   What do you think?

db/database.go
cfbeard: I resolved through a commit, but, if it is ok with you, I might leave it unresolved until we come up with a naming scheme for the API endpoint.  I think the documentation should maybe reflect what we come up with?

db/disk/submission.go
cfbeard: Resolved and noted!

api/submission/fetch_submission_history.go
cfbeard: I think I am misunderstanding the cases here.  Can you help me understand why it would be true in case 2 and 3?  From my understanding, if a user has never made a submission or if they have no directory entries it will return nil, nil from:

```go
func (this *backend) GetSubmissionResultHistory(assignment *model.Assignment, email string) ([]*model.GradingResult, error) {
    submissions := make([]*model.GradingResult, 0);

    submissionsDir := this.getUserSubmissionDir(assignment.GetCourse().GetID(), assignment.GetID(), email);
    if !util.PathExists(submissionsDir) {
        return nil, nil;
    }

    dirents, err := os.ReadDir(submissionsDir);
    if err != nil {
        return nil, fmt.Errorf("Unable to read user submissions dir '%s': '%w'.", submissionsDir, err);
    }

    if len(dirents) == 0 {
        return nil, nil;
    }

    for _, dirent := range dirents {
        submission, err := this.GetSubmissionContents(assignment, email, dirent.Name());
        if err != nil {
            return nil, fmt.Errorf("Unable to get submission contents for %s: .", email);
        }

        submissions = append(submissions, submission);
    }

    return submissions, nil;
}
``` 
Thus entering the if gradingResult == nil block here:

```go
gradingResults, err := db.GetSubmissionResultHistory(request.Assignment, request.TargetUser.Email);
    if err != nil {
        return nil, core.NewInternalError("-607", &request.APIRequestCourseUserContext, "Failed to get submissions.").
            Err(err).Add("email", request.TargetUser.Email);
    }

    if gradingResults == nil {
        return &response, nil;
    }

    response.FoundSubmissions = true;
    response.GradingResults = gradingResults;

    return &response, nil;
}
``` 
Which would return an empty slice and no error.  I have no issue removing FoundSubmissions from the implementation, I just want to make sure I understand!  Thank you!

cfbeard: In my most recent commit I have removed FoundSubmissions and am returning an empty slice for cases 2 and 3.

api/submission/routes.go
cfbeard: Make this attempts

eriq-augustine: (As discussed live.)
Let's go with attempts: `submission/fetch/attempts`.

db/disk/submission.go
```diff
@@ -323,3 +323,32 @@ func (this *backend) RemoveSubmission(assignment *model.Assignment, email string
 
     return true, nil;
 }
+
+func (this *backend) GetAttempts(assignment *model.Assignment, email string) ([]*model.GradingResult, error) {
```
eriq-augustine, line 327: Let's change this function name to "GetSubmissionAttempts" to mirror "GetSubmissionHistory", etc.

api/submission/fetch_attempts.go
```diff
@@ -0,0 +1,42 @@
...
+
+type FetchAttemptsRequest struct {
+    core.APIRequestAssignmentContext
+    core.MinRoleGrader
+
+    TargetUser core.TargetUserSelfOrGrader `json:"target-email"`
+}
+
+type FetchAttemptsResponse struct {
+    FoundUser        bool `json:"found-user"`
```
eriq-augustine, line 17: Spacing here looks a little strange.
Should it align with the open bracket on the line below?

api/submission/fetch_attempts.go
```diff
@@ -0,0 +1,42 @@
...
+
+    if !request.TargetUser.Found {
+        return &response, nil;
+    }
+
+    response.FoundUser = true;
+
+    gradingResults, err := db.GetAttempts(request.Assignment, request.TargetUser.Email);
+    if err != nil {
+        return nil, core.NewInternalError("-607", &request.APIRequestCourseUserContext, "Failed to get submissions.").
```
eriq-augustine, line 35: s/"submission"/"submission attempts"/

(In case this looks confusing, I am doing a sed-style find replace shorthand to tell you to replace '"submission'" with '"submission attempts"'.
I know that this explanation makes the shorthand moot, but now I'll use the shorthand whenever.)

api/submission/fetch_attempts_test.go
```diff
@@ -0,0 +1,75 @@
...
+        {model.RoleGrader, "grader@test.com",  true, false, []*model.GradingResult{}},
+
+        // Grader, other.
+        {model.RoleGrader, "student@test.com", true, false, studentGradingResults},
+
+        // Grader, missing.
+        {model.RoleGrader, "ZZZ@test.com",     false, false, []*model.GradingResult{}},
+
+        // Student, self.
+        {model.RoleStudent, "",                true, true, studentGradingResults},
```
eriq-augustine, line 38: Make `result` nil in this case, since (unlike the other cases) this one will error out first.

Of course, the code you have works but this is a case of making things more clear to readers.

api/submission/routes.go
```diff
@@ -12,6 +12,7 @@ var routes []*core.Route = []*core.Route{
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/scores`), HandleFetchScores),
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/submission`), HandleFetchSubmission),
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/submissions`), HandleFetchSubmissions),
+    core.NewAPIRoute(core.NewEndpoint(`submission/fetch/attempts`), HandleFetchAttempts),
```
eriq-augustine, line 15: Scooch this up to keep lexicographic order.

db/disk/submission.go
```diff
@@ -323,3 +323,32 @@ func (this *backend) RemoveSubmission(assignment *model.Assignment, email string
...
+    if !util.PathExists(submissionsDir) {
+        return submissions, nil;
+    }
+
+    dirents, err := os.ReadDir(submissionsDir);
+    if err != nil {
+        return nil, fmt.Errorf("Unable to read user submissions dir '%s': '%w'.", submissionsDir, err);
+    }
+
+    if len(dirents) == 0 {
```
eriq-augustine, line 340: You can just remove this if block.
Since the loop will never be entered and the guard (range) is a trivial check, we can simplify a bit.

db/submission_test.go
```diff
@@ -59,3 +59,54 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+    ResetForTesting();
+    defer ResetForTesting();
+
+    assignment := MustGetTestAssignment()
+
+    // User has submissions
+    studentAttempts, err := GetAttempts(assignment, "student@test.com");
+    if err != nil {
+        test.Errorf("Failed to get student attempts when should be a success: %v", err);
+    }
```
eriq-augustine, line 73: Put a blank line between the error-check if block and the size-check if block.
I understand the intention of putting them close (they are apart of the same test subcase).
But as a general rule you always* want to put vertical spacing between different if-blocks.
This is because it is very easy for a reader (especially after a few hundred lines) to see the structure and mistakenly see an "else if" where there is the "if" (because the code is structured like that) and think the two code blocks are mutually exclusive.

If we were not writing Go then I would recommend to have the second "if" be an "else if",
but since we are we do not want to mess with the super common error checking pattern of a single if checking if err is not nil.

*99.999% of the time.

db/submission_test.go
```diff
@@ -59,3 +59,54 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
     }
 }
+
+func (this *DBTests) DBTestFetchAttempts(test *testing.T) {
+    ResetForTesting();
+    defer ResetForTesting();
+
+    assignment := MustGetTestAssignment()
+
+    // User has submissions
```
eriq-augustine, line 69: Try to make comments either a complete sentence or some Title Cased Short Comment.

eriq-augustine, line 69: Also, this comment does not directly reflect the code below it.
In the context of this test, "grader@test.com" is really the "user".
We do not want to imply that "student@test.com" is the user.
(It actually tricked me for a second.)

So instead, say something like "Fetch attempts from another user."

db/submission_test.go
```diff
@@ -59,3 +59,54 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+    studentAttempts, err := GetAttempts(assignment, "student@test.com");
+    if err != nil {
+        test.Errorf("Failed to get student attempts when should be a success: %v", err);
```
eriq-augustine, line 72: In tests like this where you are not in a test case loop, you will probably want to use `test.Fatalf()` over `test.Errorf()`,
since the fatal one will stop the test from running anymore.

db/submission_test.go
```diff
@@ -59,3 +59,54 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+    // User makes a submission, then that submission is removed.
+    // This case tests GetAttempts() when the submission directory exists but is empty.
+    graderSubmission := studentAttempts[0];
+    graderSubmission.Info.User = "grader@test.com";
+
+    err = SaveSubmission(assignment, graderSubmission)
+    if err != nil {
+        test.Errorf("Failed to save grader submission %v: ", err);
+    }
+
```
eriq-augustine, line 96: Make a check here that `GetAttempts()` does return a single result.
May as well also check the content of the results in addition to the number (I think it should be an exact match (I can't remember if any other fields will be changed by the DB)).

eriq-augustine (changes requested):
Looking pretty good!

Mainly smaller semantics around tests.

db/submission_test.go
```diff
@@ -59,3 +62,72 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+    // User makes a submission, then that submission is removed.
+    // This case tests GetSubmissionAttempts() when the submission directory exists but is empty.
+    graderSubmission := studentAttempts[0];
+    graderSubmission.Info.User = "grader@test.com";
+
+    err = SaveSubmission(assignment, graderSubmission);
+    if err != nil {
+        test.Fatalf("Failed to save grader submission %v: ", err);
```
eriq-augustine, line 99: This string format is off.

db/submission_test.go
```diff
@@ -59,3 +62,72 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+    if err != nil {
+        test.Fatalf("Failed to get grader attempts when should be a success (with empty result): %v", err);
+    }
+
+    if len(graderAttempts) != 0 {
+        test.Fatalf("Got a non-empty result when should be empty.");
+    }
+
+    // User makes a submission, then that submission is removed.
+    // This case tests GetSubmissionAttempts() when the submission directory exists but is empty.
```
eriq-augustine, line 93: The comment on this line (93) is a bit misplaced.
That is the general idea of this test function, but this tries to be specific to a test case.
But then this is the wrong test case for it.

So instead of this line, put a comment on the function (right above the `func` line) that lays out the general flow of this test function.
e.g., get submissions from another user, check for zero, hijack a submission, check for 1, remove, check for zero.
I also wouldn't bother mentioning the directory, since that is specific to the disk db.

This will probably also change some of your other comments in this function.

db/submission_test.go
```diff
@@ -59,3 +62,72 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+        test.Fatalf("Failed to save grader submission %v: ", err);
+    }
+
+    graderAttempts, err = GetSubmissionAttempts(assignment, "grader@test.com");
+    if err != nil {
+        test.Fatalf("Failed to get grader attempts when there should be one %v: ", err);
+    }
+
+    if len(graderAttempts) != 1 {
+        test.Fatalf("Fetch returned unexpected number of attempts. Expected: %d, actual: %d", 1, len(graderAttempts));
```
eriq-augustine, line 108: s/actual/Actual/

Also, put a period at the end of the string.

(This is just a super nit, but since we are making other changes may as well make this one too.)

db/submission_test.go
```diff
@@ -59,3 +62,72 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+func (this *DBTests) DBTestFetchAttempts(test *testing.T) {
+    ResetForTesting();
+    defer ResetForTesting();
+
+    assignment := MustGetTestAssignment();
+
+    // Fetch attempts from another user who has submissions.
+    studentAttempts, err := GetSubmissionAttempts(assignment, "student@test.com");
+    if err != nil {
+        test.Fatalf("Failed to get student attempts when should be a success: %v", err);
```
eriq-augustine, line 75: s/%s/'%v'./

Usually we single quote errors and end with a period.
Note that this is just our convention, not a go pattern.
And we are not very attached to it, we may change in the future.

This happens in a few places in this file.

db/submission_test.go
```diff
@@ -59,3 +62,72 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+    }
+
+    // Fetch attempts from self (grader). User does not have submissions.
+    graderAttempts, err := GetSubmissionAttempts(assignment, "grader@test.com");
+    if err != nil {
+        test.Fatalf("Failed to get grader attempts when should be a success (with empty result): %v", err);
+    }
+
+    if len(graderAttempts) != 0 {
+        test.Fatalf("Got a non-empty result when should be empty.");
```
eriq-augustine, line 89: Add the size of the results to the string.

eriq-augustine (changes requested):
Almost there!

Just some styling.
Note that a lot of these styling comments are just nits that I would usually just let slide on their own.
But since there were a couple, I figure we may as well fix them up in one go.

db/submission_test.go
```diff
@@ -59,3 +62,77 @@ func (this *DBTests) DBTestRemoveSubmission(test *testing.T) {
...
+        test.Fatalf("Returned false from RemoveSubmission() when should be true.");
+    }
+
+    graderAttempts, err = GetSubmissionAttempts(assignment, "grader@test.com");
+    if err != nil {
+        test.Fatalf("Failed to get grader attempts when should be a success (with empty result): '%v'.", err);
+    }
+
+    if len(graderAttempts) != 0 {
+        test.Fatalf("Unexpected result length. Expected: '%d', Actual: '%d'", 0, len(graderAttempts));
```
eriq-augustine, line 136: Missing period.

eriq-augustine (approved):
lgtm!
(Looks Good To Me)

There is one nit, but you can get that in at your leisure.

PR #62, opened 2024-02-04
Implemented file size checking of POST requests. Test is passing.

cfbeard:


api/core/request_fields_test.go
```diff
@@ -270,6 +270,41 @@ func TestBadPostFilesStoreFail(test *testing.T) {
     }
 }
 
+func TestBadPostFilesFileSizeExceeded(test *testing.T) {
+    // Set size to 1 byte for testing. (a.txt is 2 bytes)
+    config.WEB_MAX_FILE_SIZE.Set(1);
```
eriq-augustine, line 275: You will need to reset the value at the end of the test.
You should use a defer for that.

config/options.go
```diff
@@ -41,6 +41,7 @@ var (
 
     // Server
     WEB_PORT = MustNewIntOption("web.port", 8080, "The port for the web interface to serve on.");
+    WEB_MAX_FILE_SIZE = MustNewIntOption("web.maxsize", 4 << 20, "The maximum allowed file size submitted via POST request.");
```
eriq-augustine, line 44: For options that users set, we want to make them a bit more readable.
Make this value in KB.

You can also make this a float so you can specify smaller values (like in tests).

eriq-augustine, line 44: Move it down to 2MB.
I looked at CSE40 submissions last year, and only students with issues were anywhere close to 500KB.

eriq-augustine (changes requested):
Looking pretty good.

The core of this is good, it just needs some work around the error and config semantics.

api/core/request_fields.go
```diff
@@ -205,8 +219,16 @@ func checkRequestPostFiles(request *http.Request, endpoint string, apiRequest an
     postFiles, err := storeRequestFiles(request);
 
     if (err != nil) {
-        return NewBareInternalError("-029", endpoint, "Failed to store files from POST.").Err(err).
+        var sizeExceededErr *FileSizeExceededError;
+        switch {
+        case errors.As(err, &sizeExceededErr):
```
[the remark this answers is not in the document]
cfbeard: Why does go fmt do this?  I also felt this looked weird.

api/core/request_fields.go
```diff
@@ -292,12 +314,20 @@ func storeRequestFiles(request *http.Request) (*POSTFiles, error) {
...
 
 func storeRequestFile(request *http.Request, outDir string, filename string) error {
-    inFile, _, err := request.FormFile(filename);
+    inFile, header, err := request.FormFile(filename);
     if (err != nil) {
         return fmt.Errorf("Failed to access request file '%s': '%w'.", filename, err);
     }
     defer inFile.Close();
 
+    if (int(header.Size) > config.WEB_MAX_FILE_SIZE.Get()) {
```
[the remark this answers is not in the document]
cfbeard: Oh!  Good to know!  Thank you.

config/options.go
cfbeard: My test no longer passes when I make this a float.  I will look further into it.

api/core/request_fields.go
```diff
@@ -205,8 +219,16 @@ func checkRequestPostFiles(request *http.Request, endpoint string, apiRequest an
     postFiles, err := storeRequestFiles(request);
 
     if (err != nil) {
-        return NewBareInternalError("-029", endpoint, "Failed to store files from POST.").Err(err).
+        var sizeExceededErr *FileSizeExceededError;
+        switch {
+        case errors.As(err, &sizeExceededErr):
+            msg := fmt.Sprintf("File '%s' is %d bytes. The maximum allowable size is %d bytes.", sizeExceededErr.Filename, sizeExceededErr.Filesize, config.WEB_MAX_FILE_SIZE.Get());
```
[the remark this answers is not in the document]
eriq-augustine: I don't have time to deep dive into this now, so this is just a passing comment.
But what you say make it sounds like it is being simplified, which is typically a good thing.

config/options.go
eriq-augustine: This comment is probably not necessary.
Remember that comments on my comments (replies to the thread) should indicate a discussion that I need to take part in.
No need to update me on every step, only when you need feedback or questions.
Otherwise, you can leave the comments until you are ready to resolve or continue the discussion.

eriq-augustine (changes requested):
Comments in email.

common/filespec_test.go
```diff
@@ -122,12 +122,12 @@ var testCasesCopy []*testCaseCopy = []*testCaseCopy{
     &testCaseCopy{
         FileSpec{Type: "path", Path: "_tests/files"},
         false,
-        "6481ba58a09ec62ff50325df54cf8966",
+        "3ec09caa093ac5b9211001adf722a418",
```
eriq-augustine, line 125: Instead of changing the hash here, what about making a subdirectory in files ("filespec_test"?) and putting some file in it (it could be a copy of a.txt with a new name)?
Then, we won't have to update this again if we need to add more test files.
We would still have to recompute this specific hash, but we should be protected later.

config/options.go
```diff
@@ -41,6 +43,7 @@ var (
 
     // Server
     WEB_PORT = MustNewIntOption("web.port", 8080, "The port for the web interface to serve on.");
+    WEB_MAX_FILE_SIZE = MustNewIntOption("web.maxsize", MAX_FILE_SIZE_KB, "The maximum allowed file size submitted via POST request.");
```
eriq-augustine, line 46: Put "KB" / "kb" into both the variable name and option key.
(So when people use it they already know the units.
You can see this pattern in this same file with `TASK_MIN_REST_SECS`.)

config/options.go
```diff
@@ -1,5 +1,7 @@
 package config
 
+const MAX_FILE_SIZE_KB = 2 * 1024 // 2048 KiB (2 MiB)
```
eriq-augustine, line 3: You can just put the value (without the constant variable) in the option.
Then we don't need a comment, but can mention the default value in the help message.
This also let's people looking at the help message and not the code know what the default is.

eriq-augustine (changes requested):
Looking pretty good.
Only small comments.

api/core/error.go
```diff
@@ -33,6 +33,16 @@ const (
     HTTP_PERMISSIONS_ERROR = http.StatusForbidden;
 )
 
+// A special error for when a submitted file exceeds the defined maximum allowable size.
+type fileSizeExceededError struct {
```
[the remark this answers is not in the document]
cfbeard: Just to clarify, we want to move this to the bottom of the request_fields.go file instead of here?

api/core/request_fields_test.go
```diff
@@ -270,6 +270,46 @@ func TestBadPostFilesStoreFail(test *testing.T) {
...
+    }
+
+    handler := func(request *requestType) (*any, *APIError) {
+        return nil, nil;
+    }
+
+    routes = append(routes, NewAPIRoute(endpoint, handler));
+
+    // Two paths provided: a.txt is under the size limit, 1092bytes.txt is over the size limit.
+    paths := []string{
```
eriq-augustine, line 296: You can just drop a.txt from your paths in this test.
We already tested that it worked in a previous test, so we will want to keep it simple.

config/options.go
eriq-augustine: Unresolving.

The option was updated, but this const was not removed.

api/core/request_fields.go
```diff
@@ -205,8 +207,17 @@ func checkRequestPostFiles(request *http.Request, endpoint string, apiRequest an
...
     if (err != nil) {
-        return NewBareInternalError("-029", endpoint, "Failed to store files from POST.").Err(err).
-                Add("struct-name", structName).Add("field-name", fieldType.Name);
+        var fileSizeExceededError *fileSizeExceededError
+        switch {
+            case errors.As(err, &fileSizeExceededError):
+                return NewBareBadRequestError("-036", endpoint, err.Error()).Err(err).
+                    Add("struct-name", structName).Add("field-name", fieldType.Name).
+                    Add("filename", fileSizeExceededError.Filename).Add("file-size", fileSizeExceededError.FileSizeKB).
+                    Add("max-file-size-kb", fileSizeExceededError.MaxFileSizeKB)
```
cfbeard, line 216: I know you had me use a normal type switch here but I have two arguments for this method.  One, this allows easy access of the fields of the error.  And two, if the error is wrapped (which I realize it isn't now) it would always fail the type switch.  If something were to change for some reason and this error got wrapped in the future it might introduce a stealthy bug.  Maybe I am thinking too deeply on it and this is not the way to go about it.  Lemme know what you think.

config/options.go
```diff
@@ -41,6 +41,7 @@ var (
 
     // Server
     WEB_PORT = MustNewIntOption("web.port", 8080, "The port for the web interface to serve on.");
+    WEB_MAX_FILE_SIZE_KB = MustNewIntOption("web.maxsizekb", 2 * 1024, "The maximum allowed file size (in KB) submitted via POST request.  The default is 2000 KB (2 MB).");
```
eriq-augustine, line 44: Double spaces before the last sentence.

eriq-augustine, line 44: 2000KB is not 2MB, 2000kB is 2MB (note the casing).
Outside of networks and storage, we only use K/Ki (2^10), not k (1000).
So despite any casing, we will **always** use 2^10 as k/K (ki/Ki).

eriq-augustine (changes requested):
Just a couple things left.

eriq-augustine (approved):
lgtm!

PR #66, opened 2024-03-18
Added Role Based Addressing Functionality to Report Field and Send Email Command

Lucas-Ellenberger:
Allow courses to specify a role or * (all roles) to send course reports, in addition to specific emails. The send-email command supports this functionality via a new CLA. The tests show that the new function supports an empty input, sorts emails properly, handles matching a role to users, the * to all users with a role, and removes duplicate emails. Redundant information like roles and a * do not produce multiple copies of an email. Emails are returned in all lower case to handle mixed case emails not duplicating the same email. (Ex: email@test.com, EMAIL@test.com only returns email@test.com) This works because a different casing of an email is still the same email.


cmd/send-email/main.go
```diff
@@ -6,13 +6,16 @@ import (
     "github.com/edulinq/autograder/config"
     "github.com/edulinq/autograder/email"
     "github.com/edulinq/autograder/log"
+    "github.com/edulinq/autograder/db"
```
eriq-augustine, line 9: Make sure imports are sorted within their group.

cmd/send-email/main.go
```diff
@@ -6,13 +6,16 @@ import (
...
 )
 
 var args struct {
     config.ConfigArgs
     To []string `help:"Email recipents." required:""`
     Subject string `help:"Email subject." required:""`
     Body string `help:"Email body." required:""`
+    //TODO: fix this to be an optional argument
```
eriq-augustine, line 17: Generally PRs should have TODOs.

cmd/send-email/main.go
```diff
@@ -25,7 +28,18 @@ func main() {
         log.Fatal("Could not load config options.", err);
     }
 
-    err = email.Send(args.To, args.Subject, args.Body, false);
+    // If database cannot get course and using role based addressing
```
eriq-augustine, line 31: Generally, we want comment to be complete sentences (with proper punctuation and casing).

cmd/send-email/main.go
```diff
@@ -25,7 +28,18 @@ func main() {
...
+    // Then, resolve users will return an error
+    // So, we do not error check on the get course method
+    course, err := db.GetCourse(args.Course);
```
eriq-augustine, line 34: The order/procedure here is a bit off.

If the user provides a course, then that course MUST exist and get retrieved properly.
Errors are NEVER (99.95% of the time) to be ignored (and this may ignore an error if a course was provided but failed to fetch).

cmd/send-email/main.go
```diff
@@ -25,7 +28,18 @@ func main() {
...
+    emailTo, err := db.ResolveUsers(course, args.To);
+    if (err != nil) {
+        log.Fatal("Failed to resolve users:'%s', '%w'.", course.GetName(), err);
+    }
```
eriq-augustine, line 38: You will want to separate independent if statements with vertical whitespace.

This falls under the category of "transparent coding", which I will mention sometimes.
It is about making your code as clear and readable as possible.
By having a whitespace here, it makes it easier for a reader to see that these are two different cases.
Otherwise, it is really easy for a reader to accidentally see an `if` as an `else if` and think that these cases are linked.

cmd/send-email/main.go
```diff
@@ -25,7 +28,18 @@ func main() {
...
     }
 
-    err = email.Send(args.To, args.Subject, args.Body, false);
+    // If database cannot get course and using role based addressing
+    // Then, resolve users will return an error
+    // So, we do not error check on the get course method
+    course, err := db.GetCourse(args.Course);
+    emailTo, err := db.ResolveUsers(course, args.To);
+    if (err != nil) {
+        log.Fatal("Failed to resolve users:'%s', '%w'.", course.GetName(), err);
```
eriq-augustine, line 37: Missing a space after ':'.

But that is actually moot. since courses should just be passed to a logging statement and not by name.
The logging infrastructure specifically recognizes courses and does extra work with them.
E.g., https://github.com/edulinq/autograder-server/blob/main/api/user/add.go#L81

cmd/send-email/main.go
```diff
@@ -25,7 +28,18 @@ func main() {
...
+    }
+    if (err != nil) {
+        log.Fatal("Could not resolve users.", err);
+    }
+    err = email.Send(emailTo, args.Subject, args.Body, false);
```
eriq-augustine, line 42: Add a newline above this.
This is for a similar, but different reason as the whitespace between the if statements.
This is just to increase readability.
You want a reader to see visual blocks as doing one thing / part of the same unit.

db/user.go
```diff
@@ -221,3 +222,45 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
 
     return syncResult, nil;
 }
+
+// Insert method that takes a course, email [] string and resolves based on *, roles, emails to all emails
```
eriq-augustine, line 226: Accidental comment?

This looks like a note to yourself.
This function could use a nice comment that describes what it does.

db/user.go
```diff
@@ -221,3 +222,45 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+func ResolveUsers(course *model.Course, emails []string) ([]string, error) {
+    if (backend == nil) {
+        return nil, fmt.Errorf("Database has not been opened.");
+    }
+    emailSet := map[string]any{};
+    roleSet := map[model.UserRole]any{};
+    // check for emails, roles, and * (all)
+    for _, email := range emails {
+        if (strings.Contains(email, "@") || (email == "")) {
```
eriq-augustine, line 235: It seems to me like empty strings should be thrown away.
It looks like these make it all the way through.

db/user.go
```diff
@@ -221,3 +222,45 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+            emailSet[strings.ToLower(email)] = nil;
+        } else {
+            if (email == "*") {
+                allRoles := model.GetAllRoles();
+                for role := range allRoles {
+                    roleSet[role] = nil;
+                }
+            } else {
+                roleSet[model.GetRole(strings.ToLower(email))] = nil;
```
eriq-augustine, line 244: What about invalid role strings?
Seems like a warning should be issued on those.

db/user_test.go
```diff
@@ -14,6 +15,83 @@ type SyncNewUsersTestCase struct {
     sendEmails bool
 }
 
+type ResolveUsersTestCase struct {
```
eriq-augustine, line 18: Should not be exported.

Edit:
See lower comments, this actually doesn't need to exist.
You can just use anonymous structs like in most other test cases.
E.g., https://github.com/edulinq/autograder-server/blob/main/db/submission_test.go#L13

db/user_test.go
```diff
@@ -14,6 +15,83 @@ type SyncNewUsersTestCase struct {
...
+        ResolveUsersTestCase{
+            input: make([]string, 0, 5),
+            course: course,
+        },
+        ResolveUsersTestCase{
+            input: make([]string, 0, 2),
+            course: course,
+        },
+    }
+    testCases[0].input = append(testCases[0].input, "");
```
eriq-augustine, line 78: Just define the slices when you define the test case.
Doing it here like this makes it much harder to maintain (what if you want to insert a new test case after the first one?).

db/user_test.go
```diff
@@ -14,6 +15,83 @@ type SyncNewUsersTestCase struct {
...
+    actualOutput []string
+    err error
+}
+
+func (this *DBTests) DBTestResolveUsers(test *testing.T) {
+    defer ResetForTesting();
+
+    for i, testCase := range getResolveUsersTestCases() {
+        ResetForTesting();
+        if (i == 4) {
```
eriq-augustine, line 31: Nope, never do this.
The index for a test case should only ever be used for reference (when outputting an error).
A test case should contain all the information it needs to test itself.

eriq-augustine (changes requested):
Good start.
All the core stuff is here, we just need to touch up some things.

Note that I will generally only make a comment once, but you should always ensure that your entire PR follows it.
(Like I may only comment on a style issue once, but it may apply multiple times.)

cmd/send-email/main.go
```diff
@@ -13,6 +14,7 @@ var args struct {
     To []string `help:"Email recipents." required:""`
     Subject string `help:"Email subject." required:""`
     Body string `help:"Email body." required:""`
+    Course string `help:"Course ID." default:""`
```
eriq-augustine, line 17: Add info in the help about how this is only necessary if there are user references that need to be resolved.

cmd/send-email/main.go
eriq-augustine: Unresolved.

This issue still exists.
Since the course is optional, it is not always fetched.

cmd/send-email/main.go
eriq-augustine: Unresolved.

This issue still exists.

db/user.go
```diff
@@ -221,3 +222,72 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
 
     return syncResult, nil;
 }
+
+// Convert roles to the associated user emails for a given course.
```
eriq-augustine, line 226: It's not just roles.

db/user.go
```diff
@@ -221,3 +222,72 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+// Returns a slice of lowercase emails without duplicates.
+// Will error if given non-emails without a valid course.
+// Will warn if given roles that do not exist within a course.
```
eriq-augustine, line 229: "Roles within a course" does not make sense.
All courses have the same set of available roles.

cmd/send-email/main.go
eriq-augustine: Unresolved.

This is still an issue in other comments.

db/user_test.go
```diff
@@ -14,6 +15,46 @@ type SyncNewUsersTestCase struct {
     sendEmails bool
 }
 
+func (this *DBTests) DBTestResolveUsers(test *testing.T) {
+    defer ResetForTesting();
+
+    testCases := []struct{input []string; expectedOutput []string; course *model.Course; addUsers []*model.User; removeUsers []string} {
+        {[]string{""}, make([]string, 0, 1), nil, nil, make([]string, 0, 1)},
```
eriq-augustine, line 22: Don't use make like this in test cases.
Just use a literal slice definition (like you did with `input`).

db/user_test.go
```diff
@@ -14,6 +15,46 @@ type SyncNewUsersTestCase struct {
...
+        {[]string{"b@test.com", "a@test.com", "c@test.com"}, []string{"a@test.com", "b@test.com", "c@test.com"}, nil, nil, make([]string, 0, 1)},
+        {[]string{"admin"}, []string{"admin@test.com"}, MustGetCourse(TEST_COURSE_ID), nil, make([]string, 0, 1)},
+        {[]string{"*"}, []string{"admin@test.com", "grader@test.com", "other@test.com", "owner@test.com", "student@test.com"},
+            MustGetCourse(TEST_COURSE_ID), nil, make([]string, 0, 1)},
+        {[]string{"other", "*", "grader@test.com", "zoinks@test.com", "ZoinKS@teSt.Com"}, []string{"admin@test.com", "grader@test.com",
```
eriq-augustine, line 27: Make these test cases more readable.
Put newlines between test cases.
Try not to break items, but break between items.

eriq-augustine (changes requested):
I stopped checking early.
There were too many issues that were marked as resolved from last time, but were not resolved.

db/user_test.go
Lucas-Ellenberger: Would you rather an incredibly lone single line for these cases?

cmd/send-email/main.go
```diff
@@ -4,12 +4,14 @@ import (
...
 
     "github.com/edulinq/autograder/config"
+    "github.com/edulinq/autograder/db"
     "github.com/edulinq/autograder/email"
     "github.com/edulinq/autograder/log"
 )
 
 var args struct {
     config.ConfigArgs
+    Course string `help:"Optional Course ID. The course is only required when emailing by role or the *, which denotes all roles." arg:"" optional:""`
```
eriq-augustine, line 14: The help string is reading a little awkward.
Maybe something like:
"Optional Course ID. Only required when roles or * (all course users) is are in the recipients."

cmd/send-email/main.go
```diff
@@ -25,6 +27,15 @@ func main() {
         log.Fatal("Could not load config options.", err);
     }
 
+    if (args.Course != "") {
+        course := db.MustGetCourse(args.Course);
+
+        args.To, err = db.ResolveUsers(course, args.To);
+        if (err != nil) {
+            log.Fatal("Failed to resolve users: '%w'.", err);
```
eriq-augustine, line 35: Include the course (just stick it on as an arg).

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
 
     return syncResult, nil;
 }
+
+// ResolveUsers is a helper function to map a User.Role to all User.Email's for that course.
+// The function takes an optional course and a list of strings, containing emails, roles, and * as input.
+// The function returns an alphanumerically sorted slice of lowercase emails without duplicates.
+// The function returns an error if a course cannot be opened and the input contains roles or a *.
+// The function will issue a warning if it's given a valid course and an invalid role.
```
eriq-augustine, line 230: The function comment need a little work.
 - The first line is a little factually off (since a string could be a *, and it is never actually a role (just a string representing a role).
 - Since we do not actually have users here, don't use User.* (e.g. User.Role and User.Email).
 - I don't like the "The function ..." four times in a row. Try it a little more naturally. "This function takes ... and returns ... . On an error ... . A warning will be logged if ... ."

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+func ResolveUsers(course *model.Course, emails []string) ([]string, error) {
+    emailSet := map[string]any{};
+    roleSet := map[string]any{};
+
+    // Iterate over every given string, checking for emails, roles, and *, which denotes all roles.
```
eriq-augustine, line 235: I think * is all users, not all roles.

We may have previously discussed it as all roles (since that makes sense to handle it internally like that),
but we should probably refer to * as denoting all users.

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+                    roleSet[role] = nil;
+                }
+            } else {
+                roleSet[strings.ToLower(email)] = nil;
+            }
+        }
+    }
+
+    if (len(roleSet) > 0) {
+        if (backend == nil) {
```
eriq-augustine, line 256: I think this should probably be moved up to the front like with most DB functions.

I can see an argument that this functionality doesn't necessarily require a DB (which you can try to make),
but even when sending emails we may log (which does want a database).

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+            return nil, fmt.Errorf("Database has not been opened.");
+        }
+
+        users, err := GetUsers(course);
+
```
eriq-augustine, line 261: Extra newline.

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+            return nil, err;
+        }
+
+        allowedRoles := make([]string, 0, len(model.GetAllRoleStrings()));
+        for role := range model.GetAllRoleStrings() {
+            allowedRoles = append(allowedRoles, role);
+        }
+
+        for roleString := range roleSet {
+            if (!slices.Contains(allowedRoles, roleString)) {
```
eriq-augustine, line 272: Instead of this (and building `allowedRoles`),
you can just use `model.GetRole()` and make sure the result is not the zero value (`model.RoleUnknown`).

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+                log.Warn("Resolve Users was given an invalid role.", log.NewCourseAttr(course.GetID()), log.NewAttr("Role", roleString));
```
eriq-augustine, line 273: Bad course logging.
Really make sure you understand this, we have talked about it a lot and it is important.

eriq-augustine, line 273: Use "role" instead of "Role".
We usually make out logging keys snake_case.

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+            }
+        }
+
+        for _, user := range users {
+            // Add a User.Email if their User.Role is in the role set.
```
eriq-augustine, line 278: Using the attributes in this comment make it a little harder to read (and factually inaccurate (an email does not have a role)).
Try something more natural, like:
"Add a user if their role is set."

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+// The function takes an optional course and a list of strings, containing emails, roles, and * as input.
+// The function returns an alphanumerically sorted slice of lowercase emails without duplicates.
+// The function returns an error if a course cannot be opened and the input contains roles or a *.
+// The function will issue a warning if it's given a valid course and an invalid role.
+func ResolveUsers(course *model.Course, emails []string) ([]string, error) {
+    emailSet := map[string]any{};
+    roleSet := map[string]any{};
+
+    // Iterate over every given string, checking for emails, roles, and *, which denotes all roles.
+    for _, email := range emails {
```
eriq-augustine, line 236: Go ahead and trim/strip `email` as the first thing inside this loop (even before the empty check).
Since emails and roles can't have whitespace, it's safe.

db/user.go
```diff
@@ -221,3 +222,74 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+            // Add a User.Email if their User.Role is in the role set.
+            _, ok := roleSet[model.GetRoleString(user.Role)];
+            if (ok) {
+                emailSet[strings.ToLower(user.Email)] = nil;
+            }
+        }
+    }
+
+    emailSlice := make([]string, 0, len(emailSet));
+
```
eriq-augustine, line 287: Remove this newline, so we can see `emailSlice` declared and filled in one visual block.

db/user_test.go
```diff
@@ -14,6 +15,88 @@ type SyncNewUsersTestCase struct {
...
+        // This is a test for our all roles character, the *. 
+        {
+            []string{"*"},
+            []string{"admin@test.com", "grader@test.com", "other@test.com", "owner@test.com", "student@test.com"},
+            nil,
+            []string{},
+        },
+
+        // This is a more complex test that adds multiple Users for the student role.
+        // This test case removes the only user from the "other" role, so we check that a role without any users still functions properly.
```
eriq-augustine, line 55: Try to only do one thing per test case (unless that case is specifically for the interaction of different components).
Like here, do the duplicate users, add user, and caps all in different test cases.

db/user_test.go
```diff
@@ -14,6 +15,88 @@ type SyncNewUsersTestCase struct {
...
+        // This test case is given duplicate emails of different capitalizations and duplicate roles.
+        // It tests to ensures we do not produce duplicates on this input.
+        {
+            []string{"other", "*", "grader@test.com", "zoinks@test.com", "ZoinKS@teSt.Com"},
+            []string{"admin@test.com", "grader@test.com", "other@test.com", "second_student@test.com", "student@test.com", "zoinks@test.com"},
+            []*model.User{model.NewUser("second_student@test.com", "", model.GetRole("student"))},
+            []string{"owner@test.com"},
+        },
+
+        // This test case tests if miscapitalized roles still function and that we warn on invalid roles.
```
eriq-augustine, line 65: There is no check for the warning, but there can be.
Just clear the logs and fetch warning logs.

You don't even need to check the contents, just the count should be sufficient.
Your test case struct can just have an int to represent the number of warnings created during the test case.

task/report.go
```diff
@@ -36,7 +37,12 @@ func RunReport(course *model.Course, to []string) error {
 
     subject := fmt.Sprintf("Autograder Scoring Report for %s", course.GetName());
 
-    err = email.Send(to, subject, html, true);
+    emailTo, err := db.ResolveUsers(course, to);
+    if (err != nil) {
+        return fmt.Errorf("Failed to resolve users for course '%s': '%w'.", course.GetName(), err);
```
eriq-augustine, line 42: Bad course logging.

eriq-augustine (changes requested):
Looking much better!
There is still work to do, but the progress is real good.

The semantics with cmd/send-email are really nice.

cmd/send-email/main.go
```diff
@@ -25,6 +27,18 @@ func main() {
         log.Fatal("Could not load config options.", err);
     }
 
+    if (args.Course != "") {
+        db.MustOpen();
+        defer db.MustClose();
```
eriq-augustine, line 32: The DB should generally be opened as the first thing after checking the args,
since it is used for things like logging.
So even in the case where no course is provided, the DB is required.

db/user.go
```diff
@@ -221,3 +222,71 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
 
     return syncResult, nil;
 }
+
+// ResolveUsers is a helper function to maps string representations of roles and * (all roles) to the emails for users with those roles.
```
eriq-augustine, line 226: Is it not a helper function, it's just a function.
There is no hard definition of a helper function, but this function is exported, which is definitely not true for helpers.

db/user.go
```diff
@@ -221,3 +222,71 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+// The function takes an optional course and a list of strings, containing emails, roles, and * as input and returns an alphanumerically sorted slice of lowercase emails without duplicates.
```
eriq-augustine, line 227: The course is not optional!
(We have talked about this quite a bit.)

eriq-augustine, line 227: The list is lexicographically sorted.
You can just say "sorted" or "lexicographically sorted", but "alphanumerically" is incorrect.

db/user.go
```diff
@@ -221,3 +222,71 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+// A fatal error occurs if a course cannot be opened and the input contains roles or a *.
```
eriq-augustine, line 228: Unnecessary comment (and I think not factual).
Remember, you always want your code to blend it with the code around it (e.g. in the same file and package).
No other function with a course says this.

db/user.go
```diff
@@ -221,3 +222,71 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+// A warning will be logged for every invalid role.
```
eriq-augustine, line 229: You can leave out this comment.
We normally don't comment on up-front about every warning that we may give,
and I see no reason for an exception here.

Note that this is for a very important version, not just to be lazy.
We want ever comment to be correct and stay correct (hence all the changes on this round of the PR).
This means that every comment has a maintenance burden associated with it,
and comments like that have a high burden (can easily change if anyone else modifies the function).

db/user.go
```diff
@@ -221,3 +222,71 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+func ResolveUsers(course *model.Course, emails []string) ([]string, error) {
+    if (backend == nil) {
+        return nil, fmt.Errorf("Database has not been opened.");
+    }
+
+    emailSet := map[string]any{};
+    roleSet := map[string]any{};
+
+    // Iterate over every given string, checking for emails, roles, and *, which denotes all users.
```
eriq-augustine, line 238: s/string/strings/

eriq-augustine, line 238: Use parens for the all users aside (a comma here, although correct, is confusing since it appears at the end of a sentence with a list.

s/, which denotes all users./ (which denotes all users)./

db/user.go
```diff
@@ -221,3 +222,71 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+    }
+
+    if (len(roleSet) > 0) {
+        users, err := GetUsers(course);
+        if (err != nil) {
+            return nil, err;
+        }
+
+        for roleString := range roleSet {
+            if (roleString == "unknown") {
```
eriq-augustine, line 266: This conditional is unnecessary.
The following conditional handles this case.

Remember that `model.RoleUnknown` is the zero value,
and is there fore returned from `model.GetRole()` when the result actually is "unknown" and when it is a role that is not known (e.g. "clowns").

Note that the unknown role is strictly a zero value and no user is every allowed to have that role.
So, even referring to the role (from outside the system) is not allowed.

db/user.go
```diff
@@ -221,3 +222,71 @@ func SyncUsers(course *model.Course, newUsers map[string]*model.User,
...
+            }
+
+            if (model.GetRole(roleString) == model.RoleUnknown) {
+                log.Warn("Invalid role given to ResolveUsers.", course, log.NewAttr("role", roleString))
+            }
+        }
+
+        for _, user := range users {
+            // Add a user if their role is set.
+            _, ok := roleSet[model.GetRoleString(user.Role)];
```
eriq-augustine, line 277: I'm not a huge fan of checking against the string roles map that still may have junk in it.
I would say that you should either build up a map/set of model.UserRole in the previous loop that you can check against,
or delete the bad entries where you warn in the previous loop.

db/user_test.go
```diff
@@ -14,6 +17,162 @@ type SyncNewUsersTestCase struct {
...
+
+        // This test case removes the only user from the "owner" role, so we check that a role without any users still functions properly.
+        {
+            []string{"owner", "student"},
+            []string{"student@test.com"},
+            nil,
+            []string{"owner@test.com"},
+            0,
+        },
+    };
```
eriq-augustine, line 136: Add two test cases:
 - Pass in an empty slice (you have one with only an empty string, but this one is even more empty).
 - Supply a single role, but have it resolve to nothing. (Like the last test case, without "student".)

db/user_test.go
```diff
@@ -14,6 +17,162 @@ type SyncNewUsersTestCase struct {
...
+
+    for i, testCase := range testCases {
+        ResetForTesting();
+        course := MustGetCourse(TEST_COURSE_ID);
+        // startTime := time.Time{};
```
eriq-augustine, line 141: Stray comment.

db/user_test.go
```diff
@@ -14,6 +17,162 @@ type SyncNewUsersTestCase struct {
...
+
+        for _, newUser := range testCase.addUsers {
+            SaveUser(course, newUser);
+        }
+
+        for _, oldUser := range testCase.removeUsers {
```
eriq-augustine, line 147: "oldUser" is not a good name (since the user is not old).
(I see the symmetry with "newUser, but we are using a different meaning of "new" here.)

"removeUser"?

db/user_test.go
```diff
@@ -14,6 +17,162 @@ type SyncNewUsersTestCase struct {
...
+            test.Errorf("Case %d (%+v): Incorrect Output. Expected: '%v', Actual: '%v'.", i,
+                testCase, testCase.expectedOutput, actualOutput);
+            continue;
+        }
+
+        logs, err := GetLogRecords(log.LevelWarn, time.Time{}, "", "", "")
+        if (err != nil) {
+            test.Errorf("Case %d (%+v): Error getting log records.", i, testCase);
+            continue;
+        }
```
eriq-augustine, line 167: Separate these checks with a newline.

eriq-augustine (changes requested):
Looking pretty good.
Only smaller things.

Be sure to really take your time with all your code.
Read everything over carefully.
There were a bunch of things in this one that were just factually inaccurate (like comments).

cmd/send-email/main.go
```diff
@@ -20,11 +22,23 @@ func main() {
         kong.Description("Send an email."),
     );
 
+    db.MustOpen();
```
eriq-augustine, line 25: The DB needs to be opened after the args are fully checked.
This is because the arguments/config can set where the DB lives.

Note that the previous comment mentioned the correct ordering:
> The DB should generally be opened as the first thing after checking the args,
> since it is used for things like logging.
> So even in the case where no course is provided, the DB is required.

eriq-augustine (changes requested):
Looking good.
Just one thing that needs to be fixed.

eriq-augustine (approved):
lgtm

PR #71, opened 2024-04-14
Docker limits fixed buffer

cfbeard:
Made a fixed sized buffer to limit output during docker grading.  Output limit is configurable with a default of 500 KB.  Thought about this for a while and 1 MB seemed like a lot for most use cases.  Since it is configurable, if there is a class that needs to support higher outputs, they can do so.  

Added a new error submission python file in _tests dir.  In the actual tests, I tried to mirror normal submission test functions in adaptability so even though there is only one error we are testing right now, if we added more in the future, it should be reasonably easy to do so without needing to alter the test functions (I hope).  

Made a new type of error in /common/errors.go so that we have a way of differentiating secure errors and errors we don't want to let students know about.  The buffer overflow error uses this struct.


common/errors.go
```diff
@@ -0,0 +1,10 @@
+package common
+
+// Errors that don't present security issues when returned to students.
+type SecureError struct {
```
eriq-augustine, line 4: I'm not a fan of these SecureErrors.
The concept makes sense, but I am worried about the design/implementation.

 - Are there endpoints aside from submit that need to worry about these?
 - What about api/core, does it need to know about these?
 - What about logging? We were thinking about students, but if we are thinking about sensitive information we should also think about what goes into the logs (currently no sensitive information is allowed to go in them).
 - Where is the safe place to output sensitive information?
I think doing this right requires us to go deeper.

We may also need to go in the opposite directions.
A type like this implies that all other errors are insecure, but we have not had this assumption up until now.
I think we first need to establish where error information can end up:
 - stdout/stderr
 - logs
 - api responses

Then, we can have something like "ConfidentialErrors" whose Error() method will output something safe.
But will know how to be interpreted by the `api/error` and `log` packages.
So anyone calling Error() will get something safe and the places where these errors may end up know how to better handle things.

These are my ideas, but this requires more thought.
Take a ponder on it.

eriq-augustine (changes requested):
This is looking real good,
but this PR is too big.

We need to split this into two PRs:
 1. Changing the output collection using the fixed buffer.
 2. Enforcing a size limit on submission output.

1 would include things like the test submission errors,
while 2 would have the secure errors (which has a lot to think about).

Start with just 1, since 2 depends on it.
But, think about the errors in 2 since those will take a lot of thought.

I have some critical questions for 1 that are in the comments,
but those aside  the main thing remaining is to add more tests.
Since you changed the behavior of output collection we should take this chance to add tests around there.
Tests like:
 - Just output stdout.
 - Just output stderr.
 - Output both stdout and stderr (one at a time).
 - Output both stdout and stderr (intermixed).
 - Output something and then kill the process/container. (Did we recover the output?)

eriq-augustine:
This is pretty far out-of-date, so I am going to close it.
But, we still want to go for this functionality.

PR #73, opened 2024-04-25
Api endpoint send course report

BatuhanSA:
I think it's going in the right direction; I don't think it's perfect, but it works.

### DIFF
--- a/api/submission/fetch_course_report.go
+++ b/api/submission/fetch_course_report.go
@@ -1,9 +1,11 @@
 package submission
 
 import (
-    "github.com/edulinq/autograder/api/core"
-    //"github.com/edulinq/autograder/model"
-    "github.com/edulinq/autograder/report"
+	// "fmt"
+
+	"github.com/edulinq/autograder/api/core"
+	//"github.com/edulinq/autograder/model"
+	"github.com/edulinq/autograder/report"
 )
 
 // permission erros
@@ -24,6 +26,7 @@ func HandleFetchCourseReport(request *FetchCourseReportRequest) (*FetchCourseRep
 	response := FetchCourseReportResponse{};
 
 	gettingCourseReport, err := report.GetCourseScoringReport(request.Course);
+
 
 	if err != nil {
         return nil, core.NewInternalError("-608", &request.APIRequestCourseUserContext, "Failed to get course report.").
--- a/api/submission/fetch_course_report.go
+++ b/api/submission/fetch_course_report.go
@@ -1,0 +1,36 @@
+package submission
+
+import (
+    "github.com/edulinq/autograder/api/core"
+    //"github.com/edulinq/autograder/model"
+    "github.com/edulinq/autograder/report"
+)
+
+// permission erros
+// non-existing course
+
+type FetchCourseReportRequest struct {
+    core.APIRequestCourseUserContext
+    core.MinRoleAdmin
+	
+}
+
+type FetchCourseReportResponse struct {
+	CourseReport *report.CourseScoringReport `json:"course-report"`
+}
+
+func HandleFetchCourseReport(request *FetchCourseReportRequest) (*FetchCourseReportResponse, *core.APIError) {
+
+	response := FetchCourseReportResponse{};
+
+	gettingCourseReport, err := report.GetCourseScoringReport(request.Course);
+
+	if err != nil {
+        return nil, core.NewInternalError("-608", &request.APIRequestCourseUserContext, "Failed to get course report.").
+            Err(err).Course(request.CourseID);
+    }
+
+	response.CourseReport = gettingCourseReport;
+
+	return &response, nil
+}
--- a/api/submission/fetch_course_report.go
+++ b/api/submission/fetch_course_report.go
@@ -1,20 +1,13 @@
 package submission
 
 import (
-	// "fmt"
-
 	"github.com/edulinq/autograder/api/core"
-	//"github.com/edulinq/autograder/model"
 	"github.com/edulinq/autograder/report"
 )
-
-// permission erros
-// non-existing course
 
 type FetchCourseReportRequest struct {
     core.APIRequestCourseUserContext
     core.MinRoleAdmin
-	
 }
 
 type FetchCourseReportResponse struct {
@@ -22,11 +15,9 @@ type FetchCourseReportResponse struct {
 }
 
 func HandleFetchCourseReport(request *FetchCourseReportRequest) (*FetchCourseReportResponse, *core.APIError) {
-
 	response := FetchCourseReportResponse{};
 
 	gettingCourseReport, err := report.GetCourseScoringReport(request.Course);
-
 
 	if err != nil {
         return nil, core.NewInternalError("-608", &request.APIRequestCourseUserContext, "Failed to get course report.").
--- a/api/submission/fetch_course_report_test.go
+++ b/api/submission/fetch_course_report_test.go
@@ -1,84 +1,103 @@
 package submission
-import (
-    "github.com/edulinq/autograder/api/core"
-    "github.com/edulinq/autograder/model"
-    //"github.com/edulinq/autograder/report"
-    "testing"
-	"fmt"
 
+import (
+    // "reflect"
+    "testing"
+    "reflect"
+	"github.com/edulinq/autograder/api/core"
+	"github.com/edulinq/autograder/model"
+	"github.com/edulinq/autograder/report"
+    "github.com/edulinq/autograder/util"
+    "github.com/edulinq/autograder/common"
 )
 
+
+
+
+
 func TestCourseReport(test *testing.T) {
+    testCases := []struct{
+        role model.UserRole
+        permError bool
+        result *report.CourseScoringReport
+    }{
+        // perfect test case
+        {model.RoleAdmin,false,Expected},
+        // Student access 
+        {model.RoleStudent,true,&report.CourseScoringReport{}},
+        // Nonexisting course
+        
 
-    response := core.SendTestAPIRequestFull(test, core.NewEndpoint(`submission/fetch/course-report`), nil, nil, model.RoleAdmin );
+    };
+    for i, testCase := range testCases {
+        response := core.SendTestAPIRequestFull(test, core.NewEndpoint(`submission/fetch/course-report`), nil, nil, testCase.role );
+        if (!response.Success) {
+            if (testCase.permError) {
+                expectedLocator := "-020";
+                if (response.Locator != expectedLocator) {
+                    test.Errorf("Case %d: Incorrect error returned on permissions error. Expcted '%s', found '%s'.",
+                        i, expectedLocator, response.Locator);
+                }
+            } else {
+                test.Errorf("Case %d: Response is not a success when it should be: '%v'.", i, response);
+            }
 
-    
-    fmt.Print(response);
+            continue;
+        }
+        var responseContent FetchCourseReportResponse
+        util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent);
+
+        if (!reflect.DeepEqual(testCase.result, responseContent.CourseReport)) {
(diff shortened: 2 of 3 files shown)


api/submission/fetch_course_report.go
```diff
@@ -0,0 +1,30 @@
+package submission
+
+import (
+	"github.com/edulinq/autograder/api/core"
```
eriq-augustine, line 4: Inconsistent use of tabs vs spaces in this file.
(This makes it hard to tell if your indentation is messed up in some parts.)

This may be true of other files, but I will only note it once.

api/submission/fetch_course_report.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/edulinq/autograder/report"
+)
+
+type FetchCourseReportRequest struct {
+    core.APIRequestCourseUserContext
+    core.MinRoleAdmin
```
eriq-augustine, line 10: Graders should also be able to fetch reports.

api/submission/fetch_course_report.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+type FetchCourseReportResponse struct {
+	CourseReport *report.CourseScoringReport `json:"course-report"`
+}
+
+func HandleFetchCourseReport(request *FetchCourseReportRequest) (*FetchCourseReportResponse, *core.APIError) {
+	response := FetchCourseReportResponse{};
```
eriq-augustine, line 18: Construct this closer to where it is used.
You can also (optionally) just construct it on the return (since this response is simple).

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,110 @@
+package submission
+
+import (
+	"reflect"
+	"testing"
+
+    "github.com/edulinq/autograder/common"
```
eriq-augustine, line 7: Import order is off.
Each group should be lexicographically sorted.

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,110 @@
...
+func TestCourseReport(test *testing.T) {
+    testCases := []struct{
+        role model.UserRole
+        courseId string 
+        coursError bool
+        permError bool
+        result *FetchCourseReportResponse
+    }{
+        // Admin Valid Course
+        {model.RoleAdmin,"course101",false,false,expected},
```
eriq-augustine, line 23: Commas should have whitespace after.

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+func TestCourseReport(test *testing.T) {
+    testCases := []struct{
+        role model.UserRole
+        courseId string 
+        coursError bool
+        permError bool
+        result *FetchCourseReportResponse
+    }{
+        // Admin Valid Course
```
eriq-augustine, line 22: I think you mean "Admin, Valid Course".
We want to try and keep our comments as grammatically correct as possible.

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,110 @@
...
+        {model.RoleAdmin,"course101",false,false,expected},
+
+        // Admin Invalid course
+        {model.RoleAdmin,"course102",true,false,expected},
+
+        // Student 
+        {model.RoleStudent,"course101",false,true,&FetchCourseReportResponse{}},
```
eriq-augustine, line 29: Just use a nil response on a perm error.

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,110 @@
...
+        courseId string 
+        coursError bool
+        permError bool
+        result *FetchCourseReportResponse
+    }{
+        // Admin Valid Course
+        {model.RoleAdmin,"course101",false,false,expected},
+
+        // Admin Invalid course
+        {model.RoleAdmin,"course102",true,false,expected},
```
eriq-augustine, line 26: The code you created does not deal with invalid courses (that is dealt with before your code is called by the API infrastructure).
So, that is not your responsibility to test.
In fact, we explicitly do not want you to test things that do not cover your purview.

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/edulinq/autograder/api/core"
+	"github.com/edulinq/autograder/model"
+	"github.com/edulinq/autograder/report"
+	"github.com/edulinq/autograder/util"
+)
+
+func TestCourseReport(test *testing.T) {
+    testCases := []struct{
+        role model.UserRole
+        courseId string
```
eriq-augustine, line 17: I don't think you even need to set the course.
See the comment below on line 26.

api/submission/routes.go
```diff
@@ -15,6 +15,7 @@ var routes []*core.Route = []*core.Route{
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/submissions`), HandleFetchSubmissions),
     core.NewAPIRoute(core.NewEndpoint(`submission/submit`), HandleSubmit),
     core.NewAPIRoute(core.NewEndpoint(`submission/remove`), HandleRemoveSubmission),
+    core.NewAPIRoute(core.NewEndpoint(`submission/fetch/course-report`), HandleFetchCourseReport),
```
eriq-augustine, line 18: Should in lexicographically sorted.

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,110 @@
...
+        util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent);
+        if (!reflect.DeepEqual(testCase.result, responseContent)) {
+            test.Errorf("Case %d: Unexpected submission result. Expected: '%s', actual: '%s'.", i,
+                util.MustToJSONIndent(testCase.result), util.MustToJSONIndent(responseContent));
+            continue;
+        }
+    }
+}
+
+var expected = &FetchCourseReportResponse{
```
eriq-augustine, line 65: Don't duplicate the code from report/course_test.
Just export the variable there and use it.

eriq-augustine (changes requested):
The core code it looking pretty good.

There are mostly style and maintainability things to fix.

api/submission/fetch_course_report.go
```diff
@@ -0,0 +1,26 @@
...
+    core.APIRequestCourseUserContext
+    core.MinRoleGrader
+}
+
+type FetchCourseReportResponse struct {
+    CourseReport *report.CourseScoringReport `json:"course-report"`
+}
+
+func HandleFetchCourseReport(request *FetchCourseReportRequest) (*FetchCourseReportResponse, *core.APIError) {
+    gettingCourseReport, err := report.GetCourseScoringReport(request.Course);
```
eriq-augustine, line 18: I don't like this variable name.
The "getting" pattern does not really make sense for anything that is not a promise.
I would say the obvious name here should be "courseReport".

api/submission/fetch_course_report.go
```diff
@@ -0,0 +1,26 @@
...
+
```
eriq-augustine, line 19: Remove this newline here.
We want to have the error check right next to the code that may produce the error.

As we go through this PR, you will notice comment like this (and the one above about variable names) that may seem pedantic and do not change how the code runs, but are actually very important from a software engineering viewpoint.
Things like this are important to make our code as readable (which helps towards maintainability) as possible.
Whenever you write code, try to think how someone who just read a few hundred lines of code will be able to read it.

eriq-augustine (changes requested):
Looking pretty good.
Mainly readability things.
I'm thinking we should also change the endpoint to a new report package (more thoughts in comments).

api/submission/fetch_course_report_test.go
```diff
@@ -0,0 +1,57 @@
...
+                test.Errorf("Case %d: Response is not a success when it should be: '%v'.", i, response);
+            }
+
+            continue;
+        }
+
+        var responseContent *FetchCourseReportResponse;
+        util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent);
+
+        var expectedResponse *FetchCourseReportResponse;
```
eriq-augustine, line 48: No need to serialize and deserialize the expected result.
With the response, we don't have a strict type on it, so we need to convert it, but you already have the expected type.

api/submission/routes.go
```diff
@@ -7,14 +7,15 @@ import (
 )
 
 var routes []*core.Route = []*core.Route{
-    core.NewAPIRoute(core.NewEndpoint(`submission/history`), HandleHistory),
-    core.NewAPIRoute(core.NewEndpoint(`submission/peek`), HandlePeek),
     core.NewAPIRoute(core.NewEndpoint(`submission/fetch/attempts`), HandleFetchAttempts),
+    core.NewAPIRoute(core.NewEndpoint(`submission/fetch/course-report`), HandleFetchCourseReport),
```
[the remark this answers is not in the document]
eriq-augustine: We don't want to the first one, because we may add other things in the future that may be considered "course reports".
For example, we can have a report on a courses resource utilization (disk, mem, cpu, etc).
What we have here is more of a "course submissions report".
We won't rename it internally for now (we'll wait until we have other actual reports),
but we want to be more careful with externally facing things, since we can't change those without a breaking change.

eriq-augustine (changes requested):
Responded to the question and added another comment.

api/submission/fetch_course_report_test.go
BatuhanSA: I marshaled the expected because of the formatting info that comes with the expected test from the report/test.go. Initially, the test was written for course_test.go; it has formatting fields because of the html tests. If I don't serialize and deserialize the test, the DeepEqulas will fail because of the formatting fields.

api/submission/routes.go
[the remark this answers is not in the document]
BatuhanSA: Should I relocate the files that I am working on to report/ or should I just change the API endpoint in routs?

api/submission/fetch_course_report_test.go
eriq-augustine: Okay that makes sense.
Add a comment so people will know.

api/submission/routes.go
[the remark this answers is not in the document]
eriq-augustine: Yes, you will need to move the files.

report/fetch_course_report.go
eriq-augustine, line 1: Name is a bit redundant.
Just use `fetch_course.go`.

Don't forget to update the test filename as well.

report/routs.go
eriq-augustine, line 1: Routes only makes sense in the context of the API.
They do not belong here.

report/main_test.go
```diff
@@ -4,11 +4,13 @@ import (
     "os"
     "testing"
 
+    "github.com/edulinq/autograder/api/core"
     "github.com/edulinq/autograder/db"
 )
 
 // Use the common main for all tests in this package.
 func TestMain(suite *testing.M) {
+    core.APITestingMain(suite, GetRoutes());
```
eriq-augustine, line 13: Why this change?
Why is the API called at all inside the report package?

eriq-augustine (changes requested):
You moved the files to the incorrect locations.
All API endpoints should be in the `api` package.
You should be moving things from `api/submission` to `api/report`.

api/report/routs.go
eriq-augustine, line 1: Filename is misspelled.

api/submission/routes.go
eriq-augustine, line 1: This file no longer falls under the purview of this PR, so you can drop these changes.

I know it is a bit awkward when you are cleaning something up, but then told not to clean it up.
Keeping the PRs focused is more important.
I'll go ahead and sort these on my side.

If you are confused as to how to remove your changes, you can do something like:
```
git checkout main api/submission/routes.go
```

eriq-augustine (changes requested):
Looking pretty good.

There are just a few small things.
I see a few unresolved comments.

Make those changes, and then we will hold this PR until the Python interface is also ready.

PR #74, opened 2024-04-27
Require acknowledgment for late submissions

shumbo:
Note: please review this PR with https://github.com/edulinq/autograder-py/pull/12

Currently, students can make submissions to overdue assignments without any warnings/confirmations. Students may accidentally submit old assignments for various reasons (running the submit command in the wrong directory, reusing the old `config.json` file without changing the assignment ID, etc.), and it would be nice if we could prevent students from accidentally making submissions and getting late penalties.

This PR requires a _late acknowledgment_ when students submit late if there is a late penalty. The following summarizes the changes I made:

- The `submit` API receives `LateAcknowledgment`, which defaults to `false`
- If the assignment is overdue, the late policy is set, and `LateAcknowledgment` is `false`, we reject the submission with `RejectMissingLateAcknowledgment`
- If the submission is rejected with `RejectMissingLateAcknowledgment`, the `submit` API sets `RequireLateAcknowledgment` to `True`, indicating that the submission requires an acknowledgment to be accepted.
- The [CLI](https://github.com/edulinq/autograder-py/pull/12) checks `RequireLateAcknowledgment` in case of rejection, and if true, prompts the user if they acknowledge the fact that the late policy will be applied by making the submission.
- If the user selects "yes", the CLI makes a new submission with `LateAcknowledgment=true`. The server accepts the submission and applies the late policy.

[Here is the terminal recording of this feature](https://asciinema.org/a/kYgam99aqZpdwwiLDwQhVLfd1).

### DIFF
--- a/api/submission/submit.go
+++ b/api/submission/submit.go
@@ -5,11 +5,13 @@ import (
     "github.com/edulinq/autograder/grader"
     "github.com/edulinq/autograder/log"
     "github.com/edulinq/autograder/model"
+    "github.com/edulinq/autograder/scoring"
 )
 
 type SubmitRequest struct {
     core.APIRequestAssignmentContext
     core.MinRoleStudent
+    LateAcknowledgment *bool `json:"late-acknowledgment"`
     Files core.POSTFiles
 
     Message string `json:"message"`
@@ -17,6 +19,7 @@ type SubmitRequest struct {
 
 type SubmitResponse struct {
     Rejected bool `json:"rejected"`
+    RequireLateAcknowledgment bool `json:"require-late-acknowledgment"`
     Message string `json:"message"`
 
     GradingSucess bool `json:"grading-success"`
@@ -25,6 +28,12 @@ type SubmitResponse struct {
 
 func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
     response := SubmitResponse{};
+
+    requireAcknowledgment := scoring.RequireLateAcknowledgment(request.Assignment);
+    if requireAcknowledgment && request.LateAcknowledgment != nil && !*request.LateAcknowledgment {
+        response.RequireLateAcknowledgment = true;
+        return &response, nil;
+    }
 
     result, reject, err := grader.GradeDefault(request.Assignment, request.Files.TempDir, request.User.Email, request.Message);
     if (err != nil) {
--- a/api/submission/submit.go
+++ b/api/submission/submit.go
@@ -10,7 +10,7 @@ import (
 type SubmitRequest struct {
     core.APIRequestAssignmentContext
     core.MinRoleStudent
-    LateAcknowledgment *bool `json:"late-acknowledgment"`
+    LateAcknowledgment bool `json:"late-acknowledgment"`
     Files core.POSTFiles
 
     Message string `json:"message"`
@@ -28,7 +28,7 @@ type SubmitResponse struct {
 func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
     response := SubmitResponse{};
 
-    result, reject, err := grader.GradeDefault(request.Assignment, request.Files.TempDir, request.User.Email, request.Message);
+    result, reject, err := grader.GradeDefault(request.Assignment, request.Files.TempDir, request.User.Email, request.Message, request.LateAcknowledgment);
     if (err != nil) {
         stdout := "";
         stderr := "";
--- a/api/submission/submit.go
+++ b/api/submission/submit.go
@@ -52,6 +52,7 @@ func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
         if _, ok := reject.(*grader.RejectMissingLateAcknowledgment); ok {
             response.RequireLateAcknowledgment = true;
         }
+        
         return &response, nil;
     }
 
--- a/api/submission/submit.go
+++ b/api/submission/submit.go
@@ -5,7 +5,6 @@ import (
     "github.com/edulinq/autograder/grader"
     "github.com/edulinq/autograder/log"
     "github.com/edulinq/autograder/model"
-    "github.com/edulinq/autograder/scoring"
 )
 
 type SubmitRequest struct {
@@ -29,12 +28,6 @@ type SubmitResponse struct {
 func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
     response := SubmitResponse{};
 
-    requireAcknowledgment := scoring.RequireLateAcknowledgment(request.Assignment);
-    if requireAcknowledgment && request.LateAcknowledgment != nil && !*request.LateAcknowledgment {
-        response.RequireLateAcknowledgment = true;
-        return &response, nil;
-    }
-
     result, reject, err := grader.GradeDefault(request.Assignment, request.Files.TempDir, request.User.Email, request.Message);
     if (err != nil) {
         stdout := "";
@@ -55,6 +48,10 @@ func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
 
         response.Rejected = true;
         response.Message = reject.String();
+
+        if _, ok := reject.(*grader.RejectMissingLateAcknowledgment); ok {
+            response.RequireLateAcknowledgment = true;
+        }
         return &response, nil;
     }
 
--- a/api/submission/submit.go
+++ b/api/submission/submit.go
@@ -10,10 +10,10 @@ import (
 type SubmitRequest struct {
     core.APIRequestAssignmentContext
     core.MinRoleStudent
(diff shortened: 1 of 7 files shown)


lkuper:
This change would be a big help for us in CSE114A. We've had a couple of students who copied `config.json` from an older assignment to a newer one(despite us explicitly warning them in the assignment README not to do that, and providing an example `config.example.json` file in the assignment repo), and submitting to the old assignment by mistake (and inadvertently using up their late days).

api/submission/submit.go
```diff
@@ -10,13 +10,15 @@ import (
...
     core.MinRoleStudent
+    LateAcknowledgment bool `json:"late-acknowledgment"`
     Files core.POSTFiles
 
     Message string `json:"message"`
 }
 
 type SubmitResponse struct {
     Rejected bool `json:"rejected"`
+    RequireLateAcknowledgment bool `json:"require-late-acknowledgment"`
```
eriq-augustine, line 21: I think we can generalize this fairly cheaply.
Instead of a field just for late rejections, you can all a `RejectReason` string field that pulls from an enum.

grader/reject.go
```diff
@@ -38,7 +38,17 @@ func (this *RejectWindowMax) String() string {
...
 }
 
-func checkForRejection(assignment *model.Assignment, submissionPath string, user string, message string) (RejectReason, error) {
+type RejectMissingLateAcknowledgment struct {}
+
+func (this *RejectMissingLateAcknowledgment) String() string {
+    return "This assignment is past the due date and the late policy will be applied."
+}
+
+func checkForRejection(assignment *model.Assignment, submissionPath string, user string, message string, lateAcknowledgment bool) (RejectReason, error) {
```
eriq-augustine, line 47: I'm not a fan of the specific argument (`lateAcknowledgment`) to bypass one rejection check.
I would love if we could generalize this.

I am specifically thinking about the future when we have direct/proxy submissions (a TA submitting on behalf of a student).
We may want to bypass all rejection checks.

Up in the API endpoint I mentioned using a rejection reason field instead of a specific one for being late.
We can expand on that concept and have a rejection reason bitset enum (like FileMode: https://pkg.go.dev/io/fs#FileMode).
The API endpoint can translate this enum to a string, and here we can pass a single uint32 instead of many booleans (this PR only has one, but this pattern would require more in the future).
Then, each specific rejection check can look at this value and see if their specific rejection reason should be ignored.
Note that this type would probably be in addition to our existing RejectReason type (since that carries more information with it), and not replace it (but you can try if you want).

api/submission/submit.go
```diff
@@ -46,6 +48,10 @@ func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
 
         response.Rejected = true;
         response.Message = reject.String();
+
+        if _, ok := reject.(*grader.RejectMissingLateAcknowledgment); ok {
+            response.RequireLateAcknowledgment = true;
+        }
```
eriq-augustine, line 54: Add a newline after this if.
I don't want the return visually attached to it.

grader/reject_test.go
```diff
@@ -1,14 +1,15 @@
...
 import (
-    "path/filepath"
-    "reflect"
-    "testing"
-
-    "github.com/edulinq/autograder/common"
-    "github.com/edulinq/autograder/config"
-    "github.com/edulinq/autograder/db"
-    "github.com/edulinq/autograder/model"
+	"path/filepath"
```
eriq-augustine, line 4: Inconsistent use of spaces and tabs in this file.

We have been meaning to make a ruling on go fmt, but until then try to keep files consistent.

grader/reject.go
```diff
@@ -125,3 +135,28 @@ func checkSubmissionLimitWindow(window *model.SubmittionLimitWindow,
 
     return nil, nil;
 }
+
+func checkLateAcknowledgment(assignment *model.Assignment, lateAcknowledgment bool) (RejectReason, error) {
```
eriq-augustine, line 139: I don't see a need for the error return value.

grader/reject.go
```diff
@@ -125,3 +135,28 @@ func checkSubmissionLimitWindow(window *model.SubmittionLimitWindow,
...
+    // if the user acknowledges that they are submitting late, do not reject the submission
```
eriq-augustine, line 140: Generally, we want comments to the be titles (Using Title Case) or complete sentences with punctuation.

grader/reject.go
```diff
@@ -125,3 +135,28 @@ func checkSubmissionLimitWindow(window *model.SubmittionLimitWindow,
...
+    if lateAcknowledgment {
+        return nil, nil
+    }
+
+    policy := assignment.GetLatePolicy()
+    
+    if policy.Type == model.EmptyPolicy || policy.Type == model.BaselinePolicy {
```
eriq-augustine, line 147: Why is baseline policy skipped?
Since that one is checked for (full) rejection, seems like it should be checked here as well.

grader/reject.go
```diff
@@ -125,3 +135,28 @@ func checkSubmissionLimitWindow(window *model.SubmittionLimitWindow,
...
+        return nil, nil
+    }
+
+    dueDate := assignment.DueDate
+
+    if dueDate.IsZero() {
+        return nil, nil
+    }
+
+    if common.NowTimestamp() >= dueDate {
```
eriq-augustine, line 157: I don't like the string comparison here.
This heavily depends on us not changing the timestamp format.
Instead, add a method to timestamp (`After()` maybe).

eriq-augustine (changes requested):
The concept here is looking pretty good.

There are some core changes I would like around generalizing the rejection reason and when to ignore rejection reasons.

Feel free to reach out if you want to synchronously chat.

api/submission/submit.go
```diff
@@ -10,13 +10,15 @@ import (
 type SubmitRequest struct {
     core.APIRequestAssignmentContext
     core.MinRoleStudent
+    LateAcknowledgment bool `json:"late-acknowledgment"`
```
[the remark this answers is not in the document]
shumbo: Fixed afd95a5d0a99f0c4a9de77fc04c25733c17b2512

api/submission/submit.go
shumbo: Fixed in e79974b2e3275d2c07527dc51f08f50118b472d4

I hope this is what you meant. Let me know if I misunderstood your comment.

grader/reject_test.go
shumbo: Fixed in 86203ab75f905f891ae7c49699ebcb5fbfe8a0ad

I must have run `go fmt` by accident. It would be great if we could adapt the formatter across the repository :+1:

grader/reject.go
shumbo: Fixed in a46adfacd6d39ffccde56fb8e6e068f677cdfa31, although this comment may change with other changes. But I will keep it in mind. Thanks!

grader/reject.go
shumbo: eeca6fb42f0cd2fcc0c243665a060acbb24e7edf

eriq-augustine:
@lkuper
Depending on how the changes in this PR goes, we may not be able to deploy these changes this quarter if they make a breaking API change (since there are several courses using the same server, and it would require an update to the Python interface).
We currently manage the situation of accidental submissions using `autograder.cli.submission.remove`.

@shumbo
Make sure that every PR comment is either resolved (the resolved button is clicked), or commented on if there is more discussion that needs to be had.
We explicitly don't want to respond to comments that are just resolved (since that makes more work that the review has to look at).
The comments you are making are nice, but in PRs we usually want to keep them terse.

shumbo:
@eriq-augustine Thanks again for your review.

> Make sure that every PR comment is either resolved (the resolved button is clicked), or commented on if there is more discussion that needs to be had.

Just to be sure, do you want me to resolve comments when I think I resolved them? (I know some reviewers prefer closing PR comments themselves after making sure they're actually resolved).

eriq-augustine:
> Just to be sure, do you want me to resolve comments when I think I resolved them? (I know some reviewers prefer closing PR comments themselves after making sure they're actually resolved).

I want you to resolve all threads that you believe are resolved.

grader/reject.go
eriq-augustine: Take a closer look at late policies.
I believe that the baseline policy does not reject if the submission is just late, only if it past the rejection date.

eriq-augustine:
Just checking in on this.
(No pressure, I just wanted to make sure you don't have any questions.)

eriq-augustine:
@shumbo Just another check in on this.
There is no timeline, I just wanted to see if you had any questions.

eriq-augustine:
Great, good to hear.
No rush.

eriq-augustine:
Hey Shun, did you still want to try for this?

eriq-augustine:
This has been resolved in #133.

PR #75, opened 2024-05-17
Centralized Locking PR

OliverLok:
Wanted to get your input on the current implementation of testing locking/unlocking.


common/lockmanager_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+    testUnlockingAnUnlockedLock(test)
+    
+    testUnlockingMissingKey(test)
+
+    testLockConcurrencyWithStaleCheck(test)
+}
+
+func testLockingKey() {
+    Lock(key)
```
eriq-augustine, line 34: This test cannot fail.
There is not much point in a test that cannot fail.

Since our lock manager should never panic (which is correct for most go code),
you will always need to call some method from *testing.T.

See comments on `testUnlockingKey()`.

common/lockmanager_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func testUnlockingKey(test *testing.T) {
```
eriq-augustine, line 37: There are two issues with this test and `testLockingKey()`:
 - They are dependent (each test should be independent).
 - They are too simple.

The last one may seem strange (usually we want simple),
but the problem here is that they are simple to the point of being incomplete.
Instead, just have a single test (`TestLockBase()`?) that just locks and unlocks.

common/lockmanager_test.go
```diff
@@ -0,0 +1,114 @@
...
+    err := Unlock(key);
+    if (err != nil) {
+        test.Errorf("Failed to unlock a key");
+    }
+}
+
+func testConcurrentLockingUnlocking() {
```
eriq-augustine, line 44: Vertical whitespace needs work.

eriq-augustine, line 44: For difficult tests like this, put in some comments to let people know what you are doing.

eriq-augustine, line 44: There are no test failures here?
(The test cannot fail.)

common/lockmanager_test.go
```diff
@@ -0,0 +1,114 @@
...
+        defer wg.Done();
+        Lock(key2);
+        defer Unlock(key2);
+    }();
+    Unlock(key);
+    Unlock(key2);
+
+    wg.Wait();
+}
+
```
eriq-augustine, line 63: I'm going to stop reviewing this file.
I think my comments above in this file will give you a good start and restructuring this and I think future comment will probably be repetitive.

common/main_test.go
```diff
@@ -32,4 +33,4 @@ func CleanupTestingMain() {
     if (err != nil) {
         log.Error("Error when removing temp dirs.", err);
     }
-}
+}
```
eriq-augustine, line 36: You editor seems to not be putting in final newline characters.

config/options.go
```diff
@@ -46,4 +46,6 @@ var (
     // Database
     DB_TYPE = MustNewStringOption("db.type", "disk", "The type of database to use.");
     DB_PG_URI = MustNewStringOption("db.pg.uri", "", "Connection string to connect to a Postgres Databse. Empty if not using Postgres.");
+
+    STALELOCK_DURATION = MustNewIntOption("stale.duration", 2, "Amount of time a lock can be unused before getting removed");
```
eriq-augustine, line 50: See an option like `TASK_MIN_REST_SECS`:
 - We want to put the unit at the end of the variable name so coders always know the units.
 - We can do math when computing the value to make it easier for coders to see what the value actually is (e.g. `86400` may seem a little random for a time duration, but `24 * 60 * 60` is much more clear).
 - For the actual key, we want something that tells users roughly what functionality this will be tied to. "stale" is not a good top-level key". Maybe something like "lockmanager.staleduration".

Two seconds is a very short duration.
I'm thinking a couple hours.

common/lockmanager.go
```diff
@@ -0,0 +1,90 @@
...
+    val, exists := lockMap.Load(key);
+    if !exists {
+        log.Error("Key does not exist.", log.NewAttr("key", key));
+        return fmt.Errorf("Key not found: '%s'.", key);
+    }
+    
+    lock := val.(*lockData);
+    if !lock.isLocked {
+        log.Error("Tried to unlock a lock that is unlocked with key.", log.NewAttr("key", key));
+        return fmt.Errorf("Key '%s' tried to unlock a lock that is unlocked.", key);
```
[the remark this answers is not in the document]
eriq-augustine: This one was partially missed.
Note that the comment targets two lines.
You fixed the second line, but not the first.

common/lockmanager_test.go
```diff
@@ -0,0 +1,130 @@
...
+        defer lockWaitGroup.Done();
+        Lock(key2);
+        defer func() {
+            err := Unlock(key2);
+            if (err != nil) {
+                t.Errorf("Failed to unlock a key.");
+            }
+        }()
+    }()
+
```
eriq-augustine, line 81: You need to verify that the other thread is still waiting before you unlock.
(Otherwise, you don't know if you hit the use case.)

This is actually a use case for a WaitGroup, but instead of being between the locking threads, it would be between the locking and main threads.

common/lockmanager_test.go
```diff
@@ -0,0 +1,130 @@
...
+    // Test if a lock gets past the first "if" in RemoveStaleLocksOnce but not the second because it had been acquired.
+    if (!testWithCondition(true)) {
+        t.Errorf("Lock was unexpectedly removed even though it was accessed concurrently.");
+    }
+
+    // Test if a lock gets past both "if's" and gets removed from the map.
+    if (testWithCondition(false)) {
+        t.Errorf("Stale lock was not removed.");
+    }
```
eriq-augustine, line 129: Just make these multiple tests.

Having these in multiple tests falls under the umbrella of transparent/clear/maintainable coding.
Currently, the true purpose of the test is a bit buried and disjoint from the name of the test (which people will see when the test fails).
But having multiple tests clears this up.

common/lockmanager_test.go
```diff
@@ -0,0 +1,130 @@
...
+        Lock(key1);
+        Unlock(key1);
+
+        val, _ := lockMap.Load(key1);
+        val.(*lockData).timestamp = time.Now().Add(-1 * (staleDuration + (1 * time.Second)));
+
+        lockWaitGroup.Add(1);
+        go func() {
+            defer lockWaitGroup.Done();
+            time.Sleep(100 * time.Millisecond);
```
eriq-augustine, line 106: That's a decently long sleep for a test, can we avoid (e.g. with a wait group) or shorten it?

eriq-augustine (changes requested):
Making good progress.

!!!
I see some deadlock conditions here around the use of `lockManagerMutex`.
Even in a simple case with two threads trying to lock the same key.
The second one locks the manager, and is then blocked on locking the key-based lock.
But then the first key cannot unlock itself (since it can't acquire the lock manager lock).
Take a real long think over this case.
I think the solution is for Lock() to just to release the lock manager lock before acquiring the key-based lock,
but you need to think more on this (since I just thought quickly).
Make sure you have tests for this case.
(Ideally, you would actually make the test that shows a deadlock first and then fix it.)

Needs more tests like (but not limited to):
 - Unlock an already unlocked key (see comments)
 - Normal/simple lock/unlock of multiple keys: lock1, lock2, unlock1, unlock2.
 - Calling `RemoveStaleLocksOnce()` with a  key locked.
 - Getting into the special second case when removing stale locks, where the state of the lock changes between the two checks.

common/lockmanager_test.go
OliverLok: Would it make more sense to use a channel as well? For example, the go routine could send True to the channel right before it tries to acquire the lock, and after the go func there could be a small sleep (or a waitgroup waiting) and then an if statement to confirm True got sent to the channel which indicates the go routine is waiting. I might be failing to see how just a waitgroup would be able to trigger a test fail.

eriq-augustine: You can use either.

The idea here is just that you want to wait until both threads hit a specific location to continue.
You can do this by blocking on a channel or by blocking on a wait group.
You just need your threads to block so you can ensure they both are at the correct spot.
Note that if you used a channel, you would have to be a bit careful since you don't know which thread will execute first (after the go routine is spawned).

common/lockmanager_test.go
```diff
@@ -0,0 +1,318 @@
...
+    key1 := "testkey1";
+    key2 := "testkey2";
+    var verifyWaitGroup sync.WaitGroup;
+    var threadWaitGroup sync.WaitGroup;
+
+    // Lock key1 for the first time.
+    Lock(key1);
+
+    threadWaitGroup.Add(1);
+    verifyWaitGroup.Add(1);
```
eriq-augustine, line 57: Consider more expressive names for your wait groups that are special for each test.
Like: threadCompletionBlock and key2LockedBlock.

This is not a required comment, but something you can decide.

common/lockmanager_test.go
```diff
@@ -0,0 +1,318 @@
...
+        threadsWaitGroup.Done();
+        err := Unlock(key2);
+        if (err != nil) {
+            test.Fatalf("Failed to Unlock")
+        }
+    }()
+    
+    // Wait until both threads have finished locking their keys so they 
+    // can unlock at the same time.
+    threadsWaitGroup.Wait();
```
eriq-augustine, line 116: The threads are not waiting.
WaitGroup.Done() does not block.
https://pkg.go.dev/sync#WaitGroup.Done

common/lockmanager_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+    // First thread aquires the lock a first time.
+    go func() {
+        defer firstThreadWaitGroup.Done();
+        Lock(key1);
+    }()
+
+    // Wait for the first thread to lock key1. 
+    firstThreadWaitGroup.Wait();
+
```
eriq-augustine, line 140: Extra newline.

common/lockmanager_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+    // Second thread aquires the lock a second time.
```
eriq-augustine, line 142: This comment is a little misleading.
You want to say that the second thread tries to acquire the lock, or that is blocks to acquire the lock.
We don't want the reader thinking that it was a direct acquisition like the first routine.

common/lockmanager_test.go
```diff
@@ -0,0 +1,318 @@
...
+
+    // Lock key1 for the first time.
+    Lock(key1);
+
+    threadWaitGroup.Add(1);
+    verifyWaitGroup.Add(1);
+
+    // This thread should wait to lock/unlock key1 a second time until it gets
+    // unlocked for the first time.
+    go func() {
```
eriq-augustine, line 185: The double defer pattern in this routine is very likely to confuse people (especially if they don't have a good grasp on defer ordering).
Do we need the second defer?
It will happen before the first one anyways, why not just call it directly?

common/lockmanager_test.go
```diff
@@ -0,0 +1,318 @@
...
+    var threadWaitGroup sync.WaitGroup;
+    var timestampWaitGroup sync.WaitGroup;
+
+    Lock(key1);
+    Unlock(key1);
+
+    // Load the lock data.
+    val, _ := lockMap.Load(key1); 
+    // Set the timestamp for a lock to be considered stale.
+    val.(*lockData).timestamp = time.Now().Add(-1 * (staleDuration + (1 * time.Second)));
```
eriq-augustine, line 226: If we are making the lock stale, then go ahead and make it really stale.
Just double the stale time.

common/lockmanager.go
```diff
@@ -0,0 +1,94 @@
...
+        // First check: If the conditions aren't met to remove a stale lock, return early.
+        if (time.Since(lock.timestamp) < staleDuration || lock.isLocked) {
+            return true;
+        }
+
+        // Lock the lock manager in case another thread is trying to lock/unlock.
+        lockManagerMutex.Lock();
+        defer lockManagerMutex.Unlock();
+        // Second check: If conditions are met to determine a stale lock, delete it.
+        if (time.Since(lock.timestamp) > staleDuration && lock.mutex.TryLock()) {
```
eriq-augustine, line 88: You can consider inverting this conditional too, so it matches with the one above.
Not required, your choice.

eriq-augustine (changes requested):
The tests are looking soooooo much better!
There is still stuff to do on them, but the patterns are looking good.
Note that the tests here will probably be the most rigorous ones in this entire project (and rightfully so given the nature of the locking code).

There is still one test missing, the test that passes both checks in RemoveStaleLocksOnce (note that because the inversion(s) this used to be slightly different).
I think you should be able to make this test if you manually lock/unlock `lockManagerMutex`.
So you can make sure a lock will pass the first guard, lock lockManagerMutex, call RemoveStaleLocksOnce on another thread, wait, edit the lock's lockData, unlock lockManagerMutex, make sure the lock still exists.

common/lockmanager_test.go
```diff
@@ -0,0 +1,318 @@
...
+func TestUnlockingKeyThatDoesntExist(test *testing.T) {
+    doesNotExistKey := "dne";
+
+    err := Unlock(doesNotExistKey);
+    if (err == nil) {
+        test.Fatalf("Failed to return an error when unlocking a key that does not exist.");
+    }
+}
+
+func TestLockUnlockDifferentKeysOneThread(test *testing.T) {
```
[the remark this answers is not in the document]
OliverLok: Should the function-level comment replace the individual-lined comments, and if so, should all the tests have a function-leveled comment for consistency instead of individual-lined comments?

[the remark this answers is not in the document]
eriq-augustine: If the function-level comment is enough to make the test clear, then it can replace the individual comments.
I won't know until I see it, so it's up to you to see if they would be sufficient for someone reading the tests.

common/lockmanager_test.go
OliverLok: That comment is meant for the threadsWaitGroup.Wait() line, so I'm not too sure what you mean by this.

eriq-augustine: This comment is in the correct spot.

WaitGroup.Done() does not block, so your other threads that called `threadsWaitGroup.Done()` are not synchronized.
Your comments indicates that you want those threads waiting at the same point, but they cannot be because Done() does not block.

OliverLok:
> The tests are looking soooooo much better! There is still stuff to do on them, but the patterns are looking good. Note that the tests here will probably be the most rigorous ones in this entire project (and rightfully so given the nature of the locking code).
> 
> There is still one test missing, the test that passes both checks in RemoveStaleLocksOnce (note that because the inversion(s) this used to be slightly different). I think you should be able to make this test if you manually lock/unlock `lockManagerMutex`. So you can make sure a lock will pass the first guard, lock lockManagerMutex, call RemoveStaleLocksOnce on another thread, wait, edit the lock's lockData, unlock lockManagerMutex, make sure the lock still exists.

Is this not describing what TestStaleLockRetention does? The only difference being manually locking/unlocking the lock manager and manually editing the locks lockData, which TestStaleLockRetention does by just calling Lock and Unlock which automatically acquires the lockmanagermutex and resets the timestamp.

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+    err := Unlock(key1);
+    if (err != nil) {
+        test.Fatalf("Failed to unlock.");
+    }
+}
+
+// Lock(1) -> %Lock(1) -> Unlock(1) -> %Unlock(1).
+// % denotes a thread.
+func TestLockingTwiceWithSameKeyConcurrently(test *testing.T) {
```
eriq-augustine, line 151: Currently, there is nothing in the test that makes sure that this didn't actually happen:
Lock(1) -> %Lock(1) -> %Unlock(1) -> Unlock(1).

If the code is working correctly this won't happen, but the whole point of tests is to ensure that we are working correctly.
So, you need a check before the first Unlock to make sure that the child thread has not locked yet.
I can see a couple ways to do this (and there are probably more):
 - Check the lock timestamp after the first lock and ensure that is has not changed before the first unlock.
 - Write to a channel after the lock in the child thread. Make sure that the channel is empty in the main thread before its unlock.

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+    // Unlock for the first time.
+    err := Unlock(key1);
+    if (err != nil) {
+        test.Fatalf("Failed to unlock for the first time.");
+    }
+    
+    // Wait for the thread to lock and unlock key1.
+    threadFinishBlock.Wait();
+}
+
```
eriq-augustine, line 195: Missing comment.

eriq-augustine: There's like four or so test functions with missing lock-syntax comments.

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+func TestStaleLockWithNonStaleLock(test *testing.T) {
+    Clear();
+    defer Clear();
+
+    key1 := "testkey1";
+    staleDuration := time.Duration(config.STALELOCK_DURATION_SECS.Get()) * time.Minute;
```
eriq-augustine, line 201: Incorrect time conversion
(mins instead of secs).

Check all conversions.

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+    var threadExecutionBlock sync.WaitGroup;
+    var timestampBlock sync.WaitGroup;
+
+    Lock(key1);
```
eriq-augustine, line 205: Add a comment to this lock/unlock.
Whenever you have code that some people may see as useless, you should probably have a comment on it.

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+    Unlock(key1);
+
+    val, _ := lockMap.Load(key1); 
+    // Set the timestamp for the lock to be considered stale.
+    val.(*lockData).timestamp = time.Now().Add(-1 * (staleDuration + (1 * time.Minute)));
```
eriq-augustine, line 210: You missed a previous comment from me that said to make stale timestamps very stale by doubling the time.

eriq-augustine, line 210: In this case, there is not even a reason to set this.
I think it makes the test a little more confusing to set this and then overwrite it without using it.
This test is just about making sure non-stale locks remain, so let's keep it simple.

If you want, you can make a test that ensure that locking and unlocking updates timestamps.
I am not requiring it, but it more directly tests what you are doing here.
(The test would be something like: set timestamp to -1, lock, check timestamp, set timestamp to -1, unlock, check timestamp.)

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+    err := Unlock(key1);
+    if (err != nil) {
+        test.Fatalf("Failed to unlock for the first time.");
+    }
+    
+    // Wait for the thread to lock and unlock key1.
+    threadFinishBlock.Wait();
+}
+
+func TestStaleLockWithNonStaleLock(test *testing.T) {
```
eriq-augustine, line 196: I don't like the test name, it's not very clear.
What about something like "TestStaleLockRemovalWithNonStaleLock" or "TestLockRetentionWithNonStaleLock".

This comment applies to several of your tests.
Review the test names.

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+    go func() {
+        defer staleCheckBlock.Done();
+
+        RemoveStaleLocksOnce();
+    }()
+    
+    // Wait for the goroutine to finish its execution.
+    staleCheckBlock.Wait();
+
+    // Check if the lock still exists in the lock map and return it.
```
eriq-augustine, line 268: I don't like this comment.
It talks about returning, but we are not returning anything
(the function we calls returns something, but we don't usually talk like that).

I also don't like how the comment above the conditional breaks the flow of code a bit.

To fix both of these, merge this comment with the one two lines down and put them both on this line.
This way, we have a higher quality comment and an uninterrupted block of code.

Note that this happens in a few places (at least the second part), look for them in this file.

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+    _, exists := lockMap.Load(key1);
+    // Test if the Locked lock got removed.
+    if (!exists) {
+        test.Fatalf("Failed to retain lock even though it was locked.");
+    }
+}
+
+func TestStaleLockPassesFirstCheckButNotSecond(test *testing.T) {
```
eriq-augustine, line 276: Bad name.

You generally don't want to reference code structures in your names.
Talk about what is happening in the test, not what lines of code you are trying to run.

Something like "TestLockRetentionWithMidCheckActivity".

common/lockmanager_test.go
```diff
@@ -0,0 +1,355 @@
...
+
+    // This thread will pass the first guard but wait until the lock's timestamp gets reset
+    // to pass the second guard.
+    go func() {
+        defer finishThreadBlock.Done();
+
+        RemoveStaleLocksOnce();
+    }()
+
+    // Small sleep to give the thread time to pass the first check in RemoveStaleLocksOnce().
```
eriq-augustine, line 305: You need another wait group before this sleep.
You generally don't want to just use a sleep to wait for a thread to start.
There is a TON of variability on how fast a thread can start.
So, use a lock (wait group in this case) to wait for a thread to start, and then use a sleep to let it run.

eriq-augustine (changes requested):
Looking pretty good.
We just need some more tweaking on the tests.

The overview comment makes the tests much more clear.
People know what to expect when coming into the function.
But, you don't have them on all tests.

common/lockmanager_test.go
OliverLok: Should I also not be referencing code structures in individual comments?

eriq-augustine: You want to try to avoid if it you can,
since it makes the comment more fragile and easier to become out-of-date
(e.g. what if a new check was added before the two existing checks).
Sometimes we can't avoid it, but we should always try to.

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+    key2LockUnlockBlock.Wait();
+
+    err := Unlock(key1);
+    if (err != nil) {
+        test.Fatalf("Failed to unlock key1 in the main thread.");
+    }
+}
+
+// Lock(1) -> %Lock(1) -> Unlock(1) -> %Unlock(1).
+func TestMainThreadUnlockFirst(test *testing.T) {
```
eriq-augustine, line 152: This name is too close to "TestMainThreadUnlocksFirst" (note the "s").
Either one can change their name, or both.

In this situation, I would probably scope each name with something like "OneLock"/"SingleLock" and "TwoLocks"/"DoubleLocks".
You can do it like that, or however you like.

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+    clear();
+    defer clear();
+
+    key1 := "testkey1";
+    var threadLockBlock sync.WaitGroup;
+    var threadFinishBlock sync.WaitGroup;
+
+    threadLockBlock.Add(1);
+    threadFinishBlock.Add(1);
+    unlockInChildThreadFirst := make(chan struct{}, 1);
```
eriq-augustine, line 162: I prefer a simpler chan, if possible.
What about a `chan bool` or `chan any`?

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    // Unlock for the first time.
+    err := Unlock(key1);
+    if (err != nil) {
+        test.Fatalf("Failed to unlock for the first time in the main thread.");
+    }
+    
+    // Wait for the thread to lock and unlock key1.
+    threadFinishBlock.Wait();
```
eriq-augustine, line 204: Add a final check on the chan size to make sure that the lock happened.

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+    go func() {
+        defer threadExecutionBlock.Done();
+
+        RemoveStaleLocksOnce();
+    }();
+
+    // Wait for the thread to finish checking for staleness.
+    threadExecutionBlock.Wait();
+    
+    // Check if the stale lock got removed.
```
eriq-augustine, line 230: The comment is a bit misleading, since the lock is not stale.

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+    _, exists := lockMap.Load(key1);
+    if (!exists) {
+        test.Fatalf("Failed to retain lock even though it was accessed concurrently.");
+    }
+} 
+
+// Lock(1) -> %Stale Check.
```
eriq-augustine, line 237: "Stale Check" is inconsistent.
In the previous test, it was "StaleCheck".

eriq-augustine, line 237: Maybe add something to this to indicate what you are doing with the timestamp.
Like: `Lock(1) -> MakeTimestampStale(1) -> %Stale Check.`.

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+    }();
+    
+    // Wait for the thread to finish checking for staleness.
+    staleCheckBlock.Wait();
+
+    // Check if the locked stale lock got removed. 
+    _, exists := lockMap.Load(key1);
+    if (!exists) {
+        test.Fatalf("Failed to retain lock even though it was locked.");
+    }
```
eriq-augustine, line 269: Add in the final unlock.

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// Lock(1) -> Unlock(1) -> Make Stale -> %Stale Check first check -> Update timestamp -> %Stale Check second check.
```
eriq-augustine, line 272: In addition to my other comments in the previous test applying,
we do not want to talk about "first" and "second" checks (what happens when we add in a new check between them).
Instead, you can just talk about "parts" or "halves", .e.g., `%StaleCheck(First Part)`.

common/lockmanager_test.go
```diff
@@ -0,0 +1,362 @@
...
+    lockManagerMutex.Unlock();
+
+    // Wait for the thread to finish checking for staleness.
+    finishThreadBlock.Wait();
+
+    // Check if the stale lock changed to un-stale mid check got removed.
+    _, exists := lockMap.Load(key1);
+    if (!exists) {
+        test.Fatalf("Failed to retain lock even though it was accessed concurrently.");
+    }
```
eriq-augustine, line 326: Add final unlock.

config/options.go
```diff
@@ -46,4 +46,6 @@ var (
     // Database
     DB_TYPE = MustNewStringOption("db.type", "disk", "The type of database to use.");
     DB_PG_URI = MustNewStringOption("db.pg.uri", "", "Connection string to connect to a Postgres Databse. Empty if not using Postgres.");
+
+    STALELOCK_DURATION_SECS = MustNewIntOption("lockmanager.staleduration", 2 * 60 * 60, "Amount of time a lock can be unused before getting removed.");
```
eriq-augustine, line 50: s/Amount of time/Number of seconds/

eriq-augustine (changes requested):
The tests are looking real nice.
We just need some cleanup here and there (the content is all good).

The section on the lock syntax needs some work.
I would make sure not to underestimate documentation like this.
It can be pretty tricky to write clearly.

common/lockmanager_test.go
OliverLok: It gets unlocked on line 284 though.

eriq-augustine: Yes, you are correct.

common/lockmanager.go
```diff
@@ -0,0 +1,121 @@
...
+    "time"
+
+    "github.com/edulinq/autograder/config"
+    "github.com/edulinq/autograder/log"
+)
+
+type lockData struct {
+    timestamp time.Time;
+    mutex sync.RWMutex;
+    isLocked int;
```
eriq-augustine, line 15: Change this to something that indicates its non-binary nature.
Maybe something like "lockCount"?

common/lockmanager.go
```diff
@@ -0,0 +1,121 @@
...
+    lock := val.(*lockData);
+
+    // Unlock the lockManagerMutex before acquiring the lock to avoid a deadlock.
+    lockManagerMutex.Unlock();
+
+    if (read) {
+        lock.mutex.RLock();
+    } else {
+        lock.mutex.Lock();
+        lock.isLocked += 1;
```
eriq-augustine, line 60: We actually want to track both cases with isLocked.
A lock is still not stale even if only read locks are being acquired.

Also, you can use ++ in go: `lock.isLocked++`.

common/lockmanager.go
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    if (read) {
+        lock.mutex.RUnlock();
+    } else {
+        lock.isLocked -= 1;
+        lock.mutex.Unlock();
+    }
+
+    lock.timestamp = time.Now();
```
eriq-augustine, line 89: Put a newline after this.

common/lockmanager.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func RemoveStaleLocksOnce() {
+    staleDuration := time.Duration(time.Duration(config.STALELOCK_DURATION_SECS.Get()) * time.Second);
+
+    lockMap.Range(func(key, val any) bool {
+        lock := val.(*lockData);
+
+        // First check: If the lock isn't stale or is locked, return early.
+        if (time.Since(lock.timestamp) < staleDuration || lock.isLocked == 1) {
```
eriq-augustine, line 106: We just want to check if the lock count is greater than zero.

common/lockmanager_test.go
```diff
@@ -0,0 +1,545 @@
+package common
+
+// Tests in this file will deal with locking, unlocking, and checking for stale locks (sometimes on multiple threads).
+// Since this can quickly get confusing, we will have comments for each test using the following syntax:
+// "%" denotes a thread off the main thread.
+// "!" denotes that the associated operation will produce an error.
```
eriq-augustine, line 6: s/will/should/g

A small thing, but having "should" here indicates that we will always test that.

common/lockmanager_test.go
```diff
@@ -0,0 +1,545 @@
...
+        test.Fatalf("Failed to unlock the lock.");
+    }
+
+    err = Unlock(key1);
+    if (err == nil) {
+        test.Fatalf("Failed to return an error when unlocking a lock that is already unlocked.");
+    }
+}
+
+// Unlock(key1).
```
eriq-augustine, line 52: !Unlock(key1)

common/lockmanager_test.go
```diff
@@ -0,0 +1,545 @@
...
+    finishThreadBlock.Wait();
+
+    // Check if the stale lock changed to not stale mid check got removed.
+    _, exists := lockMap.Load(key1);
+    if (!exists) {
+        test.Fatalf("Failed to retain lock even though it was accessed concurrently.");
+    }
+}
+
+// Lock(key1) -> Unlock(key1) -> Make Stale -> %StaleCheck.
```
eriq-augustine, line 335: "Make Stale" is inconsistent with previous tests ("MakeTimestampStale(key1)")

eriq-augustine (changes requested):
Looking real good!

Just the issue with the lock count for read locking and some small test things.

common/lockmanager.go
```diff
@@ -0,0 +1,124 @@
...
+    lockManagerMutex.Unlock();
+
+    if (read) {
+        lock.mutex.RLock();
+    } else {
+        lock.mutex.Lock();
+    }
+
+    lock.lockCount++;
+
```
eriq-augustine, line 63: Remove this newline.
We want to create a visual "update lock metadata" block.
(Applies to unlock() as well.)

eriq-augustine (approved):
LGTM!

There is one small vertical spacing issue I would liked changed.
But, you can include that in your next PR that replaces some current locking in the project with uses of the lock manager.

PR #76, opened 2024-05-29
course disk db uses the lockmanager

OliverLok:
Could also use course.getID() for the key instead of globally locking it.

### DIFF
--- a/common/lockmanager.go
+++ b/common/lockmanager.go
@@ -1,0 +1,124 @@
+package common
+
+import (
+    "fmt"
+    "sync"
+    "time"
+
+    "github.com/edulinq/autograder/config"
+    "github.com/edulinq/autograder/log"
+)
+
+type lockData struct {
+    timestamp time.Time;
+    mutex sync.RWMutex;
+    lockCount int;
+}
+
+const TICKER_DURATION_HOUR = 1.0;
+
+var (
+    lockManagerMutex sync.Mutex;
+    lockMap sync.Map;
+    ticker *time.Ticker;
+)
+
+func init() {
+    ticker = time.NewTicker(TICKER_DURATION_HOUR * time.Hour);
+    go removeStaleLocks();
+}
+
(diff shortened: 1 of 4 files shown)


db/disk/course.go
```diff
@@ -7,13 +7,14 @@ import (
     "github.com/edulinq/autograder/log"
     "github.com/edulinq/autograder/model"
     "github.com/edulinq/autograder/util"
+    "github.com/edulinq/autograder/common"
```
eriq-augustine, line 10: Import ordering is off.

db/disk/course.go
```diff
@@ -7,13 +7,14 @@ import (
...
 )
 
 const DISK_DB_COURSES_DIR = "courses";
 
 func (this *backend) ClearCourse(course *model.Course) error {
-    this.lock.Lock();
-    defer this.lock.Unlock();
+    common.Lock("course");
+    common.Unlock("course");
```
eriq-augustine, line 17: Missed defer.
Don't let your guard down in a simple PR.

eriq-augustine (changes requested):
See email.

On the bright side, the lock manager is looking super each to use.

But unfortunately, you missed the key point here.
This is still more or less a "global" lock.
Every course that is using the autograder shares the same lock (and therefore cannot do things at the same time).
But look at the resource that is being locked: `this.getCourseDir(course)`.
Which eventually becomes: `filepath.Join(this.baseDir, DISK_DB_COURSES_DIR, courseID)`.
https://github.com/edulinq/autograder-server/blob/3f3572bc5ea18893bb1333721e1bc8b237831f64/db/disk/course.go#L143

So, each course is working with their own specific directory.
They don't have to worry about interrupting eachother's work.
They can each lock independently of one another.
This is the "contextual locking" that is the whole intention of of the lock manager.
The code will be the same for each course, but they will each lock using their own "context".

PR #77, opened 2024-06-01
made db/disk/course.go use the lockmanager with contextual locking

OliverLok:


db/disk/course.go
```diff
@@ -107,8 +113,8 @@ func (this *backend) GetCourse(courseID string) (*model.Course, error) {
 }
 
 func (this *backend) GetCourses() (map[string]*model.Course, error) {
-    this.lock.RLock();
-    defer this.lock.RUnlock();
+    common.ReadLock("all-courses");
```
eriq-augustine, line 116: If we were going to go this route, then we would need to make a constant and add a prefix/suffix for the disk db.
Remember that the whole project will share the name space, so we want to be sure and not conflict.

However, I think the way to go here is to actually lock each one as we go.
You would need to make sure that each one gets unlocked.
There are several ways to do that, I would probably use an anonymous function so that I could continue to use `defer`.

eriq-augustine (changes requested):
Looking pretty good.

As you probably know, the real outlier here is LoadCourse() since it doesn't have an id at the ready.
I'm thinking that the way to handle this, is to change the locking key from the course's id, to the course config's absolute path.
This way, all the functions will have access to that path before it needs to read/write any resources.

Make sure to add a filepath.Abs() to the relevant path building functions at the bottom.
https://pkg.go.dev/path/filepath#Abs
filepath.Abs() will also clean the path and ensure that we are always using the canonical path (making it work as a lock key).

db/disk/course.go
```diff
@@ -24,8 +25,13 @@ func (this *backend) ClearCourse(course *model.Course) error {
 }
 
 func (this *backend) LoadCourse(path string) (*model.Course, error) {
-    this.lock.Lock();
-    defer this.lock.Unlock();
+    absolutePath, err := filepath.Abs(path);
```
eriq-augustine, line 28: Just use util.ShouldAbs().

eriq-augustine (changes requested):
Looks like there may be some key misconceptions here.

There is a mix of using the courses ID and path as a locking key.
But these both represent the same object, they should be locked with the same key.

Really take some time to make sure you understand how contextual locking works in the context of a course.

db/disk/course.go
```diff
@@ -124,6 +127,11 @@ func (this *backend) GetCourses() (map[string]*model.Course, error) {
             return nil, fmt.Errorf("Failed to load course '%s': '%w'.", configPath, err);
         }
 
+        func() {
+            common.ReadLock(course.GetID());
+            defer common.ReadUnlock(course.GetID());
```
eriq-augustine, line 132: Also, there is no logging on the error (mentioned via email).

db/disk/course.go
```diff
@@ -124,6 +127,11 @@ func (this *backend) GetCourses() (map[string]*model.Course, error) {
             return nil, fmt.Errorf("Failed to load course '%s': '%w'.", configPath, err);
         }
 
+        func() {
```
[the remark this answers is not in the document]
OliverLok: Should the anonymous function encapsulate everything inside the for loop then, or are you hinting at how it should be doing this asynchronously?

db/disk/course.go
eriq-augustine: It should NOT be done async.
Everything that you need to be inside the function should be in it.
(Sorry for the somewhat recursive definition, but you just need to put whatever you need to in it.)

db/disk/course.go
```diff
@@ -12,10 +13,13 @@ import (
 const DISK_DB_COURSES_DIR = "courses";
 
 func (this *backend) ClearCourse(course *model.Course) error {
-    this.lock.Lock();
-    defer this.lock.Unlock();
+    courseDir := this.getCourseDir(course);
+    courseDir = util.ShouldAbs(courseDir);
```
eriq-augustine, line 17: Instead of calling util.ShouldAbs() multiple times, just call it in getCourseDirFromID().

db/disk/course.go
```diff
@@ -24,16 +28,18 @@ func (this *backend) ClearCourse(course *model.Course) error {
 }
 
 func (this *backend) LoadCourse(path string) (*model.Course, error) {
-    this.lock.Lock();
-    defer this.lock.Unlock();
+    courseDir := util.ShouldAbs(path);
```
eriq-augustine, line 31: This is a path (to a file) not a dir.

Note that "path" can be used for both a file and a dir (and technically many other things).
But, in this case the path should be pointing to a file (usually called "course.json" in this case).

db/disk/course.go
```diff
@@ -59,12 +65,15 @@ func (this *backend) SaveCourse(course *model.Course) error {
...
+
     if (acquireLock) {
-        this.lock.Lock();
-        defer this.lock.Unlock();
+        common.Lock(courseDir);
+        defer common.Unlock(courseDir);
     }
 
-    util.MkDir(this.getCourseDir(course));
+    util.MkDir(courseDir);
```
eriq-augustine, line 76: Careful because of the above comment.
Make sure this gets changed when the above comment gets resolved.

eriq-augustine (changes requested):
The more consistent locking is looking good, but there is a bug (see below).

!!
There is a subtle and important bug here.
In some places, you are locking the course directory (e.g., clearCourse(), saveCourseLock()) and other places you are locking the course config file (e.g. LoadCourse()).
This can make for some subtle locking bugs.

I would advice locking the file over the directory.
(Since we can also put things like users in the directory and want to lock that independently.)

Note that these function examples are not exhaustive, I just stopped looking after a few.

db/disk/course.go
OliverLok: Do you mean calling it from getCoursePathFromID(), since you said you I should lock the file over the directory?

eriq-augustine: Take a look at the four path-creation functions at the bottom of this file.
I believe that all of them wrap getCoursePathFromID().
So if you abs there, they should all be absolute.

OliverLok: Yes but it looks like getCoursePathFromID adds the config filename while getCourseDirFromID just returns the directory of the course. Ex. /autograder/courses/CS101 vs /autograder/courses/CS101/courseConfig.json. Since you mentioned locking the file (assuming you were mentioning the config file) instead of the directory, would it not make more sense to use getCoursePathFromID for the key?

eriq-augustine: I think there are two key things here.
1) You should add abs to getCourseDirFromID(). I said the wrong function name before, but you probably saw that this is actually the function that is being wrapped.
2) It's important to conceptually think about what the locking key is and not just think of it as the output of a function. So I don't like this part from the previous comment: "would it not make more sense to use getCoursePathFromID for the key". We want to use the abs path to the course config file (which is held in the db). That can come from getCoursePathFromID(), but in other cases (like LoadCourse()), it is passed in directly (but needs to be normalized).

eriq-augustine:
LGTM!

PR #82, opened 2024-07-10
Users/Auth: New API Endpoint

Lucas-Ellenberger:
Added a new API endpoint for users/auth. The endpoint allows anyone to verify that the email, password pairing successfully authenticates with the server through a no-op.

### DIFF
--- a/internal/api/users/auth.go
+++ b/internal/api/users/auth.go
@@ -13,8 +13,7 @@ type AuthResponse struct {
 }
 
 func HandleAuth(request *AuthRequest) (*AuthResponse, *core.APIError) {
-	response := AuthResponse{}
-	response.Success = true
+	response := AuthResponse{true}
 
 	return &response, nil
 }
--- a/internal/api/users/auth.go
+++ b/internal/api/users/auth.go
@@ -1,0 +1,20 @@
+package users
+
+import (
+	"github.com/edulinq/autograder/internal/api/core"
+)
+
+type AuthRequest struct {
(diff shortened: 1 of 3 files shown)


internal/api/users/auth.go
```diff
@@ -0,0 +1,20 @@
...
+	core.APIRequestUserContext
+}
+
+type AuthResponse struct {
+	Success bool `json:"success"`
+}
+
+func HandleAuth(request *AuthRequest) (*AuthResponse, *core.APIError) {
+	response := AuthResponse{}
+	response.Success = true
```
eriq-augustine, line 17: Go ahead and init Success on construction.

internal/api/users/auth_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestUsersAuth(test *testing.T) {
+	testCases := []struct {
+		email    string
+		pass     string
+		success  bool
+		expected AuthResponse
```
eriq-augustine, line 15: Just use a single boolean value instead of these two values.

eriq-augustine (changes requested):
Looking good.
Just some small things.

eriq-augustine (approved):
lgtm

Before this gets in, we will need:
1. The Python-side PR.
2. Merge in a new version of main to avoid the conflict.

PR #84, opened 2024-07-24
Added an API Endpoint for listing users.

Lucas-Ellenberger:
Implemented an API endpoint for listing all users on the server.


internal/api/core/user_server.go
```diff
@@ -54,3 +56,67 @@ func MustNewServerUserInfo(user *model.ServerUser) *ServerUserInfo {
...
+
+	if b == nil {
+		return -1
+	}
+
+	return strings.Compare(a.Email, b.Email)
+}
+
+// Get server user info from a generic map (like what an API response would have).
+func ServerUserInfoFromMap(data map[string]any) *ServerUserInfo {
```
eriq-augustine, line 93: Do we need this?
Do we also need `CourseUserInfoFromMap`?
(Normally you don't touch things not related to your PR, but I'll give special permission for this.)

internal/api/users/list.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type ListResponse struct {
+	Users []*core.ServerUserInfo `json:"users"`
+}
+
+func HandleList(request *ListRequest) (*ListResponse, *core.APIError) {
+	response := ListResponse{}
+
+	users_map, err := db.GetServerUsers()
```
eriq-augustine, line 21: What's with the `snake_case`?

internal/api/users/list.go
```diff
@@ -0,0 +1,39 @@
...
+	core.APIRequestUserContext
+	core.MinServerRoleAdmin
+}
+
+type ListResponse struct {
+	Users []*core.ServerUserInfo `json:"users"`
+}
+
+func HandleList(request *ListRequest) (*ListResponse, *core.APIError) {
+	response := ListResponse{}
```
eriq-augustine, line 19: Create this closer to where it is used.
(In general (to all programming languages that do not hoist), you want to declare things close to where they are used.)

internal/api/users/list_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"reflect"
+	"slices"
+	"testing"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/model"
+)
+
+func TestList(test *testing.T) {
+	expectedUsers := []core.ServerUserInfo{
```
eriq-augustine, line 13: Note that you are testing the API endpoint `users/list`.
You are NOT testing the functionality of pulling users from the database (leave that to `internal/db` to test).

Simplify this test to assess the API endpoint instead of the DB.

internal/api/users/list_test.go
```diff
@@ -0,0 +1,135 @@
...
+		{
+			Email:   "server-user@test.com",
+			Name:    "server-user",
+			Role:    model.GetServerUserRole("user"),
+			Courses: map[string]core.EnrollmentInfo{},
+		},
+	}
+
+	response := core.SendTestAPIRequest(test, core.NewEndpoint(`users/list`), nil)
+
```
eriq-augustine, line 111: Bad whitespace.

internal/api/users/list_test.go
```diff
@@ -0,0 +1,135 @@
...
+	if !response.Success {
+		test.Fatalf("Response is not a success: '%v'.", response)
+	}
+
+	responseContent := response.Content.(map[string]any)
```
eriq-augustine, line 116: Any reason you are deserializing the response like this instead of the standard way in all the other API tests?
https://github.com/edulinq/autograder-server/blob/main/internal/api/users/get_test.go#L66

internal/api/users/list_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	rawUsers := responseContent["users"].([]any)
+	actualUsers := make([]core.ServerUserInfo, 0, len(rawUsers))
+
+	for _, rawUser := range rawUsers {
+		actualUsers = append(actualUsers, *core.ServerUserInfoFromMap(rawUser.(map[string]any)))
+	}
+
+	slices.SortFunc(expectedUsers, core.CompareServerUserInfo)
+	slices.SortFunc(actualUsers, core.CompareServerUserInfo)
```
eriq-augustine, line 130: You should not sort both your actual and expected results (in this case).
We expect the API response to be sorted, so we want to error if it is not.

internal/api/users/list_test.go
```diff
@@ -0,0 +1,72 @@
...
+	users := make([]*model.ServerUser, 0, len(usersMap))
+	for _, user := range usersMap {
+		users = append(users, user)
+	}
+
+	testCases := []struct {
+		email     string
+		pass      string
+		permError bool
+		users     []*model.ServerUser
```
eriq-augustine, line 25: Since the users will always be the same, you can get rid of this field.

internal/api/users/list_test.go
```diff
@@ -0,0 +1,72 @@
...
+	usersMap := db.MustGetServerUsers()
+
+	users := make([]*model.ServerUser, 0, len(usersMap))
+	for _, user := range usersMap {
+		users = append(users, user)
+	}
+
+	testCases := []struct {
+		email     string
+		pass      string
```
eriq-augustine, line 23: This is a minor change, but since a user's name is always their password,
you can just use `email` in the test case and then lookup the name/password in `usersMap`.

internal/api/users/list_test.go
```diff
@@ -0,0 +1,72 @@
...
+		permError bool
+		users     []*model.ServerUser
+	}{
+		// Invalid permissions.
+		{"server-user@test.com", "server-user", true, nil},
+		{"server-creator@test.com", "server-creator", true, nil},
+
+		// Valid permissions.
+		{"server-admin@test.com", "server-admin", false, users},
+		{"admin@test.com", "admin", false, users},
```
eriq-augustine, line 33: I think you can leave out `admin@test.com`.
Since they are mainly used in a course context, people may not know right away what their server permissions are.
Instead, how about you just use all the server-* users (which I think would just mean adding `server-owner`).

eriq-augustine (changes requested):
Looking good.
Only very minor notes.

You should be good to go for the Python PR whenever you are ready.

eriq-augustine (approved):
lgtm

internal/api/core/user.go
```diff
@@ -0,0 +1,171 @@
...
+			Name:  user.GetDisplayName(),
+		},
+		Role:    user.Role,
+		Courses: make(map[string]EnrollmentInfo, len(user.CourseInfo)),
+	}
+
+	for courseID, courseInfo := range user.CourseInfo {
+		course, err := db.GetCourse(courseID)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to get course (%s) for user (%s).", courseID, user.Email)
```
OliverLok, line 54: Instead of saying (%s), I think it may be better to say '%s', as that format is consistent throughout the repo. Same goes for the rest.

internal/api/core/user.go
```diff
@@ -0,0 +1,171 @@
...
+func CompareServerUserInfo(a ServerUserInfo, b *ServerUserInfo) int {
+	return strings.Compare(a.Email, b.Email)
+}
+
+// An API-safe representation of a course user.
+type CourseUserInfo struct {
+	UserInfo
+	Role  model.CourseUserRole `json:"role"`
+	LMSID string               `json:"lms-id"`
+}
```
OliverLok, line 126: Types should probably be at the top of the file

internal/api/core/user.go
```diff
@@ -0,0 +1,171 @@
...
+	info, err := NewServerUserInfo(user)
+	if err != nil {
+		log.Fatal("Failed to convert server user to API info.", err, user)
+	}
+
+	return info
+}
+
+func NewServerUserInfos(users []*model.ServerUser) ([]*ServerUserInfo, error) {
+	infos := make([]*ServerUserInfo, 0, len(users))
```
OliverLok, line 77: Needs a new line.

internal/api/core/user.go
```diff
@@ -0,0 +1,171 @@
...
+		},
+		Role:  user.Role,
+		LMSID: user.GetLMSID(),
+	}
+
+	return info
+}
+
+func NewCourseUserInfos(users []*model.CourseUser) []*CourseUserInfo {
+	result := make([]*CourseUserInfo, 0, len(users))
```
OliverLok, line 143: Needs a new line.

internal/api/core/user_test.go
```diff
@@ -0,0 +1,230 @@
...
+					UserInfo: UserInfo{
+						Type:  ServerUserInfoType,
+						Email: "admin@test.com",
+						Name:  "admin",
+					},
+					Role: model.GetServerUserRole("admin"),
+					Courses: map[string]EnrollmentInfo{
+						"course-languages": {
+							CourseID:   "course-languages",
+							CourseName: "Course Using Different Languages.",
```
OliverLok, line 41: Is there a reason CourseName is the only one that uses a period?

internal/api/courses/routes.go
```diff
@@ -0,0 +1,18 @@
+package courses
+
+// All routes handled by this package.
```
OliverLok, line 3: Maybe say "All the API endpoints handled by this package." since that's what the other files say.

internal/api/courses/users/list.go
```diff
@@ -0,0 +1,29 @@
...
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleGrader
+	Users core.CourseUsers `json:"-"`
+}
+
+type ListResponse struct {
+	Users []*core.CourseUserInfo `json:"users"`
+}
+
+func HandleList(request *ListRequest) (*ListResponse, *core.APIError) {
```
OliverLok, line 18: One of the return types is an APIError but the only error that gets returned is nil. Is this on purpose?

internal/api/courses/users/list_test.go
```diff
@@ -0,0 +1,74 @@
...
+		email     string
+		permError bool
+	}{
+		// Invalid permissions.
+		{"other@test.com", true},
+		{"student@test.com", true},
+
+		// Valid permissions.
+		{"admin@test.com", false},
+		{"owner@test.com", false},
```
OliverLok, line 35: Maybe include grader in here? I'm not 100% sure but I'd assume they would have invalid permissions.

internal/api/server/users/list_test.go
```diff
@@ -0,0 +1,70 @@
...
+				}
+			} else {
+				test.Errorf("Case %d: Response is not a success when it should be: '%v'.", i, response)
+			}
+
+			continue
+		}
+
+		if testCase.permError {
+			test.Errorf("Case %d: Did not get an expected permissions error.", i)
```
OliverLok, line 56: Consider having an expected/actual here for more clarify.

OliverLok (changes requested):
Already looks pretty polished. Just some minor things.

internal/api/courses/users/list.go
Lucas-Ellenberger: All Handler's return a core.APIError, so it is here to maintain consistency.

internal/api/server/users/list_test.go
Lucas-Ellenberger: The statement tells us that we expected an error but did not get it, so there is an expected/actual in the language. This error message is consistent with similar tests, like [users/get_test.go](https://github.com/edulinq/autograder-server/blob/e45720fb2e3b3493e50c98ac04244ec886a72dc4/internal/api/users/get_test.go#L61).

internal/api/courses/users/list.go
OliverLok: Gotcha ok you can resolve this then.

internal/api/server/users/list_test.go
OliverLok: Ok you can resolve this then.

OliverLok:
lgtm

OliverLok (approved):
lgtm!

eriq-augustine (changes requested):
Looking pretty good.
There's some work to be done, but the core is good.

@OliverLok, also make sure to read these comments closely.

eriq-augustine (approved):
lgtm

PR #85, opened 2024-08-01
Autograder daemon with unix socket support for CMD requests 

OliverLok:


.ci/verify_py_test_data.sh
```diff
@@ -24,6 +24,7 @@ function verify_test_data() {
 
 function run() {
     cd "${BASE_DIR}"
+    rm -f /tmp/autograder.pid /tmp/autograder.sock
```
eriq-augustine, line 27: These paths will not always be correct (they assume the default config value).

eriq-augustine (changes requested):
A good start.

We also need some test cases.

internal/api/server.go
```diff
@@ -1,18 +1,167 @@
...
+		log.Fatal("Failed to read size of the request buffer.", err)
+	}
+
+	size := binary.BigEndian.Uint64(sizeBuffer)
+
+	jsonBuffer := make([]byte, size)
+	_, err = conn.Read(jsonBuffer)
+	if err != nil {
+		log.Error("Failed to read the request from the buffer.", err)
+	}
```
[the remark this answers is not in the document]
OliverLok: Would making a new function inside of util make sense?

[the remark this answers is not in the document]
eriq-augustine: No, I don't think so.
No one else would use it, but the point of util is being a common place that anyone can use.

internal/common/pid_sock_status.go
```diff
@@ -0,0 +1,83 @@
...
+
+	unixFileNumber, err := util.RandHex(UNIX_SOCKET_RANDNUM)
+	if err != nil {
+		return "", fmt.Errorf("Failed to generate a random number for the unix socket path.")
+	}
+
+	unixSocketPath := filepath.Join("/tmp", fmt.Sprintf("autograder-%s.sock", unixFileNumber))
+	statusJson.UnixSocketPath = unixSocketPath
+
+	err = util.ToJSONFile(statusJson, statusPath)
```
eriq-augustine, line 77: I'm not a fan of the ordering here.
The file is created, and then this function comes in later to re-write the file with the socket path.

How about just creating the socket path when the stats file is first created.
We don't need to create the actual socket file,
so we can just generate the random number and reserve the path for later use.

internal/db/disk/user.go
```diff
@@ -22,6 +22,10 @@ func (this *backend) GetCourseUsers(course *model.Course) (map[string]*model.Cou
 
 	courseUsers := make(map[string]*model.CourseUser)
 	for email, user := range users {
+		if email == "root" {
```
eriq-augustine, line 25: Note the special case with a comment.

eriq-augustine, line 25: Use a constant to refer to root's email.

internal/db/user_test.go
```diff
@@ -22,6 +22,13 @@ func (this *DBTests) DBTestUserGetServerUsersBase(test *testing.T) {
 		test.Fatalf("Could not get server users: '%v'.", err)
 	}
 
+	_, exists := users["root"]
```
eriq-augustine, line 25: Explain this check and delete in a comment.

cmd/peek-submission/main.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+		TargetSubmission: args.TargetSubmission,
+	}
+
+	requestMap := map[string]interface{}{
+		"endpoint": core.NewEndpoint(`courses/assignments/submissions/fetch/user/peek`),
+		"request":  request,
+	}
+
+	jsonRequest := util.MustToJSONIndent(requestMap)
```
eriq-augustine, line 68: This is fine for now.
Note that it will need to transitioned to a non-Must function when generalized.

internal/procedures/server/start.go
```diff
@@ -0,0 +1,57 @@
...
+	defer api.StopServers()
+
+	log.Info("Autograder Version.", log.NewAttr("version", util.GetAutograderFullVersion()))
+
+	err := common.WriteAndHandlePidStatus()
+	if err != nil {
+		return err
+	}
+
+	db.MustOpen()
```
eriq-augustine, line 26: Since this code has been moved to a more general location, it can no longer use Must functions.

internal/util/unix_socket_rw.go
```diff
@@ -0,0 +1,47 @@
+package util
+
+import (
+	"bytes"
+	"encoding/binary"
+	"net"
+)
+
+const BUFFER_SIZE = 8
```
eriq-augustine, line 9: Inaccurate name.
This is only for the size component of the communication.

cmd/peek-submission/main.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		log.Fatal("Failed to load config options.", err)
+	}
+
+	socketPath, err := common.GetUnixSocketPath()
+	if err != nil {
+		log.Fatal("Failed to get the unix socket path.", err)
+	}
+
+	conn, err := net.Dial("unix", socketPath)
```
eriq-augustine, line 42: I prefer "connection" over "conn".
We have plenty of space.

eriq-augustine (changes requested):
Nice.
There's some real strong code here.

I have several comments (mainly in `internal/api/server.go`) that talks about how some small things are off with the http/unix sockets/servers (as far as design and language goes).
I think you need to think about this a little differently.
There is the "autograder server" or "server".
The server has an http socket which is responsible for serving the API,
and a unix socket which is responsible for local connections.
You may need to rename or move around some stuff accordingly.

eriq-augustine (changes requested):
I made some comments based on my previous comments,
but have not done a full review.

internal/common/pid.go
```diff
@@ -8,34 +8,32 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
-func CheckAndHandlePid() bool {
+func checkAndHandleStalePid() bool {
```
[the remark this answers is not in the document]
OliverLok: I had it as logging an error because you made the suggestion earlier to log it and return a bool. If we're now going to be returning an error, would saying "another instance of the autograder is already running" be a valid error on line 37? I'm only asking because you mentioned we shouldn't say that and just let the client deal with it.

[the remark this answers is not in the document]
eriq-augustine: > I'm not a fan of most the logs happening in here.

That's why I said "most" of the logging.
The warn is fine, but we don't want to over-generalize.

I believe you need the two returns, because there are three cases here:
1. An error happened (err != nil).
2. Another instance of the server is up. (false, nil)
3. No other server is up, we are good to go. (true, nil)

internal/util/unix_socket_rw.go
```diff
@@ -6,28 +6,28 @@ import (
...
+	_, err = connection.Read(jsonBuffer)
 	if err != nil {
 		return nil, err
 	}
 
 	return jsonBuffer, nil
 }
 
-func WriteToUnixSocket(conn net.Conn, data []byte) error {
+func WriteToUnixSocket(connection net.Conn, data []byte) error {
```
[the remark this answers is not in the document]
OliverLok: size is based off the size of the payload though so how could it be too large?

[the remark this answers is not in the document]
eriq-augustine: They could be sending files?
(An option on the server would have to be set to allow for this.)

Also, it's a simple error check to do, so it's nice to future proof things.

docs/development.md
```diff
@@ -130,6 +130,8 @@ Here are three common regex standards you may encounter, along with key consider
 
 A server instance is defined by its working directory,
 so anything that wants to interact with a server (including a cmd) needs to make sure they share the same working directory.
-This can be done by setting the base directory in the command line whenever interacting with the server (-c dirs.base).
-e.g. go run cmd/peek-submission/main.go course-student@test.edulinq.org course101 hw0 -c dirs.base="${WORKING_DIRECTORY}"
+This can be done by setting the base directory in the command line whenever interacting with the server (-c dirs.base). Here's an example:
```
eriq-augustine, line 133: I don't like "Here's an example",
it sounds too informal (not that we are trying to be extra formal).
I would just go with "For example:".

internal/api/core/request.go
```diff
@@ -17,9 +17,10 @@ import (
 // Once validated, callers should feel safe calling reflection methods on this without extra checks.
 type ValidAPIRequest any
 
-// A random nonce is generated for each root user request (CMDs),
-// gets stored in RootUserNonces, and gets attached to the request.
-// It's later validated on the server to confirm the request came from a valid root user.
+// A random nonce is generated for each root user request (CMDs).
+// The nonce is stored in RootUserNonces and is attached to the request.
+// It's later validated when processing the request through the http socket
```
eriq-augustine, line 22: Mention that it is deleted as soon as it is checked.

internal/util/netmsg.go
```diff
@@ -27,8 +34,16 @@ func ReadFromUnixSocket(connection net.Conn) ([]byte, error) {
...
 }
 
-func WriteToUnixSocket(connection net.Conn, data []byte) error {
+// Write a message to a network connection.
+// First 8 bytes: the size of the message being written.
+// Remaining bytes: the message content.
+func WriteToNetworkConnection(connection net.Conn, data []byte) error {
 	size := uint64(len(data))
+
+	if size > MAX_FORM_MEM_SIZE_BYTES {
```
eriq-augustine, line 43: This is incorrect.

The limit on the size of the message comes from the length of the size value written as the first thing in the connection.
You can make it a constant, but it should not be related to the max form size.
The max form size is a constant right now, but should probably be a config option.
Whereas the max socket message size should just be a large fixed constant.
Just make it something like 2GB and define it here.

Also, defining it as a constant also allows you to reject messages that are too large on the read side as well.

internal/common/pid_sock_status.go
```diff
@@ -24,16 +27,23 @@ func GetStatusPath() string {
 	return filepath.Join(config.GetWorkDir(), STATUS_FILENAME)
 }
 
-func WriteAndHandleStatusFile() error {
+func WriteAndHandleStatusFile() (err error) {
 	Lock(PID_SOCK_LOCK)
-	defer Unlock(PID_SOCK_LOCK)
+	defer func() {
+		err = errors.Join(err, Unlock(PID_SOCK_LOCK))
```
eriq-augustine, line 33: No need to handle errors on the Unlock().

You know how strict I am about errors,
but Unlock() is one of the cases that an error would be so unheard of
(and only if we messed up),
that it's fine to skip.

internal/common/pid_sock_status.go
```diff
@@ -24,16 +27,23 @@ func GetStatusPath() string {
...
-	if !checkAndHandleStalePid() {
-		return fmt.Errorf("Failed to check and handle the pid.")
+	ok, err := checkAndHandleStalePid()
+	if !ok {
+		if err != nil {
+			return err
+		}
+
+		return fmt.Errorf("Failed to create the status file.")
 	}
```
eriq-augustine, line 47: Handle these in two distinct conditionals.
(This is common golang style.)

First check `(err != nil)`, then check `!ok`.

internal/common/pid_sock_status.go
```diff
@@ -42,19 +52,51 @@ func WriteAndHandleStatusFile() error {
 	if err != nil {
 		return fmt.Errorf("Failed to generate a random number for the unix socket path: '%w'.", err)
 	}
-	statusJson.UnixSocketPath = fmt.Sprintf("/tmp/autograder-%s.sock", unixFileNumber)
+	statusJson.UnixSocketPath = filepath.Join("/tmp", fmt.Sprintf("autograder-%s.sock", unixFileNumber))
```
eriq-augustine, line 55: Still a bit off.
"/tmp" are two path elements ("/" and "tmp") that need to be joined.

internal/common/pid_sock_status.go
```diff
@@ -42,19 +52,51 @@ func WriteAndHandleStatusFile() error {
...
 
 	err = util.ToJSONFile(statusJson, statusPath)
 	if err != nil {
-		return fmt.Errorf("Failed to write the pid to the status file: '%w'.", err)
+		return fmt.Errorf("Failed to write to the status file: '%w'.", err)
```
eriq-augustine, line 59: Include the path in this error.
(This applies to most errors (and the log) on the status file where you have the path available.)

internal/common/pid_sock_status.go
```diff
@@ -42,19 +52,51 @@ func WriteAndHandleStatusFile() error {
...
 	}
 
-	return nil
+	return err
 }
 
-func GetUnixSocketPath() (string, error) {
-	Lock(PID_SOCK_LOCK)
-	defer Unlock(PID_SOCK_LOCK)
+func checkAndHandleStalePid() (bool, error) {
```
eriq-augustine, line 65: Include a comment discussing the return semantics.

internal/common/pid_sock_status.go
```diff
@@ -42,19 +52,51 @@ func WriteAndHandleStatusFile() error {
...
+	statusPath := GetStatusPath()
+
+	if !util.IsFile(statusPath) {
+		return true, nil
+	}
+
+	var statusJson StatusInfo
+
```
eriq-augustine, line 73: Remove this newline.
(Just a nit, but I think it is cleaner to declare it and fill it together.)

internal/common/pid_sock_status.go
```diff
@@ -42,19 +52,51 @@ func WriteAndHandleStatusFile() error {
...
+	err := util.JSONFromFile(statusPath, &statusJson)
+	if err != nil {
+		return false, fmt.Errorf("Failed to read the status file: '%w'.", err)
+	}
+
+	process, _ := os.FindProcess(statusJson.Pid)
```
eriq-augustine, line 79: Make a function (isAlive()?) that handles this.
It's not a lot of code, but it's confusing enough that it would be good to box it off.
Then you can also include a function-level comment on it.

internal/common/pid_sock_status.go
```diff
@@ -42,19 +52,51 @@ func WriteAndHandleStatusFile() error {
...
+	err = process.Signal(syscall.Signal(0))
+	if err != nil {
+		log.Warn("Removing stale status file.")
+
+		err := util.RemoveDirent(GetStatusPath())
```
eriq-augustine, line 84: Why not `statusPath`?

internal/common/pid_sock_status.go
```diff
@@ -42,19 +52,51 @@ func WriteAndHandleStatusFile() error {
...
+	var statusJson StatusInfo
+
+	err := util.JSONFromFile(statusPath, &statusJson)
+	if err != nil {
+		return false, fmt.Errorf("Failed to read the status file: '%w'.", err)
+	}
+
+	process, _ := os.FindProcess(statusJson.Pid)
+	err = process.Signal(syscall.Signal(0))
+	if err != nil {
```
eriq-augustine, line 81: Invert.

In general, you want to try and return as early as you can.
This allows readers to worry about one less case as they progress through your code.

internal/util/netmsg.go
eriq-augustine, line 1: This filenames does not really fit with our other filenames
(we typically use fill words with underscores).
In another project, this name would fit in just fine.

I think just something like "socket.go", "net.go", or "connection.go" is sufficient.
Note that since this is a filename and not a package name,
the stakes are not high at all.

internal/util/netmsg.go
```diff
@@ -3,12 +3,19 @@ package util
...
 
-const RESPONSE_BUFFER_SIZE = 8
+const (
+	MAX_FORM_MEM_SIZE_BYTES = 10 << 20 // 20 MB
+	RESPONSE_BUFFER_SIZE    = 8
+)
 
-func ReadFromUnixSocket(connection net.Conn) ([]byte, error) {
+// Read a message from a network connection.
+// First 8 bytes: the size of the message being read.
```
eriq-augustine, line 16: This needs to be a tab bit more precise.
Specifically, you don't say if the initial size includes the 8 bytes for the size itself.

Also, I'm not a fan of the colons here.
I would go with something like: "The first 8 bytes is ..., and the remaining content should be x bytes of ... .".

eriq-augustine (changes requested):
Getting closer.

This is still not a full review,
just comments on previously changed code.

cmd/peek-submission/main.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/edulinq/autograder/internal/util"
+)
+
+var args struct {
+	config.ConfigArgs
+	TargetEmail      string `help:"Email of the user to fetch." arg:""`
+	CourseID         string `help:"ID of the course." arg:""`
+	AssignmentID     string `help:"ID of the assignment." arg:""`
+	TargetSubmission string `help:"ID of the submission. Defaults to the latest submission." arg:"" optional:""`
+	ShortForm        bool   `help:"Use short form output." long:"short-form"`
```
eriq-augustine, line 25: I would change it to just something like `--short`.

As a side note, doesn't Kong get that same long form from the field's name
(I think it observes casing).

internal/util/exit.go
eriq-augustine, line 1: I do not like this.
This is not a patter we generally do, as it does not offer a lot of flexibility or guarantees.
(Anyone can just override our "function".)

Make an actual function that calls os.Exit.
Log the exit to DEBUG first.
Have two variables that live in this file:
 - One that controls whether or not to actually exit.
   Tests can set this to avoid a hard exit.
 - One that keeps track of the last exit code (which we will always set prior to exiting).

Your tests can just override the exit flag (defer setting it back to normal),
and call a function to check the variable for the last exit status.

cmd/peek-submission/main_test.go
```diff
@@ -0,0 +1,92 @@
...
+			continue
+		}
+
+		if len(stderr) > 0 {
+			test.Errorf("Case %d: CMD has content in stderr: '%s'.", i, stderr)
+			continue
+		}
+
+		if exitCode != testCase.expectedExitCode {
+			test.Errorf("Unexpected exit code. Expected: %d, Actual: %d", testCase.expectedExitCode, exitCode)
```
eriq-augustine, line 63: Case number is not listed.

cmd/peek-submission/main_test.go
```diff
@@ -0,0 +1,92 @@
...
+			continue
+		}
+
+		if testCase.expectedExitCode != 0 {
```
eriq-augustine, line 67: There is still output on an error.

internal/procedures/server/start.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/edulinq/autograder/internal/log"
+	"github.com/edulinq/autograder/internal/model"
+	pcourses "github.com/edulinq/autograder/internal/procedures/courses"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func Start() (err error) {
+	defer server.StopServer()
+
+	log.Info("Autograder Version.", log.NewAttr("version", util.GetAutograderFullVersion()))
```
eriq-augustine, line 19: Downgrade all the log.Infos in this file (**except this one**) to debug.

internal/api/server/server.go
```diff
@@ -0,0 +1,58 @@
...
+	"os/signal"
+	"syscall"
+
+	"github.com/edulinq/autograder/internal/api"
+	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+// Run the autograder server and listen on an http and unix socket.
+func StartServer() (err error) {
```
eriq-augustine, line 15: Make this something to indicate that we are not just starting and leaving, but waiting as well.
Maybe "RunServer" or "StartServerAndListen" like in http, or something like that.

eriq-augustine (changes requested):
Looking sooo good.
There is just a little bit left.

You need to also look into why the CI tests are failing.
I can understand failing to validate the Python data,
but your tests should not be failing there.

internal/util/exit.go
```diff
@@ -1,5 +1,29 @@
...
+import (
+	"os"
 
-var Exit = os.Exit
+	"github.com/edulinq/autograder/internal/log"
+)
+
+var (
+	ShouldExit   = true
+	LastExitCode = 0
```
eriq-augustine, line 11: Don't export this variable.

internal/util/exit.go
```diff
@@ -1,5 +1,29 @@
...
-import "os"
+import (
+	"os"
 
-var Exit = os.Exit
+	"github.com/edulinq/autograder/internal/log"
+)
+
+var (
+	ShouldExit   = true
```
eriq-augustine, line 10: Make the name something that makes it super obvious that normal people should not be using this.
I usually add "for testing" to then end of the name (with matching casing/style).
For example:
https://github.com/edulinq/autograder-server/blob/main/internal/util/dir.go#L15

internal/util/exit.go
```diff
@@ -1,5 +1,29 @@
...
+	LastExitCode = 0
+)
+
+func Exit(code int) {
+	SetExitCode(code)
+
+	if ShouldExit {
+		log.Debug("Exiting with code.", log.NewAttr("code", code))
```
eriq-augustine, line 18: Move this outside the conditional (so it always logs).

eriq-augustine, line 18: You can just say "Exiting" (since all exits always have a code (even if it is zero)).

cmd/peek-submission/main_test.go
```diff
@@ -22,30 +22,38 @@ func TestPeekBase(test *testing.T) {
...
 
+	util.ShouldExit = false
+	defer func() {
+		util.ShouldExit = true
+	}()
+
+	oldExitCode := util.GetLastExitCode()
+	defer func() {
+		util.SetExitCode(oldExitCode)
+	}()
```
eriq-augustine, line 48: No need to track/replace the last exit code.
We just need it for the test cases.

internal/util/exit.go
```diff
@@ -1,5 +1,29 @@
...
+		os.Exit(code)
+	}
+}
+
+func GetLastExitCode() int {
+	return LastExitCode
+}
+
+func SetExitCode(code int) {
```
eriq-augustine, line 27: We don't want anyone outside setting this variable.

cmd/peek-submission/main_test.go
```diff
@@ -22,30 +22,38 @@ func TestPeekBase(test *testing.T) {
...
 
-		{"course-student@test.edulinq.org", "ZZZ", "hw0", "", "", 2},
-		{"course-student@test.edulinq.org", "course101", "ZZZ", "", "", 2},
-		{"course-student@test.edulinq.org", "ZZZ", "ZZZ", "", "", 2},
+		{"course-student@test.edulinq.org", "ZZZ", "hw0", "", "", 2, "-018", "error"},
+		{"course-student@test.edulinq.org", "course101", "ZZZ", "", "", 2, "-022", "error"},
+		{"course-student@test.edulinq.org", "ZZZ", "ZZZ", "", "", 2, "-018", "error"},
 	}
 
+	util.ShouldExit = false
```
eriq-augustine, line 40: Add a comment on this.

cmd/peek-submission/main_test.go
```diff
@@ -22,30 +22,38 @@ func TestPeekBase(test *testing.T) {
...
+	defer func() {
+		util.ShouldExit = true
+	}()
```
eriq-augustine, line 43: I don't think this is the right place for this.
Instead, I'm thinking that `cmd.RunCMDTest()` could do this automatically and return the exit status.
Then all CMD tests can get free access to it.

cmd/peek-submission/main_test.go
```diff
@@ -22,30 +22,38 @@ func TestPeekBase(test *testing.T) {
...
 			testCase.targetEmail,
 			testCase.courseID,
 			testCase.assignmentID,
 			testCase.targetSubmission,
-		}
-
-		var exitCode int
-		util.Exit = func(code int) {
-			exitCode = code
+			"--log-level", testCase.logLevel,
```
eriq-augustine, line 56: Why does the log level need to be set?
Can't we just set the level to fatal for this test (and thereby ignore all the logs)?

Or maybe even do it in internal/cmd, like we do in internal/db:
https://github.com/edulinq/autograder-server/blob/main/internal/db/test.go#L31

eriq-augustine (changes requested):
Nice, soooo close.
Just some small (and mainly test) things.

Nice root check and tests.

eriq-augustine (approved):
LGTM!

PR #86, opened 2024-08-02
Fix for the api locator scripts on mac

OliverLok:


scripts/get_max_api_locators.sh
```diff
@@ -23,7 +23,7 @@ function main() {
 
         local largestLocator=$(grep -R '("-' "${path}" | sed 's/^.*"\(-[0-9]\{3\}\)".*$/\1/' | sort | uniq | tail -n 1)
         # Remove input zero padding (so bash does not think the numebr is octal.
-        local cleanLargestLocator=$(echo "${largestLocator}" | sed 's/-0\+/-/g')
+        local cleanLargestLocator=$(echo "${largestLocator}" | sed 's/^-0*/-/g')
```
eriq-augustine, line 26: The anchor makes sense, but I think plus makes sense here more than star.
Because of the previous line they will both work the same,
but I think the star more clearly shows our intentions.
One of the harder things to really wrap your head around when writing maintainable code is creating code that not only correctly does what you want, but that also expresses your intention.

scripts/check_duplicate_api_locators.sh
```diff
@@ -18,7 +18,7 @@ function main() {
 
     cd "${ROOT_DIR}"
 
-    local duplicateLocators=$(grep -R 'Error("-' "${API_DIR}" | sed 's/^.*"\(-[0-9]\{3,\}\)".*$/\1/' | sort | uniq -c | grep -ZvP '^\s+1\s+' | sed 's/^\s*\([0-9]\+\)\s\+\(-[0-9]\{3,\}\)$/\2/')
+    local duplicateLocators=$(grep -R 'Error("-' "${API_DIR}" | sed 's/^.*"\(-[0-9]\{3,\}\)".*$/\1/' | sort | uniq -c | grep -EZv '^[[:space:]]*1[[:space:]]+' | sed 's/^\s*\([0-9]\+\)\s\+\(-[0-9]\{3,\}\)$/\2/')
```
eriq-augustine, line 21: I think plus would be better here than star (similar reasons with the other comment).

eriq-augustine (changes requested):
Nice work.
I have some small changes.

Aside from those, we should probably start to build up a doc that discusses do/don't of cross-platform shell scripting.
We already have a place to put it (until it grows too large):
https://github.com/edulinq/autograder-server/blob/main/docs/development.md

Note that I'm not asking for a full tutorial on cross-platform scripting,
just some a starting place that people can easily extend later.
(Also, I couldn't even find a tutorial on cross-platform scripting with a simple search, so maybe this will eventually be that.)
It should note that our targets are generally POSIX systems,
but specifically general Linux and Mac systems.
I'm thinking things like what tools we can use with what options (this would be where you tell people they can use `grep`, but with some specific options).

scripts/get_max_api_locators.sh
OliverLok: After playing around with it more, it turns out the only thing both sed's needed was a -E flag. Should I still include the anchor even though it works without it now? Also, did you mean to say the plus more clearly shows our intentions?

docs/development.md
```diff
@@ -35,3 +35,15 @@ Since this is very unlikely and most times an abs path is not strictly necessary
 "Should" functions can be used anywhere,
 but care should be taken not to overuse them when a fallback value is not sufficient.
 "Must" variants should be favored in testing code, since they will more obviously fail tests.
+
+## Do/Don't for cross-platform shell scripting
```
eriq-augustine, line 39: Use Title Case for Section Titles

eriq-augustine, line 39: There is some general structural work for this section that I am going to put here in a single comment.
Overall, we want this section to be more friendly/accessible.

Start with changing the title to something like "Cross-Platform Scripting".

Then, ass in an introduction paragraph that says something about why this section needs to exist.
Like "we are writing bash scripts with the intention of being run in a POSIX environment, but there are still some differences you need to be aware of ...". 

Then, break it up by tools.
For example, `-P`/`-E` are specifically for grep.

eriq-augustine (changes requested):
Code is looking good.
Just needs some work on the documentation.

docs/development.md
```diff
@@ -35,3 +35,20 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+
+## Cross-Platform Scripting
+
+We are writing bash scripts with the intention of running them in a POSIX environment. However, differences between operating systems, particularly between Linux and MacOS, can still affect a script's behavior. This section highlights key considerations to ensure your scripts run consistently across platforms.
+
+### Grep
+
+The `-P` flag, which enables perl-compatible regular expressions, does not work consistently across operating systems. More specifically, while it may work as intended on Linux, it behaves differently on Mac. This can lead to unexpected results when run on MacOS. To avoid inconsistency across operating systems, avoid using the `-P` flag. Sometimes, using the `-E` flag, which enables extended regular expressions, may produce the same output that `-P` was going for. 
+
+### Sed and Regular Expression
```
eriq-augustine, line 47: Since regular expressions will cross multiple commands, we should have a section (probably the first subsection) just for them.
There you can talk about BRE/ERE.

docs/development.md
```diff
@@ -35,3 +35,20 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+
+When using regular expressions in shell commands with tools like `sed`, there are differences between how Linux and MacOS handle them due to differences between basic regular expressions and extended regular expressions.
+
+* **BRE (Basic Regular Expressions)**: Requires escaping certain metacharacters such as `+` to be able to interpret them as special characters. 
+* **ERE (Extended Regular Expressions)**: Does not require escaping these metacharacters since they are treated as special characters by default.
```
eriq-augustine, line 52: You should link a reference to these (I just favor Wikipedia pages).

We also probably don't need to give full definitions.
Just the name, abbreviation, link to reference material, (maybe a short description), and then tell them which one they should be using.

docs/development.md
```diff
@@ -35,3 +35,20 @@ Since this is very unlikely and most times an abs path is not strictly necessary
 "Should" functions can be used anywhere,
 but care should be taken not to overuse them when a fallback value is not sufficient.
 "Must" variants should be favored in testing code, since they will more obviously fail tests.
+
+## Cross-Platform Scripting
+
+We are writing bash scripts with the intention of running them in a POSIX environment. However, differences between operating systems, particularly between Linux and MacOS, can still affect a script's behavior. This section highlights key considerations to ensure your scripts run consistently across platforms.
```
eriq-augustine, line 41: I tend to put sentences/clauses on lines by themselves so that diffs come out easier to read and work with (and less likely to cause merge conflicts).

docs/development.md
```diff
@@ -35,3 +35,20 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+
+### Grep
```
eriq-augustine, line 43: Most official documentation I see uses undercase for sed and grep.

docs/development.md
```diff
@@ -35,3 +35,20 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+
+The `-P` flag, which enables perl-compatible regular expressions, does not work consistently across operating systems. More specifically, while it may work as intended on Linux, it behaves differently on Mac. This can lead to unexpected results when run on MacOS. To avoid inconsistency across operating systems, avoid using the `-P` flag. Sometimes, using the `-E` flag, which enables extended regular expressions, may produce the same output that `-P` was going for.
```
eriq-augustine, line 45: s/perl/Perl/

docs/development.md
```diff
@@ -35,3 +35,20 @@ Since this is very unlikely and most times an abs path is not strictly necessary
 "Should" functions can be used anywhere,
 but care should be taken not to overuse them when a fallback value is not sufficient.
 "Must" variants should be favored in testing code, since they will more obviously fail tests.
+
+## Cross-Platform Scripting
```
eriq-augustine, line 39: All the information here is good, but I think we need to change up the style so it is better for people looking at it quickly as reference material.
They will usually just want to look up things quickly and not read a paragraph.

I would take a look at Google Style guides for reference.
The [C++ guide](https://google.github.io/styleguide/cppguide.html) is the gold standard, 
but I think it is a bit detailed for our case.
The [Golang guide](https://google.github.io/styleguide/go/best-practices) would probably be a closer (but not exact) match.

eriq-augustine (changes requested):
Code is looking great.
Content in the doc is also looking great, but we will want to change up the style/presentation.

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+
+We are writing bash scripts with the intention of running them in a POSIX environment.
+However, differences between operating systems, particularly between Linux and MacOS,
+can still affect a script's behavior. 
+This section highlights key considerations to ensure your scripts run consistently across platforms.
+
+### Regular Expressions
```
eriq-augustine, line 46: Missing newline after header.
See full PR.

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
 "Should" functions can be used anywhere,
 but care should be taken not to overuse them when a fallback value is not sufficient.
 "Must" variants should be favored in testing code, since they will more obviously fail tests.
+
+## Cross-Platform Scripting
+
+We are writing bash scripts with the intention of running them in a POSIX environment.
+However, differences between operating systems, particularly between Linux and MacOS,
```
eriq-augustine, line 42: Instead of Linux and MacOS, a more correct pairing would be Linux and BSD (which MacOS) is based on.
Note that this applies to the whole PR.

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+This section highlights key considerations to ensure your scripts run consistently across platforms.
+
+### Regular Expressions
+Differences between how Linux and MacOS handle regular expressions can cause inconsistent outcomes when writing scripts.
+Be aware of the following:
+
+- **BRE (Basic Regular Expressions)**
+  - **Usage**: Default in many tools.
+  - **Consideration**: Limited feature set.
+  - **[More Info](https://en.wikibooks.org/wiki/Regular_Expressions/Basic_Regular_Expressions)**
```
eriq-augustine, line 53: There is something missing here.
If this was fully true, then we would just always use BREs.
The functionality may be "limited" (I would never say that about the core set of Regex features),
but they are almost always all we need (easily 98% of cases would be solved with BREs).

So, why don't we just only use BREs?

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+## Cross-Platform Scripting
+
+We are writing bash scripts with the intention of running them in a POSIX environment.
+However, differences between operating systems, particularly between Linux and MacOS,
+can still affect a script's behavior. 
+This section highlights key considerations to ensure your scripts run consistently across platforms.
+
+### Regular Expressions
+Differences between how Linux and MacOS handle regular expressions can cause inconsistent outcomes when writing scripts.
+Be aware of the following:
```
eriq-augustine, line 48: This sentence needs work.
It is not informative.

You are trying to introduce different standard types of regular expressions,
so say something like that.

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+
+- **BRE (Basic Regular Expressions)**
+  - **Usage**: Default in many tools.
+  - **Consideration**: Limited feature set.
+  - **[More Info](https://en.wikibooks.org/wiki/Regular_Expressions/Basic_Regular_Expressions)**
+
+- **ERE (Extended Regular Expressions)**
+  - **Usage**: Enabled with `-E` flag in tools like `grep` and `sed`.
```
eriq-augustine, line 56: I would soften this statement up a bit, since you can't say it is `-E` in every (or even most) tools.
Maybe something like: "Often enabled with the `-E` flag (like in `grep` and `sed`).

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+  - **Consideration**: More expressive than BRE.
+  - **[More Info](https://en.wikibooks.org/wiki/Regular_Expressions/POSIX-Extended_Regular_Expressions)**
+
+- **PCRE (Perl-Compatible Regular Expressions)**
+  - **Usage**: Enabled with `-P` flag in `grep`.
```
eriq-augustine, line 61: s/ flag/ the flag/

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+
+### Tool-Specific Guidelines
+
+#### `grep`
+- **Avoid**: `-P` flag (PCRE) for cross-platform scripts.
+- **Alternative**: Use `-E` flag (ERE) to achieve similar functionality with more consistent behavior.
+
+#### `sed`
+- **Use**: `-E` flag to enable ERE.
+- **Avoid**: BRE syntax that may not behave consistently across platforms (e.g., `\+` in BRE).
```
eriq-augustine, line 73: This doesn't matter, since we are telling people to always use `-E`.

eriq-augustine (changes requested):
Looking pretty good.

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+This section highlights key considerations to ensure your scripts run consistently across platforms.
+
+### Regular Expressions
+
+Differences between how Linux and BSD handle different standards of regular expressions can cause inconsistent outcomes when writing scripts.
+Here are three common regex standards you may encounter, along with key considerations for each:
+
+- **BRE (Basic Regular Expressions)**
+  - **Usage**: Default in many tools.
+  - **Consideration**: While BREs cover most common use cases, they lack more advanced features could simplify and lower the chance of compatibility issues with complex regular expressions.
```
eriq-augustine, line 53: I still don't think this statement is accurate.
I think the real issue is just that the default behavior is just less consistent,
i.e. the differences we saw when using `+` and sed.
`+` isn't in BRE, but sed accepted it on Linux, but not on Mac.

Just say that it is simple and inconsistently used as a default.

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+  - **[More Info](https://en.wikibooks.org/wiki/Regular_Expressions/Basic_Regular_Expressions)**
```
eriq-augustine, line 54: I would prefer wikipedia links of wikibooks links.
The wikibooks one is more concise, but wikipedia often has more relevant information,
e.g. https://en.wikipedia.org/wiki/Regular_expression#IEEE_POSIX_Standard .

docs/development.md
```diff
@@ -35,3 +35,39 @@ Since this is very unlikely and most times an abs path is not strictly necessary
...
+  - **[More Info](https://en.wikibooks.org/wiki/Regular_Expressions/Perl-Compatible_Regular_Expressions)**
+
+### Tool-Specific Guidelines
+
+#### `grep`
+- **Avoid**: The `-P` flag (PCRE) for cross-platform scripts.
+- **Alternative**: Use the `-E` flag (ERE) to achieve similar functionality with more consistent behavior.
+
+#### `sed`
+- **Use**: The `-E` flag to enable ERE.
```
eriq-augustine, line 73: Note that BRE is inconsistent here so that they should always use ERE.

eriq-augustine (changes requested):
Getting closer.

eriq-augustine (approved):
lgtm!

PR #87, opened 2024-08-06
Made a bash assignment for no-docker tests

OliverLok:


internal/grader/test.go
```diff
@@ -41,6 +42,10 @@ func GetTestSubmissions(baseDir string) ([]*TestSubmissionInfo, error) {
 			return nil, fmt.Errorf("Could not find assignment for test submission '%s': '%w'.", testSubmissionPath, err)
 		}
 
+		if assignment.ID != "bash" && config.DOCKER_DISABLE.Get() {
```
eriq-augustine, line 45: Needs a comment.

testdata/course-languages/bash/grader.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# Compile the project and run the grader.
```
eriq-augustine, line 3: Old comment.

testdata/course-languages/bash/grader.sh
```diff
@@ -0,0 +1,72 @@
...
+    if [[ $? -ne 0 ]] ; then
+        echo "ERROR: Failed to run grader."
+        return 3
+    fi
+
+    return 0
+}
+
+function grade() {
+    source "${THIS_DIR}/assignment.sh"
```
eriq-augustine, line 36: Needs a comment.

testdata/course-languages/bash/test-submissions/not-implemented/assignment.sh
```diff
@@ -0,0 +1,3 @@
+function add() {
+    echo "-1"
+}
```
eriq-augustine, line 3: Missing newline (not an additional newline, just the newline that all files should end with).
Same in the other test submission.

testdata/course-languages/bash/grader.sh
```diff
@@ -0,0 +1,72 @@
...
+
+    # Allow the grader to run locally by changing the output location
+    # if not in the docker image.
+    local outputPath="${DEFAULT_OUTPUT_PATH}"
+    if [[ ! -d $(dirname "${outputPath}") ]] ; then
+        outputPath="$(basename "${outputPath}")"
+    fi
+    
+    # Run grader.
+    grade > "${outputPath}"
```
eriq-augustine, line 26: Instead of echoing the output in the function and redirecting it here,
I would pass in `outputPath` to the function and write it out in `grade()`.
In this way, a student can't echo in their own code and mess up the grader.

internal/grader/test.go
eriq-augustine, line 45: I don't think `config.DOCKER_DISABLE` would be the thing to check here.
What about a user that is running the tests on a machine where docker is runner?
We want them the have the same experience as someone who does not have docker running (i.e., the same set of tests should run for TestNoDockerSubmissions in both cases).

Instead, this function should probably take a boolean (`useDocker`?) that tells this function if we want those other test cases.

eriq-augustine (approved):
lgtm

PR #88, opened 2024-08-14
New API: courses/users/list

Lucas-Ellenberger:
This PR must be reviewed with the corresponding autograder-py PR. This PR should be reviewed after the PR for users/list.

~~This PR adds a new role 'CourseRoleSuper' which cannot be held by a normal user within a course (it will be a validation error). The role is reserved for users with serverRoleAdmin or above that are using course specific APIs. These server admins are granted a higher role to have power over the courseRoleOwner role. This distinction is important for actions that cannot be performed against users with an equal role (like the future courses/users/drop).~~ (Edit: Outdated)

The PR also adds a new API endpoint for listing course users: courses/users/list.

Note: There are change that overlap with the PR for users/list as they are essential for both PRs to function, hence why the users/list PR should be reviewed and merged first.

Also, the run_tests.sh script is flaky and randomly fails to create the container that is necessary for the 'TestDockerSubmissions' test. Specifically, it has a chance to fail on 'course-languages/bash/test-submissions/solution/test-solution/test-submission.json'.


internal/api/core/user.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// An API-safe representation of a server user.
+type ServerUserInfo struct {
+	UserInfo
+	Role    model.ServerUserRole      `json:"role"`
+	Courses map[string]EnrollmentInfo `json:"courses"`
+}
+
+type EnrollmentInfo struct {
```
OliverLok, line 34: Consider adding a comment here too.

internal/api/core/user.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+
+	return info, nil
+}
+
+func MustNewServerUserInfo(user *model.ServerUser) *ServerUserInfo {
+	info, err := NewServerUserInfo(user)
+	if err != nil {
+		log.Fatal("Failed to convert server user to API info.", err, user)
```
OliverLok, line 77: Use log.NewAttr("user",  user) or something to list the user instead to maintain consistency.

internal/api/core/user_test.go
```diff
@@ -0,0 +1,230 @@
...
+	for i, testCase := range testCases {
+		if !testCase.toCourseUser {
+			serverUserInfos := make([]*ServerUserInfo, 0, len(serverUsers))
+			for _, user := range serverUsers {
+				serverUserInfos = append(serverUserInfos, MustNewServerUserInfo(user))
+			}
+
+			for _, serverUserInfo := range serverUserInfos {
+				if serverUserInfo.Type != ServerUserInfoType {
+					test.Errorf("Test %d: Unexpected user info type. Expected: '%s', actual: '%s'.", i, ServerUserInfoType, serverUserInfo.Type)
```
OliverLok, line 185: Say unexpected server user info type for more clarity.

internal/api/core/user_test.go
```diff
@@ -0,0 +1,230 @@
...
+			}
+
+			courseUserInfos := make([]*CourseUserInfo, 0, len(courseUsers))
+			for _, courseUser := range courseUsers {
+				courseUserInfos = append(courseUserInfos, NewCourseUserInfo(courseUser))
+			}
+
+			for _, courseUserInfo := range courseUserInfos {
+				if courseUserInfo.Type != CourseUserInfoType {
+					test.Errorf("Test %d: Unexpected user info type. Expected: '%s', actual: '%s'.", i, CourseUserInfoType, courseUserInfo.Type)
```
OliverLok, line 216: Say unexpected course user info type for more clarity.

internal/model/user_server.go
```diff
@@ -165,6 +165,28 @@ func (this *ServerUser) ToCourseUser(courseID string) (*CourseUser, error) {
...
+	courseUser, err := this.ToCourseUser(courseID)
+	if err != nil {
+		return courseUser, err
+	}
+
+	if courseUser == nil {
+		courseUser = &CourseUser{
+			Email: this.Email,
+			Name:  this.Name,
+			Role:  CourseRoleSuper,
```
OliverLok, line 180: You could probably remove this since the role is being added on a few lines down anyway.

OliverLok (changes requested):
Looking good, just small things.

internal/api/core/user.go
OliverLok: Still unresolved.

internal/api/core/user.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if b == nil {
+		return -1
+	}
+
+	return strings.Compare(a.Email, b.Email)
+}
+
+func CompareServerUserInfo(a ServerUserInfo, b *ServerUserInfo) int {
```
OliverLok, line 124: Is there a reason b is a pointer here but in CompareCourseUserInfo, none of them are pointers?

internal/api/core/user_test.go
```diff
@@ -0,0 +1,232 @@
...
+	"reflect"
+	"slices"
+	"testing"
+
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestNewUserInfos(test *testing.T) {
```
OliverLok, line 13: Consider adding some more cases. Ex. testing if the user isn't enrolled, testing server admin/owner -> course owner, etc.

OliverLok (changes requested):
Looking good, just a couple things.

internal/api/courses/users/list_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestList(test *testing.T) {
+	course := db.MustGetCourse(db.TEST_COURSE_ID)
+	usersMap, err := db.GetCourseUsers(course)
+	if err != nil {
+		test.Fatalf("Unable to get course users.")
```
OliverLok, line 17: Add the error to the end of the message

internal/api/courses/users/list.go
OliverLok, line 1: I noticed that listing server users and listing course users now share the same function/type names (ListRequest, ListResponse, HandleList, TestList. This seems like it could be problematic especially for TestList if both are the same name and one fails, it may be hard to know which one failed.

OliverLok (commented):
Looking solid just a couple things.

internal/api/courses/users/list.go
Lucas-Ellenberger: We can tell them apart by the packages that each function/type is in. When testing, go shows the package that is being tested so you can tell which function is failing.

We even already have this overlap with the API endpoint for listing assignments in a course, which can be found [here](https://github.com/edulinq/autograder-server/blob/main/internal/api/courses/assignments/list.go).

OliverLok: Ok, I just thought the more clarity the better but if it's already being done then you can resolve this.

OliverLok (approved):
lgtm!

internal/api/courses/users/list_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestList(test *testing.T) {
```
eriq-augustine, line 13: Add a test case for an empty course.
You would have to empty out a course and use a server admin to do it.
I would just make another function for it (since it doesn't fit with the pattern here).

internal/api/courses/users/list.go
```diff
@@ -0,0 +1,29 @@
...
+
+func HandleList(request *ListRequest) (*ListResponse, *core.APIError) {
+	users := make([]*model.CourseUser, 0, len(request.Users))
+	for _, user := range request.Users {
+		users = append(users, user)
+	}
+
+	infos := core.NewCourseUserInfos(users)
+
+	response := ListResponse{infos}
```
eriq-augustine, line 26: (nit)

I would just embed directly into to response (instead of first putting it inside a variable).

internal/api/courses/users/list_test.go
```diff
@@ -0,0 +1,79 @@
...
+	course := db.MustGetCourse(db.TEST_COURSE_ID)
```
eriq-augustine, line 14: (nit)

I would use `db.MustGetTestCourse()` instead.

internal/api/courses/users/list_test.go
```diff
@@ -0,0 +1,79 @@
...
+	usersMap, err := db.GetCourseUsers(course)
```
eriq-augustine, line 15: (nit)

Put a newline above this (separate getting the course and users).

eriq-augustine (changes requested):
Looks very good.

I'm just looking for one more test case,
and I put in some style nits (but I would have let those in if I didn't want the extra test case).

eriq-augustine (changes requested):
Just one thing.

eriq-augustine (approved):
lgtm

PR #89, opened 2024-08-18
New Role: CourseRoleSuper. Server admin can access any course level command.

Lucas-Ellenberger:
This PR adds a new role 'CourseRoleSuper' which cannot be held by a normal user within a course (it will be a validation error). The role is reserved for users with serverRoleAdmin or above that are using course specific APIs. These server admins are granted a higher role to have power over the courseRoleOwner role. This distinction is important for actions that cannot be performed against users with an equal role (like the future courses/users/drop).


docs/development.md
```diff
@@ -44,6 +44,17 @@ Cleartext is never stored in anything other than memory (and event then it will
 When we are working with a "password", it should always be a [SHA-256](https://en.wikipedia.org/wiki/SHA-2) hash of the cleartext.
 Any variables/arguments actually using cleartext should make that apparent with the variable name (i.e. the name should have "cleartext" in it).
 
+### Test Users
+
+All test users that are server-role@test.com are used to test the corresponding ServerRoles.
```
eriq-augustine, line 49: Not in code, we just say "server roles" (same for course roles).

docs/development.md
```diff
@@ -44,6 +44,17 @@ Cleartext is never stored in anything other than memory (and event then it will
...
+These users are not enrolled in any test courses by default.
+
+The role@test.com users denote the corresponding CourseRole.
```
eriq-augustine, line 52: This name is bad.
I assume you are referring to the users like "student@test.com",
but your sentence is indicating that there is a literal "role@test.com" user.

docs/development.md
```diff
@@ -44,6 +44,17 @@ Cleartext is never stored in anything other than memory (and event then it will
 When we are working with a "password", it should always be a [SHA-256](https://en.wikipedia.org/wiki/SHA-2) hash of the cleartext.
 Any variables/arguments actually using cleartext should make that apparent with the variable name (i.e. the name should have "cleartext" in it).
 
+### Test Users
```
eriq-augustine, line 47: This information is fine (and we can leave it in),
but it is not what we are looking for here.
You need to describe the semantics of how a server admin is treated when turned into a course user.

docs/development.md
```diff
@@ -44,6 +44,17 @@ Cleartext is never stored in anything other than memory (and event then it will
...
+
+All test users that are server-role@test.com are used to test the corresponding ServerRoles.
+These users are not enrolled in any test courses by default.
+
+The role@test.com users denote the corresponding CourseRole.
+To test courses actions, these users are enrolled in various test courses.
+All of these users are given ServerRoleUser to simplify test cases.
```
eriq-augustine, line 54: This is not true.
They are given the standard role that most users on a server (and almost every user enrolled in a course) will have.

docs/development.md
```diff
@@ -44,6 +44,17 @@ Cleartext is never stored in anything other than memory (and event then it will
...
+The exception to this rule is owner@test.com, who is also a ServerRoleOwner.
```
eriq-augustine, line 55: What about tests that need a course owner, but not a server owner?
You should probably just add another user instead.

internal/model/user_server.go
```diff
@@ -160,10 +160,22 @@ func (this *ServerUser) GetCourseRole(courseID string) CourseUserRole {
 }
 
 // Convert this server user into a course user for the specific course.
+// Set courseOnly to false if high level server users should be converted to a course user.
 // Will return (nil, nil) if the user is not enrolled in the given course.
-func (this *ServerUser) ToCourseUser(courseID string) (*CourseUser, error) {
+func (this *ServerUser) ToCourseUser(courseID string, courseOnly bool) (*CourseUser, error) {
```
eriq-augustine, line 165: The idea of adding in a parameter here is good, but I think it should be inverted.
I think that `courseOnly` is not intuitive,
and it should be inverted into `escalateServerAdmin` or something like that.

Changing the name can also hint in a change of semantics where we can escalate server admins who are enrolled in a course to course owners.
This would change it so that escalation only happens on request.

internal/model/user_server.go
```diff
@@ -160,10 +160,22 @@ func (this *ServerUser) GetCourseRole(courseID string) CourseUserRole {
...
 	info, exists := this.CourseInfo[courseID]
```
eriq-augustine, line 166: Swap out `exists` with `enrolled`.
`exists` is fine for general cases, but we are checking something specific in this case.

internal/model/user_server.go
```diff
@@ -160,10 +160,22 @@ func (this *ServerUser) GetCourseRole(courseID string) CourseUserRole {
...
 	if !exists {
+		if !courseOnly && this.Role >= ServerRoleAdmin {
```
eriq-augustine, line 168: I'm not a fan of the current conditional structure,
it has more corners than I would like.

Instead you can store whether or not to escalate in a variable.
Then you can have something like:
```
if (!enrolled && !escalate) {
    return nil, nil;
}

courseUser := ... basic info ...

if (enrolled) {
    // Set course info.
}

if (escalate) {
    // Set role.
}

return courseUser, nil;

File: internal/procedures/users/upsert_test.go [hunk 9]
```diff
@@ -382,42 +382,42 @@ func TestUpsertUser(test *testing.T) {
 			options: UpsertUsersOptions{
 				RawUsers: []*model.RawUserData{
 					&model.RawUserData{
-						Email: "admin@test.com",
```
eriq-augustine, line 385: Can you not use "server-admin@test.com" instead of "owner@test.com"?

That applies to a lot of these tests.

internal/api/courses/assignments/list_test.go
```diff
@@ -10,25 +10,49 @@ import (
...
-		{"course-without-source"},
-		// One Assignment
-		{"course101-with-zero-limit"},
-		// Multiple Assignments
-		{"course-languages"},
+		// Invalid permissions.
+		{"server-user@test.com", true, "course-without-source"},
+		{"server-creator@test.com", true, "course-without-source"},
+
+		// Valid permissions, Empty
```
eriq-augustine, line 24: No need to repeat "Valid Permissions" on each comment.
Just make one comment denoting this "section" of tests.

eriq-augustine (changes requested):
The core design here looks good.

I feel like there are a lot of tests missing here.
Or rather, modifications to existing tests to add in a user that will be escalated.
I understand that you are short on time,
so I will compromise here (something I RARELY do when it comes to code quality).
In the conversation tab on this PR, I want you to list all the tests that should be expanded with an escalated user (e.g. tests that could throw a perm error (I would just look for API endpoints that have a APIRequestCourseUserContext or APIRequestAssignmentContext based request).
Your next change after the hack-a-thon must be a PR that augments all those tests.
I recommend changing core.SendTestAPIRequest* to take in an email instead of a course role.
If it sees an email without a password,
it knows to look up a password using the user's name.

docs/development.md
Lucas-Ellenberger: I will add users with a "course-" prefix for each role in course roles. All of these test users will be server users. I will convert the original (non-prefixed) test users (i.e. admin@test.com) to be varying server roles, like they were before. This will take too long to create and fix test cases, so it will accompany my next PR that expands test cases.

Lucas-Ellenberger:
The following APIs need more test cases:

CourseUserContext:
- admin/logs_fetch.go
- admin/update_course
- lms/user_get
- lms/upload_scores
- courses/assignments/list

AssignmentContext:
- submissions/fetch_scores
- submissions/fetch_submissions
- submissions/submit
- submissions/remove_submissions
- submissions/peek
- submissions/fetch_submissions
- submissions/history
- submissions/fetch_attempts
- courses/assignments/get

The list of APIs will change following the refactoring of these APIs and their packages. There will be new APIs that will need tests as well. After expanding the test data, I will need to update a lot of test cases to work properly.

docs/development.md
eriq-augustine: I did not ask for that and that change is unrelated to this PR.
This documentation is already unrelated to your PR, what you are proposing here is to add more to your PR based on documentation that shouldn't be in your PR.

eriq-augustine (approved):
lgtm

PR #91, opened 2024-08-20
New API Endpoint: users/pass/reset

Lucas-Ellenberger:
Added a new API endpoint, users/pass/reset, that allows users to reset their password if they forgot their login.


eriq-augustine (approved):
lgtm

PR #93, opened 2024-08-21
Improvements to the testing infrastructure.

Lucas-Ellenberger:
This PR improves the testing infrastructure in 3 ways.

The PR removes test user role ambiguity by prefixing all test data users with "server-" or "course-". This change ensures test cases are clearly testing server or course role permissions.

The `SendTestAPIRequestFull()` requires a user email is provided instead of a role. This allows test cases to easily test the API with any of our test users. If an email for a known user passed to `SendTestAPIRequestFull()`, the function automatically unpacks their password via the users name. Callers are still able to overwrite the user-email and user-pass that are sent to the API via the fields parameter. The default user for testing API requests is "course-admin@test.edulinq.org" which is used by `SendTestAPIRequest`.

Test cases are expanded to test role escalation for many CourseUserContext and AssignmentContext APIs. In cases where perm errors are expected, test cases include the expected locator to differentiate between invalid role escalation and invalid course role permissions.


docs/development.md
```diff
@@ -44,16 +44,35 @@ Cleartext is never stored in anything other than memory (and event then it will
...
+- course-with-zero-limit: Tests courses without a submission limit.
+- course-without-source: Tests courses without a source.
+
 ### Test Users
 
-All test users that are prefixed with "server-" (i.e. "server-owner@test.com") are used to test the corresponding server roles.
+Test users allow testing of actions made by the various role users can hold on the server.
+By default, the name and password of test users are their email without "@test.edulinq.org".
+
+All test users that are prefixed with "server-" (i.e. "server-owner@test.edulinq.org") are used to test the corresponding server roles.
```
eriq-augustine, line 67: s/All test users that are prefixed/Users prefixed/

This keeps it symmetric with the course users part.

docs/development.md
```diff
@@ -44,16 +44,35 @@ Cleartext is never stored in anything other than memory (and event then it will
...
 
-Test users without a prefix (i.e. "admin@test.com") denote users with the corresponding course role.
-To test courses actions, these users are enrolled in various test courses.
+Users prefixed with "course-" (i.e. "course-owner@test.edulinq.org") are used to test the corresponding course roles.
 All of these users are given the standard server role, which is server user.
-The exception to this rule is owner@test.com, who is also a server owner.
-This allows testing of server admin commands for users that are also enrolled in the course.
+To test courses actions, these users are enrolled in various test courses.
+
+The user "no-lms-id@test.edulinq.org" breaks the pattern and is a server user and a course admin.
```
eriq-augustine, line 74: Do we need this user in the test data.
If they are just used on one test,
can we just make that user in that test?

internal/api/core/test.go
```diff
@@ -52,18 +51,22 @@ func APITestingMain(suite *testing.M, routes *[]*Route) {
...
+func SendTestAPIRequestFull(test *testing.T, endpoint string, fields map[string]any, paths []string, email string) *APIResponse {
 	url := serverURL + endpoint
 
-	email := role.String() + "@test.com"
-	pass := util.Sha256HexFromString(role.String())
+	user := db.MustGetServerUser(email, true)
+
+	pass := ""
+	if user != nil {
+		pass = util.Sha256HexFromString(*user.Name)
```
eriq-augustine, line 68: Use `user.GetName(false)` to prevent nil pointer.

eriq-augustine (changes requested):
This is **real strong** work, good job.

Good catch on some of the typos.

I have an idea that could work or be annoying,
so you let me know what you think.
For `SendTestAPIRequestFull()`, what if when we see an email that doesn't have an "@" in it, we just append "@test.edulinq.org".
Then test makers wouldn't have to repeat it so much.
Do you think this could be useful, or just too much?

eriq-augustine (approved):
lgtm!

PR #94, opened 2024-08-29
New API Endpoint: Users Upsert

Lucas-Ellenberger:
This PR adds a new API endpoint, users/upsert.

The upsert test cases are basic because the upsert behavior is thoroughly tested in `internal/procedures/users/upsert_test.go`.


internal/model/user_op.go
```diff
@@ -36,66 +38,146 @@ type UserOpResult struct {
...
+	SystemError *LocatableError
+
+	// The following error occurred during this operation, but not because of the provided data,
+	// i.e., the system was unable to communicate the results.
+	// These errors are not guarenteed to be safe for users,
+	// and the calling code should decide how they should be managed.
+	CommunicationError *LocatableError
 
 	// The following cleartext password was generated during this operation.
 	// Care should be taken to not expose their field.
```
eriq-augustine, line 67: s/their/this/

I know this is not your text, but go ahead and change it while you are in here.
Thanks.

scripts/get_max_locators.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+readonly THIS_DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd | xargs realpath)"
+readonly ROOT_DIR="${THIS_DIR}/.."
+readonly INTERNAL_DIR="${ROOT_DIR}/internal"
+
+function process_directory() {
+    local path="$1"
+
+    local dir=$(basename "${path}")
```
eriq-augustine, line 10: Call this `dirname`.
Like "filename" we usually use "dirname" when the value is a name (instead of a path) and then "dir" when it is a path.
This is a pretty loose convention, but nice when done consistently.

scripts/get_max_locators.sh
```diff
@@ -0,0 +1,57 @@
...
+    local parent_dir=$(basename "$(dirname "${path}")")
+
+    local largestLocator=$(grep -RoP '("\-[0-9]{3,4}")' "${path}" 2>/dev/null | sed 's/^.*"\(-[0-9]\{3,4\}\)".*$/\1/' | sort | uniq | tail -n 1)
```
eriq-augustine, line 13: See `documentation/development.md` on advice on grep flags (we want to use `-E` over `-P`).
You may also want to use `-h` in combination with `-o` (which you already use).

scripts/get_max_locators.sh
```diff
@@ -0,0 +1,57 @@
...
+    cd "${ROOT_DIR}"
+
+    for dir in "${INTERNAL_DIR}"/* ; do
+        local found=0
+        for path in "${dir}"/* ; do
+            if [[ -d "${path}" ]]; then
+                process_directory "${path}"
+                if [[ $? -eq 1 ]]; then
+                    found=1
+                fi
```
eriq-augustine, line 45: Instead of the conditional, how about just:
```
(( found += $? ))
```

scripts/get_max_locators.sh
```diff
@@ -0,0 +1,57 @@
...
+            fi
+        done
+
+        if [[ ${found} -eq 0 ]]; then
```
eriq-augustine, line 49: Put a space between the `]]` and `;`.

We (mainly my fault) are inconsistent about this, but we should tighten it up.

eriq-augustine (changes requested):
Looking great!
Just a few small things.

docs/development.md
```diff
@@ -71,6 +71,30 @@ Users prefixed with "course-" (i.e. "course-owner@test.edulinq.org") are used to
...
+If a package you are working on requires locatable errors, the package gets the next chunk of 1000 locators.
+Update this document to include the new top-level allocation.
+
+### Top-Level Package Allocations
+
+|Package    |Upper Bound |Lower Bound |
+|-----------|------------|------------|
+|API        |-001        |-999        |
+|Procedures |-1000       |-1999       |
+|LMS        |-2000       |-2999       |
```
eriq-augustine, line 90: Use the package's native casing (like how it is on disk (lower case)).

docs/development.md
```diff
@@ -71,6 +71,30 @@ Users prefixed with "course-" (i.e. "course-owner@test.edulinq.org") are used to
...
+
+### API Errors
+
+All API errors are locatable errors.
+We allocate 3 digit negative numbers (-001 to -999) as the locators for all API errors.
+Within this range, each package in the API package is given a range of locators.
```
eriq-augustine, line 96: s/range of locators/range of 100 locators/

eriq-augustine (changes requested):
Just two small things.

eriq-augustine (approved):
lgtm

PR #95, opened 2024-09-06
Added an API endpoint to remove a user from the server.

Lucas-Ellenberger:
Added users/remove which handles removing a user from the server.

### DIFF
--- a/internal/api/users/remove.go
+++ b/internal/api/users/remove.go
@@ -1,0 +1,40 @@
+package users
+
+import (
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/db"
+)
+
+type RemoveRequest struct {
+	core.APIRequestUserContext
+	core.MinServerRoleAdmin
+
+	TargetUser core.TargetServerUser `json:"target-email"`
+}
+
+type RemoveResponse struct {
+	FoundUser bool `json:"found-user"`
+}
+
+func HandleRemove(request *RemoveRequest) (*RemoveResponse, *core.APIError) {
+	response := RemoveResponse{}
+
+	if !request.TargetUser.Found {
+		return &response, nil
+	}
+
+	response.FoundUser = true
+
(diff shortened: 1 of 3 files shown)


internal/api/users/remove.go
```diff
@@ -0,0 +1,40 @@
...
+
+	if request.TargetUser.User.Role >= request.ServerUser.Role {
+		return nil, core.NewBadServerPermissionsError("-811", &request.APIRequestUserContext, request.TargetUser.User.Role,
+			"Cannot remove a user with an equal or higher role.").Add("target-user", request.TargetUser.User.Email)
+	}
+
+	_, err := db.DeleteUser(request.TargetUser.Email)
+	if err != nil {
+		return nil, core.NewUsertContextInternalError("-812", &request.APIRequestUserContext,
+			"Failed to remove user.").Err(err).Add("target-user", request.TargetUser.Email)
```
eriq-augustine, line 36: Use `request.TargetUser.User.Email` like in this other error (-811).
This is a nit, but `request.TargetUser.User.Email` should have been validated and potentially normalized.

internal/api/users/remove_test.go
```diff
@@ -0,0 +1,89 @@
...
+		var responseContent RemoveResponse
+		util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
+
+		if testCase.expected != responseContent {
+			test.Errorf("Case %d: Unexpected remove result. Expected: '%s', actual: '%s'.",
+				i, util.MustToJSONIndent(testCase.expected), util.MustToJSONIndent(responseContent))
+			continue
+		}
+
+		if !testCase.expected.FoundUser {
```
eriq-augustine, line 72: Instead of returning at this point, make sure that the target user still exists
(like on the permission errors).

eriq-augustine (changes requested):
Real strong work here.
Just one nit and one test change.

eriq-augustine (approved):
LGTM

PR #97, opened 2024-09-19
Reworked version functionality for non-git environments.

BatuhanSA:


cmd/version/main.go
```diff
@@ -25,6 +35,26 @@ func main() {
...
+
+		versionSplice := strings.Split(util.GetAutograderFullVersion(),"-")
+
+		version:= Version{
+			Short: util.GetAutograderVersion(),
+			Hash : versionSplice[1],
+			Status: versionSplice[2],
+		}
+
+		err := util.ToJSONFileIndentCustom(&version,versionJSONPath,""," ")
```
eriq-augustine, line 49: Why a custom indent here?

cmd/version/main.go
```diff
@@ -25,6 +35,26 @@ func main() {
...
+		if err != nil {
+			log.Error("Failed to write to the JSON file", err, log.NewAttr("path", versionJSONPath))
+	
+		}
+
+		return 
+	}
+
 	fmt.Println("Autograder")
```
eriq-augustine, line 58: Why this line here?
We probably don't need it.

cmd/version/main.go
```diff
@@ -25,6 +35,26 @@ func main() {
...
 	fmt.Printf("Short Version: %s\n", util.GetAutograderVersion())
 	fmt.Printf("Full  Version: %s\n", util.GetAutograderFullVersion())
```
eriq-augustine, line 60: I don't like how the API version is handled here instead of in util/version.
It would be good if they could be handled in the same place
(maybe the API version can also live in VERSION.json).

internal/util/version.go
```diff
@@ -10,48 +10,91 @@ import (
...
 	if !IsFile(versionPath) {
 		log.Error("Version file does not exist.", log.NewAttr("path", versionPath))
 		return UNKNOWN_VERSION
 	}
 
-	version, err := ReadFile(versionPath)
+	var readVersion Version
+	readVersion.Short = UNKNOWN_VERSION
+
+	err := JSONFromFile(versionPath,&readVersion)
```
eriq-augustine, line 34: I don't like how the file is read in two locations.
Just have one function that handles the file (computes path, logs, reads, etc) and return a `(*Version, nil)`.

internal/util/version.go
```diff
@@ -10,48 +10,91 @@ import (
...
 	if err != nil {
-		dirtySuffix = "-" + UNKNOWN_VERSION
+		status = UNKNOWN_VERSION
 	}
 
 	if isDirty {
-		dirtySuffix = "-" + DIRTY_SUFFIX
+		status = DIRTY_SUFFIX
+	}else {
+		status = "clean"
```
eriq-augustine, line 64: If you have two siffixes, then they should both be constants.

internal/util/version.go
```diff
@@ -10,48 +10,91 @@ import (
...
+	if readVersion.Hash == "" || readVersion.Status == ""{
+		shortVersion , gitHash, status = ComputeAutograderFullVersion()
+		return shortVersion + "-" + gitHash + "-" + status
+	}
+
+	shortVersion = readVersion.Short
+	gitHash = readVersion.Hash
+	status = readVersion.Status
+	
+	return shortVersion + "-" + gitHash + "-" + status
```
eriq-augustine, line 99: Clean commits should not have a suffix.

internal/util/version.go
```diff
@@ -10,48 +10,91 @@ import (
...
+	}
+
+	if readVersion.Hash == "" || readVersion.Status == ""{
+		shortVersion , gitHash, status = ComputeAutograderFullVersion()
+		return shortVersion + "-" + gitHash + "-" + status
+	}
+
+	shortVersion = readVersion.Short
+	gitHash = readVersion.Hash
+	status = readVersion.Status
```
eriq-augustine, line 97: I don't see the use of these variables, instead of just using the Version struct you already have.

eriq-augustine (changes requested):
The core stuff looks good.
There are a good amount of styling things.

internal/util/version.go
eriq-augustine, line 1: The function names in this file need work.

I suspect that several of them can be unexported.

Because you now how the struct, most people don't need to know about short vs long versions.
I suggest just having `GetVersion()` which returns the struct, and whatever specific access function that are required (like `MustGetAPIVersion()` (which is only required because it needs to be used in a static var declaration)).

internal/util/version.go
```diff
@@ -10,48 +11,116 @@ import (
...
+	version, err := ReadVersionFromJSON()
+	if err != nil {
+		log.Error("Failed to read the version from JSON file.")
+		return UNKNOWN_VERSION
+	}
+
+	return strings.TrimSpace(version.Short)
+}
+
+func ReadVersionFromJSON() (*Version, error) {
```
eriq-augustine, line 47: I am very suspect that this function needs to be exported.

internal/util/version.go
```diff
@@ -10,48 +11,116 @@ import (
...
+	VERSION_FILENAME string = "VERSION.json"
 	DIRTY_SUFFIX     string = "dirty"
+	UNKNOWN_API      int    = -1
 	HASH_LENGTH      int    = 8
 )
 
+type Version struct {
+	Short  string `json:"short-version"`
+	Hash   string `json:"git-hash"`
+	Status string `json:"status"`
```
eriq-augustine, line 23: Why is this a string?
Doesn't this just track a binary value (is dirty?)?

eriq-augustine (changes requested):
There are several things here that you really need to work on (aside from the code itself).

1. There are issues that I have already commented on in your previous round. Reviewers do not like to see mistakes repeated over-and-over again. You need to ensure that you don't repeat mistakes. If necessary, write down every comment on a piece of paper and review each function and check-off each comment for each function.
2. You have multiple places that are grossly inefficient. We are not trying to write super optimized code here, we are not even trying to write optimized code. But you have multiple places that calls multiple version functions (which does disk IO (reads a file)). When instead you could have just got your version struct (util.Version) and reference that multiple times.

eriq-augustine (changes requested):
Back to you.
(These were all discussed synchronously.)

cmd/version/main.go
```diff
@@ -25,8 +25,21 @@ func main() {
...
-	fmt.Printf("Full  Version: %s\n", util.GetAutograderFullVersion())
-	fmt.Printf("API   Version: %d\n", core.API_VERSION)
+	version, err := util.GetAutograderVersion()
+	if err != nil {
+		log.Fatal("Failed to get the autograder version.", err)
+	}
+
+	if args.Out == "" {
+		fmt.Printf("Short Version: %s\n", version.Short)
+		fmt.Printf("Full  Version: %s\n", util.Version.String(version))
```
eriq-augustine, line 35: Call the method normally, like in almost any language.

cmd/version/main.go
```diff
@@ -25,8 +25,21 @@ func main() {
...
+		fmt.Printf("API   Version: %d\n", version.Api)
+
+		return
+	}
+
+	err = util.ToJSONFileIndent(&version, args.Out)
```
eriq-augustine, line 41: Why the pointer?

internal/api/core/error.go
```diff
@@ -135,10 +135,15 @@ func (this *APIError) ToResponse() *APIResponse {
 		locator = ""
 	}
 
+	version, err := util.GetAutograderVersion()
+	if err != nil {
+		log.Error("Failed to get the autograder version.", err)
```
eriq-augustine, line 140: Downgrade this to a warning.

internal/procedures/server/start.go
```diff
@@ -16,7 +16,12 @@ import (
 func Start() (err error) {
 	defer server.StopServer()
 
-	log.Info("Autograder Version.", log.NewAttr("version", util.GetAutograderFullVersion()))
+	version, err := util.GetAutograderVersion()
+	if err != nil {
+		log.Error("Failed to get the autograder version.", err)
```
eriq-augustine, line 21: Downgrade to a warning.

internal/util/version.go
```diff
@@ -1,57 +1,92 @@
...
+func MustGetAPIVersion() int {
+	version, err := versionFromJSON()
+	if err != nil {
+		log.Fatal("Failed to read the version from JSON file.", err)
+	}
+
+	return version.Api
+}
+
+func versionFromJSON() (Version, error) {
```
eriq-augustine, line 44: I'm not a fan of this function name (it implies that you are passing some JSON information to the function).
Something like "readVersion()" would be better.

internal/util/version.go
```diff
@@ -1,57 +1,92 @@
...
+	var version Version
```
eriq-augustine, line 45: You are missing the initialization of this (you don't want the default empty value, you want a custom empty/default value).

internal/util/version.go
```diff
@@ -1,57 +1,92 @@
...
-		return UNKNOWN_VERSION
+		return version, err
 	}
 
-	return strings.TrimSpace(version)
+	return version, nil
 }
 
-func GetAutograderFullVersion() string {
+func getGitInfo() (string, bool) {
```
eriq-augustine, line 60: You should return an error here (instead of logging) as well.

internal/util/version.go
```diff
@@ -1,57 +1,92 @@
...
 	if err != nil {
-		dirtySuffix = "-" + UNKNOWN_VERSION
+		log.Error("Failed to get state of the git repository ", err, log.NewAttr("path", repoPath))
+		return hash[0:HASH_LENGTH], true
+	}
+
+	return hash[0:HASH_LENGTH], isDirty
+}
+
+func GetAutograderVersion() (Version, error) {
```
eriq-augustine, line 78: Take a lot of care here to try to put together as much of the version as you can (for example, you can still try to get the git information even if you couldn't read the file).
Just join together the errors (via [errors.Join()](https://pkg.go.dev/errors#Join)).

internal/util/version_test.go
```diff
@@ -1,20 +1,21 @@
...
-	version := GetAutograderFullVersion()
-	if strings.HasPrefix(version, UNKNOWN_VERSION) {
+	if version.Short == UNKNOWN_VERSION {
 		test.Fatalf("Did not get an actual version (check error log).")
 	}
+
+	if version.Api == UNKNOWN_API {
+		test.Fatalf("Did not get an actual API version (check error log).")
+	}
+
```
eriq-augustine, line 20: Extra newline.

eriq-augustine (changes requested):
Looking much better.

cmd/version/main.go
```diff
@@ -25,8 +25,21 @@ func main() {
 		log.Fatal("Could not load config options.", err)
 	}
 
-	fmt.Println("Autograder")
-	fmt.Printf("Short Version: %s\n", util.GetAutograderVersion())
-	fmt.Printf("Full  Version: %s\n", util.GetAutograderFullVersion())
-	fmt.Printf("API   Version: %d\n", core.API_VERSION)
+	version, err := util.GetAutograderVersion()
+	if err != nil {
+		log.Error("Failed to get the autograder version.", err)
```
eriq-augustine, line 30: This should be Fatal.
Just include the version in the log (as an attribute).

eriq-augustine (changes requested):
Nice, getting real close.

eriq-augustine (approved):
lgtm

PR #98, opened 2024-09-19
New API Endpoint: Courses Users Enroll

Lucas-Ellenberger:
This PR adds a new API endpoint, courses/users/enroll.

This endpoint is meant for course-admin to enroll users into their course.

This PR must be reviewed with it's associated python interface PR, which can be found [here](https://github.com/edulinq/autograder-py/pull/28).

This PR is dependent on the [users/upsert PR](https://github.com/edulinq/autograder-server/pull/94). The users/upsert PR must be merged before reviewing this PR.


internal/model/user_raw.go
```diff
@@ -16,6 +16,44 @@ type RawUserData struct {
 	CourseLMSID string `json:"course-lms-id" help:"LMS ID for the new user in the specified course."`
 }
 
+// Raw/dirty data for a course user.
+// This struct is used for raw data coming from a single course.
+type RawCourseUserData struct {
```
eriq-augustine, line 21: Since we have "RawCourseUserData",
we should probably rename "RawUserData" to "RawServerUserData".

internal/procedures/users/upsert.go
```diff
@@ -322,6 +334,12 @@ func checkCourseUpdatePermissions(newUser *model.ServerUser, oldUser *model.Serv
 			fmt.Errorf("User has a course role of '%s', which is not high enough to update a user with course role of '%s'.", options.ContextCourseRole.String(), oldCourseRole.String()))
 	}
 
+	// Cannot update a user to have a higher course role than the context user.
+	if options.ContextCourseRole < newCourseRole {
+		return model.NewUserOpResultValidationError("-1022", newUser.Email,
```
eriq-augustine, line 339: Let's get a test case for this.

internal/procedures/users/upsert.go
```diff
@@ -142,6 +149,11 @@ func insertUser(user *model.ServerUser, options UpsertUsersOptions, rawData *mod
 		result.Enrolled = append(result.Enrolled, courseID)
 	}
 
+	err = user.Validate()
+	if err != nil {
+		return model.NewUserOpResultValidationError("-1023", user.Email, err)
```
eriq-augustine, line 154: Let's get a test case for this.
I imagine you can probably just insert the user, but with a bad email or something.

eriq-augustine (changes requested):
Nice!
Looking real good with some tricky and subtle pieces.

internal/procedures/users/upsert.go
Lucas-Ellenberger: This error should never be returned. We perform a weak validation and return error code -1003 for all common validation errors. Since this validation check is the strong version, it could only catch unknown server roles (that wouldn't be caught be error -1003).

But, we explicitly set the role to be a known value before this validation.

I included this validation as a sanity check.

eriq-augustine: kk, sounds good.

eriq-augustine (approved):
lgtm

PR #99, opened 2024-09-19
Added an API endpoint to drop a user from a course.

Lucas-Ellenberger:
New API endpoint: courses/users/drop

This endpoint allows course admin or server admin to drop users from a course.

This PR must be reviewed with the associated python interface PR, which can be found [here](https://github.com/edulinq/autograder-py/pull/29).


internal/api/courses/users/drop_test.go
```diff
@@ -0,0 +1,111 @@
...
+	}{
+		// Valid permissions.
+		{"course-admin", "course-other@test.edulinq.org", false, "", DropResponse{true}},
+		{"course-admin", "course-student@test.edulinq.org", false, "", DropResponse{true}},
+		{"course-owner", "course-grader@test.edulinq.org", false, "", DropResponse{true}},
+		{"course-owner", "course-admin@test.edulinq.org", false, "", DropResponse{true}},
+
+		// Valid permissions, role escalation.
+		{"server-owner", "course-other@test.edulinq.org", false, "", DropResponse{true}},
+		{"server-owner", "course-student@test.edulinq.org", false, "", DropResponse{true}},
```
eriq-augustine, line 31: Change these two to `server-admin`.

internal/api/courses/users/drop.go
```diff
@@ -0,0 +1,41 @@
...
+func HandleDrop(request *DropRequest) (*DropResponse, *core.APIError) {
+	response := DropResponse{}
+
+	if !request.TargetCourseUser.Found {
+		return &response, nil
+	}
+
+	response.FoundUser = true
+
+	if (request.ServerUser.Role < model.ServerRoleAdmin) && (request.TargetCourseUser.User.Role >= request.User.Role) {
```
eriq-augustine, line 29: Make a function (checkPermissions()?) that handles this permissions check.
Then once there, handle this in two checks:
 1. Server admins can do what they want (return early). Put a comment here that we are letting a server admin drop a server owner, but that we don't really care since it does not change things on the server-level.
 2. If you are not a server admin, you need high enough relative permissions.

The code you have here is correct (and I would probably let it in if it was the only change I was considering for this PR),
but the nice thing about having it in another function is just how clear you can make it to other developers what is going on in it.
And since this is a potential security task, we want to be as clear as possible.

internal/api/courses/users/drop_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+		// Invalid permissions.
+		{"course-student", "course-other@test.edulinq.org", true, "-020", DropResponse{false}},
+		{"course-grader", "course-other@test.edulinq.org", true, "-020", DropResponse{false}},
+
+		// Invalid permissions, role escalation.
+		{"server-creator", "server-user@test.edulinq.org", true, "-040", DropResponse{false}},
+		{"server-user", "server-user@test.edulinq.org", true, "-040", DropResponse{false}},
+
+		// Target user is not found.
```
eriq-augustine, line 44: Add test for valid permission, but the user us not enrolled in the course.
I'm fine with the API response being the same as it is now (no error or warning),
since by the end of the API call the user will not be enrolled in the course (even if they were not enrolled before).

eriq-augustine (changes requested):
Looking pretty good.
I have just some test things and a readability nit.

internal/api/courses/users/drop.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if !request.TargetCourseUser.Found {
+		return &response, nil
+	}
+
+	response.FoundUser = true
+
+	if !checkDropPermissions(request) {
+		return nil, core.NewBadCoursePermissionsError("-612", &request.APIRequestCourseUserContext, request.TargetCourseUser.User.Role,
+			"Cannot drop a user with an equal or higher role.").Add("target-course-user", request.TargetCourseUser.User.Email)
```
eriq-augustine, line 31: Form the error in the `checkDropPermissions()` function so you can make more specific errors.
Note that the `NewBad*Permissions()` functions hide the message,
so you can be specific without worries.

eriq-augustine (changes requested):
Just one thing.

internal/api/courses/users/drop.go
Lucas-Ellenberger: I formed the error in `checkDropPermissions()` but I can't get more specific with the errors. I can't error just because they don't have proper server roles. I have to wait until after the course role check, and then it's a blanket bad perm error because they don't meet either qualification.

eriq-augustine (approved):
lgtm

PR #101, opened 2024-09-19
Added common functions for cmds to use to send and print requests.

OliverLok:


cmd/peek-submission/main.go
```diff
@@ -61,43 +44,10 @@ func main() {
...
-	}
-
-	if args.Short {
-		var responseContent submissions.FetchUserPeekResponse
-		util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
-		fmt.Println(util.MustToJSONIndent(responseContent))
-	} else {
-		fmt.Println(util.MustToJSONIndent(response))
-	}
+	cmd.PrintCMDResponse(response, submissions.FetchUserPeekResponse{}, args.Short)
```
eriq-augustine, line 52: What about wrapping together `cmd.SendCMDRequest()` and `cmd.PrintCMDResponse()` into one call?

eriq-augustine (changes requested):
Generally looking pretty good.

For this PR, add in another CMD that uses this system.
When generalizing things, it's usually easier to have multiple things that you are trying to generalize to use the same code
(it can be easier to see what needs to (and what can't) be shared).

Also, go ahead and add `cmd` to the package hierarchy:
https://github.com/edulinq/autograder-server/blob/main/docs/packages.md

cmd/user-list/main.go
```diff
@@ -17,8 +19,8 @@ var args struct {
 	config.ConfigArgs
 
 	Emails []string `help:"Optional list of users to limit the results to (leave empty for all users)." arg:"" optional:""`
```
eriq-augustine, line 21: I don't like this option.
It looks like you are relying on old (pre user split) code for an example, but that will probably lead you down a bad road.

Instead, you should be looking towards the Python interface's behavior as an example:
https://github.com/edulinq/autograder-py/blob/main/autograder/cli/users/list.py
(Since this was re-written for the new user model.)

But you can keep it even simpler, since only admins will be using this and we trust them to have basic knowledge.
So I would leave out niceties like the `--normalize` flag (just normalize by default).

cmd/user-list/main.go
```diff
@@ -17,8 +19,8 @@ var args struct {
...
-	Course string   `help:"Only include user enrolled in this course."`
 	Table  bool     `help:"Output data to stdout as a TSV." default:"false"`
+	Short  bool     `help:"Use short form output."`
```
eriq-augustine, line 23: Now that we can see multiple CMDs, I think that `--short` doesn't work very well.
You should probably swap it to `--verbose`.
I'm thinking that we tell the user when they use `--verbose`,
the output will not be in any specific format (like JSON).
Then, you can always just output the full request and response (like we do in the Python interface).
And after you output the response, then you just do the output however you would do before.
This means that in some places there is kind of "double" output,
but that's okay since the user asked for verbose mode.

cmd/user-list/main.go
```diff
@@ -31,64 +33,52 @@ func main() {
...
-func listCourseUsers(emails []string, courseID string, table bool) {
-	course := db.MustGetCourse(courseID)
+	return nil
+}
 
-	users, err := db.GetCourseUsers(course)
-	if err != nil {
-		log.Fatal("Failed to get course users.", err)
+func filterUsersByEmail(userList users.ListResponse, emails []string) users.ListResponse {
+	emailSet := make(map[string]struct{}, len(emails))
```
eriq-augustine, line 71: This function will probably go away based on earlier comments,
but I just wanted to note that for sets you should probably use either `any` or `bool` for the value types in your map (and `nil` or `true` respectively as the placeholder values).

cmd/user-list/main.go
eriq-augustine, line 1: I'm thinking that we should change the names of these CMDs to match the endpoint they are using.
Peek will be annoying, but `users/list` -> `users-list` makes sense.
(Note that nesting directories will not work since the binary's name is chosen from the parent directory.)

Try out peek and see how you like it (or if you can come up with another scheme for organizing the CMD names).

eriq-augustine (changes requested):
A good start.

I held off on some comments,
because (as per your suggestion) this would be a good PR for Lucas to review.

internal/cmd/request.go
```diff
@@ -0,0 +1,85 @@
...
+	"net"
+	"reflect"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/api/server"
+	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func SendAndPrintCMDRequest(endpoint string, request any, responseType any, shortForm bool) error {
```
Lucas-Ellenberger, line 14: Per Eriq's comment, you will want to swap the name of shortForm to verbose.

internal/cmd/request.go
```diff
@@ -0,0 +1,85 @@
...
+func SendCMDRequest(endpoint string, request any) (core.APIResponse, error) {
+	socketPath, err := common.GetUnixSocketPath()
+	if err != nil {
+		return core.APIResponse{}, fmt.Errorf("Failed to get the unix socket path: '%w'.", err)
+	}
+
+	connection, err := net.Dial("unix", socketPath)
+	if err != nil {
+		return core.APIResponse{}, fmt.Errorf("Failed to dial the unix socket: '%w'", err)
+	}
```
Lucas-Ellenberger, line 35: Extra newline after the if block?

internal/model/user_server.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
-var SERVER_USER_ROW_COLUMNS []string = []string{"email", "name", "server-role", "salt", "password", "tokens", "course-info"}
+var SERVER_USER_ROW_COLUMNS []string = []string{"email", "name", "server-role", "type"}
```
Lucas-Ellenberger, line 15: I might have misunderstood your question in discord. But we don't care to print out the type of the user info in column form (every user in a users-list will be print "server" in this column, which isn't helpful).

Instead, we want to see the email, name, server-role, and courses.

Once you print out courses from the user info struct in TSV format, you will why I suggested the `--normalize` option.

Also, I don't like that model defines the expected columns for a server user. The past variable supports printing all of the server user information (the unsafe version of user info).

Your header definition is clearly geared to support the API safe representation of user information. But, the model package is below the API package and should not define the rows for the API structs.

I would suggest adding a new variable to the api/core/user.go file. You might want to leave the old one alone and add a note that this is an unsafe representation of user info (that includes passwords and tokens) and should be handled with care.

Lucas-Ellenberger (changes requested):
The new send and print CMD suite looks incredible!

I left some notes on better ways to print the server user info.

cmd/user-list/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+		var response core.APIResponse
+		util.MustJSONFromString(stdout, &response)
+		var responseContent users.ListResponse
+		util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
+
+		expectedUsers, err := strconv.Atoi(testCase.expectedNumUsers)
+		if err != nil {
+			test.Errorf("Case %d: Failed to convert string to int.", i)
+			continue
+		}
```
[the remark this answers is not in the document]
OliverLok: The RunCMDTest function expects args as a list of strings, so converting to an int after seemed like the cleanest way to do it.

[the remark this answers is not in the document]
Lucas-Ellenberger: I'm confused how the RunCMDTest works, you pass the `test` which is a `*testing.T,` `args` which is an empty slice of strings, and a function `main`. To me it doesn't look like `testCase.expectedNumUsers` is passed to RunCMDTest, so why is it constrained to the string datatype?

internal/cmd/users.go
```diff
@@ -0,0 +1,77 @@
...
+	"strings"
+
+	"github.com/edulinq/autograder/internal/api/core"
+)
+
+const INDENT = "    "
+
+var rows [][]string
+
+func ListUsers(users []*core.ServerUserInfo, courseUsers bool, table bool) {
```
Lucas-Ellenberger, line 14: `courseUsers` is unused.

I know you want it included because it will eventually be needed, but it makes more sense for this PR to leave it out. 

We can easily expand the function and update the one function call when we have a need for listing course users (soon hopefully).

Lucas-Ellenberger (changes requested):
Real solid work!

I think splitting out the common printing functions looks a lot cleaner and will be easier to maintain.

Just a testing note and a PR style note.

internal/cmd/request.go
```diff
@@ -68,39 +63,15 @@ func SendCMDRequest(endpoint string, request any) (core.APIResponse, error) {
...
-		}
-
-		util.Exit(2)
-		return
-	}
-
-	if testingMode {
-		fmt.Println(util.MustToJSONIndent(response))
-		return
+func PrintCMDResponse(request any, response core.APIResponse, responseType any, verbose bool) {
```
Lucas-Ellenberger, line 66: I would include a function description that includes something about printing the response content as JSON.

You should also describe the behavior of verbose. Just a quick note saying we will dump the entire request and response.

cmd/peek-submission/main_test.go
```diff
@@ -60,6 +60,10 @@ func TestPeekBase(test *testing.T) {
 			continue
 		}
 
+		if testCase.expectedExitCode != 0 {
```
Lucas-Ellenberger, line 63: You include test cases with non-zero exit codes that have locators (which makes sense). But here, we skip over the locator check for non-zero exit codes.

Since we want to include the check, we need to restructure these checks. Here is the API [version](https://github.com/edulinq/autograder-server/blob/e2448c4b3a9436ac441cf4f578a53de7b7054710/internal/api/courses/assignments/submissions/fetch_user_peek_test.go#L88).

Ignore this comment and remove the expected locator from the test struct IF a non-zero exit code prevents you from checking anything else.

Lucas-Ellenberger (changes requested):
Super close!

I like the unified `PrintCMDResponse()` with verbose as an option.

Just some testing changes and a request for a function comment.

Lucas-Ellenberger (approved):
lgtm!

cmd/peek-submission/main.go
```diff
@@ -21,7 +16,7 @@ var args struct {
 	CourseID         string `help:"ID of the course." arg:""`
 	AssignmentID     string `help:"ID of the assignment." arg:""`
 	TargetSubmission string `help:"ID of the submission. Defaults to the latest submission." arg:"" optional:""`
-	Short            bool   `help:"Use short form output."`
+	Verbose          bool   `help:"Use verbose output to show full request/response without specific formatting." default:"false"`
```
eriq-augustine, line 19: I think we need to be a little creative here.

Note that this option is really owned by internal/cmd,
since it is the one that actually defines the behavior.
I think we can do something similar to config.ConfigArgs here and make a struct that contains all the base CMD arguments (as of now it is only verbose).
Then this args struct can just embed the common CMD args struct and pass those args along to the cmd functions (see how we pass args.ConfigArgs).

cmd/peek-submission/main_test.go
```diff
@@ -22,18 +22,17 @@ func TestPeekBase(test *testing.T) {
...
 
-		{"course-student@test.edulinq.org", "ZZZ", "hw0", "", "", 2, "-018"},
-		{"course-student@test.edulinq.org", "course101", "ZZZ", "", "", 2, "-022"},
-		{"course-student@test.edulinq.org", "ZZZ", "ZZZ", "", "", 2, "-018"},
+		{"course-student@test.edulinq.org", "ZZZ", "hw0", "", "", 2},
+		{"course-student@test.edulinq.org", "course101", "ZZZ", "", "", 2},
+		{"course-student@test.edulinq.org", "ZZZ", "ZZZ", "", "", 2},
 	}
 
 	for i, testCase := range testCases {
```
eriq-augustine, line 38: Make the exit code check the first check after the err check.

cmd/peek-submission/main_test.go
```diff
@@ -60,20 +59,15 @@ func TestPeekBase(test *testing.T) {
...
 
-		var response core.APIResponse
-		util.MustJSONFromString(stdout, &response)
-
-		if response.Locator != testCase.expectedLocator {
-			test.Errorf("Case %d: Unexpected locator code. Expected: '%s', Actual: '%s'.", i, testCase.expectedLocator, response.Locator)
-			continue
-		}
-
 		if testCase.expectedExitCode != 0 {
```
eriq-augustine, line 62: You are not checking any output (aside from the exit status) when an error occurs.
You should be checking stdout (which should contain an error message/string).
The output may not be JSON, but you can still check it.

cmd/peek-submission/main_test.go
```diff
@@ -60,20 +59,15 @@ func TestPeekBase(test *testing.T) {
...
 			continue
 		}
 
+		var response core.APIResponse
+		util.MustJSONFromString(stdout, &response)
```
eriq-augustine, line 67: What do you do with this?
Does it have a purpose?

cmd/user-list/main.go
```diff
@@ -2,93 +2,61 @@ package main
...
-		}
-	} else {
-		fmt.Println(util.MustToJSONIndent(users))
+		log.Fatal("Failed to list server users.", err)
 	}
 }
 
-func listCourseUsers(emails []string, courseID string, table bool) {
-	course := db.MustGetCourse(courseID)
+func listServerUsers(table bool) error {
```
eriq-augustine, line 38: Your semantics here are off.
They do not match the semantics in cmd.PrintCMDResponse().

Instead of trying to recreate the semantics on every CMD that needs some custom printing,
I recommend just passing a function that handles printing the response content.
You have probably seen the "full" pattern that I have used before where I overload functions by calling the one with all the argument "full".
You can use this pattern and by default use a function that just outputs JSON (like you do in cmd.PrintCMDResponse() right now.

cmd/user-list/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+		actualNumUsers := len(responseContent.Users)
+		if testCase.expectedNumUsers != actualNumUsers {
+			test.Errorf("Case %d: Unexpected number of server users. Expected: '%d', Actual: '%d'.", i, testCase.expectedNumUsers, actualNumUsers)
+			continue
+		}
+
+		actualEmailList := make([]string, 0, len(responseContent.Users))
+		for _, user := range responseContent.Users {
+			actualEmailList = append(actualEmailList, user.Email)
+		}
```
eriq-augustine, line 65: This is not sufficient.

For CLI tests, one of the major things you need to test is that the formatting is correct.
THis is important, because CLI programs are used in scripts, and slight changes in formatting there can break scripts.

I recommend you first checking the JSON output by a direct map comparison (probably via reflect.DeepEqual()),
then checking table output via string comparison.

cmd/user-list/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+		sort.Strings(expectedEmailList)
+		sort.Strings(actualEmailList)
```
eriq-augustine, line 68: When you eventually compare table output (maps don't need sorting for equality checks),
you need to make sure not to sort the actual output.
We expect the output of users-list to be sorted always.

cmd/user-list/main_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+		if len(stderr) > 0 {
+			test.Errorf("Case %d: CMD has content in stderr: '%s'.", i, stderr)
+			continue
+		}
+
+		if testCase.expectedExitCode != exitCode {
+			test.Errorf("Case %d: Unexpected exit code. Expected: '%d', Actual: '%d'.", i, testCase.expectedExitCode, exitCode)
+			continue
+		}
```
eriq-augustine, line 51: Of course there needs to be some changes here to sync with peek's test,
but I am seeing some opportunity for common code.

This is not hard required (yet), as I am just brainstorming.

What about some functionality in internal/cmd that defines a standard test case struct,
and a function that takes the test case (and some other stuff),
does all these things things, and returns the response content (probably an error as well).

internal/cmd/users.go
eriq-augustine, line 1: You missed the main point of my initial comment:
https://github.com/edulinq/autograder-server/pull/101#discussion_r1770541907

Note that a very important point of that comment was "keeping it even simpler".
I doubt most (if any) of this file needs to survive.
users-list is either JSON or a table, so listServerUsers() is useless.

I also made a mistake in that comment and told you to normalize,
but I actually wanted you to NOT normalize
(so there should be one row per user and just output courses as JSON in the row).
This means the bulk of listServerUsersTable() will not need to exist.

internal/api/core/user.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/edulinq/autograder/internal/model"
 )
 
+var SERVER_USER_INFO_ROW_COLUMNS []string = []string{"email", "name", "server-role", "course-name", "course-role"}
```
eriq-augustine, line 14: I am very skeptical of moving this to a higher dependency location (where it used to be in model).
However, with the changes below, this may not even need to exist
(or should just be a constant in cmd/).

eriq-augustine (changes requested):
The core functionality is here,
but the testing and output needs some work.

cmd/user-list/main_test.go
OliverLok: db.MustGetServerUsers gives back model.ServerUser while the user-list cmd gives back core.ServerUserInfo.
I suppose I could directly call the api for the expected output, but since that's what the cmd is doing that feels a bit pointless. 
Or I could manually build up the map of serverusers after the cmd calls the api by calling GetServerUser on each email. 

What do you think?

cmd/user-list/main_test.go
OliverLok: Good idea, I can try it out to see how it feels. Should this be in a separate PR?

eriq-augustine: No, it should be in this PR.
You are already making common tests, there's no need to do them twice.

cmd/peek-submission/main.go
```diff
@@ -61,43 +45,8 @@ func main() {
...
 
-	requestMap := map[string]any{
-		server.ENDPOINT_KEY: core.NewEndpoint(`courses/assignments/submissions/fetch/user/peek`),
-		server.REQUEST_KEY:  request,
-	}
-
-	jsonRequest := util.MustToJSONIndent(requestMap)
-	jsonBytes := []byte(jsonRequest)
-	err = util.WriteToNetworkConnection(connection, jsonBytes)
+	err = cmd.SendAndPrintCMDRequest(`courses/assignments/submissions/fetch/user/peek`, request, submissions.FetchUserPeekResponse{}, args.Verbose, nil)
```
eriq-augustine, line 48: Instead of sending verbose directly,
just send `args.CommonCMDArgs`.

internal/cmd/request.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/edulinq/autograder/internal/api/server"
+	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type CommonCMDArgs struct {
+	Verbose bool `help:"Use verbose output to show full request/response without specific formatting." default:"false"`
+}
+
+func SendAndPrintCMDRequest(endpoint string, request any, responseType any, verbose bool, customPrint func()) error {
```
eriq-augustine, line 18: Use the "full" pattern here as well (defaulting the print function).

internal/cmd/request.go
```diff
@@ -0,0 +1,100 @@
...
+	return response, nil
+}
+
+func PrintCMDResponse(request any, response core.APIResponse, responseType any) {
+	PrintCMDResponseFull(request, response, responseType, false, nil)
+}
+
+// Print the CMD response in it's expected JSON format.
+// If verbose is true, it also displays the complete request and response.
+func PrintCMDResponseFull(request any, response core.APIResponse, responseType any, verbose bool, customPrint func()) {
```
eriq-augustine, line 80: In pretty much all these functions, pass the arguments instead of verbose directly.

internal/cmd/request.go
```diff
@@ -0,0 +1,100 @@
...
+	"net"
+	"reflect"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/api/server"
+	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type CommonCMDArgs struct {
```
eriq-augustine, line 14: Probably change this name since we are using these as options as well as arguments.
So maybe "Options" or "CommonOptions".

internal/cmd/request.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return response, nil
+}
+
+func PrintCMDResponse(request any, response core.APIResponse, responseType any) {
+	PrintCMDResponseFull(request, response, responseType, false, nil)
+}
+
+// Print the CMD response in it's expected JSON format.
```
eriq-augustine, line 78: Outdated comment?

internal/cmd/request.go
```diff
@@ -0,0 +1,100 @@
...
+		fmt.Println(response.Message)
+
+		util.Exit(2)
+		return
+	}
+
+	if verbose {
+		fmt.Printf("\nAutograder Request:\n---\n%s\n---\n", util.MustToJSONIndent(request))
+		fmt.Printf("\nAutograder Response:\n---\n%s\n---\n", util.MustToJSONIndent(response))
+	}
```
eriq-augustine, line 91: This should probably be first, so we still get verbose output when there is an error.

internal/cmd/request.go
eriq-augustine, line 80: I don't like how the function does not take any arguments.
It should take in the object that it is supposed to print
(instead of using a closure to capture the object to print).

internal/cmd/request.go
```diff
@@ -0,0 +1,100 @@
...
+	PrintCMDResponseFull(request, response, responseType, false, nil)
+}
+
+// Print the CMD response in it's expected JSON format.
+// If verbose is true, it also displays the complete request and response.
+func PrintCMDResponseFull(request any, response core.APIResponse, responseType any, verbose bool, customPrint func()) {
+	if !response.Success {
+		fmt.Println(response.Message)
+
+		util.Exit(2)
```
eriq-augustine, line 84: It's a bit strange to exit in a print method.
It could easily trap an unsuspecting developer.
SendAndPrintCMDRequest() is a more appropriate place to exit.

internal/cmd/request.go
eriq-augustine, line 18: Maybe we should change up this name and semantics a bit to handle more and be more serious.

One possibility is to call it something like "MustHandleCMDRequestAndExit",
then you can do logs here and always util.Exit.

internal/cmd/request.go
```diff
@@ -0,0 +1,100 @@
...
+	response, err := SendCMDRequest(endpoint, request)
+	if err != nil {
+		return fmt.Errorf("Failed to send the CMD request: '%w'.", err)
+	}
+
+	if customPrint != nil {
+		PrintCMDResponseFull(request, response, responseType, verbose, customPrint)
+	} else {
+		PrintCMDResponseFull(request, response, responseType, verbose, nil)
+	}
```
eriq-augustine, line 28: These cases are effectively the same.

internal/cmd/test.go
```diff
@@ -138,3 +139,20 @@ func runCMD(mainFunc func(), args []string) (err error) {
 
 	return err
 }
+
+func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, expectedExitCode int, index int) (output string, errs error) {
```
eriq-augustine, line 143: I don't like the index here (since not all tests uses the test cases pattern).
Instead, pass a prefix (string) that the caller can build.
Tests with test cases can build it (e.g. `fmt.Sprintf("Case %d: ", i)`),
and other tests can just pass an empty string.

internal/cmd/test.go
```diff
@@ -138,3 +139,20 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args)
+	if err != nil {
+		errs = errors.Join(errs, fmt.Errorf("Case %d: CMD run returned an error: '%v'.", index, err))
```
eriq-augustine, line 146: I don't think this should be returning errors.
Just `test.Errorf()` directly.

We also don't want to be cascading these errors.

internal/cmd/test.go
eriq-augustine, line 143: I'm thinking that we can take this a bit further and also pass expected stdout and stderr (which will usually just be an empty string).

internal/cmd/test.go
```diff
@@ -138,3 +139,20 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+
+	if len(stderr) > 0 {
+		errs = errors.Join(errs, fmt.Errorf("Case %d: CMD has content in stderr: '%s'.", index, stderr))
+	}
+
+	if expectedExitCode != exitCode {
+		errs = errors.Join(errs, fmt.Errorf("Case %d: Unexpected exit code. Expected: '%d', Actual: '%d'.", index, expectedExitCode, exitCode))
+	}
+
+	return stdout, errs
```
eriq-augustine, line 157: I would just return `stdout, stderr, exitCode`.
Most people will ignore all returns from this function,
but we can include all just in case.

internal/cmd/users.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func ListServerUsersTable(users []*core.ServerUserInfo) {
```
eriq-augustine, line 11: I would consider returning a string instead of printing.
(To build the string you should probably use a strings.Builder).

eriq-augustine (changes requested):
The core code here is making sense.

cmd/peek-submission/expected_output.go
eriq-augustine, line 1: We don't want to include this in our compiled programs.
(It's not a log of space, but it's a bad program.)

Instead, I would put this data in a `testdata` directory
(which golang ignores when not running tests).
You can also put each on in a different file and then just read when you need it.

internal/cmd/request.go
```diff
@@ -0,0 +1,107 @@
...
+
+type CommonOptions struct {
+	Verbose bool `help:"Use verbose output to show full request/response without specific formatting." default:"false"`
+}
+
+func MustHandleCMDRequestAndExit(endpoint string, request any, responseType any) error {
+	return MustHandleCMDRequestAndExitFull(endpoint, request, responseType, CommonOptions{}, nil)
+}
+
+func MustHandleCMDRequestAndExitFull(endpoint string, request any, responseType any, options CommonOptions, customPrintFunc any) error {
```
eriq-augustine, line 22: "Must" functions should not return an error.
They either work or panic/fatal/exit.

I see three options:
 1. Make these log.Fatal() on error.
 2. Make these non-Must functions that return an error.
 3. Do 2, but also make Must versions that call the non-must versions (like we do a lot in the db).

Since you also need to exit with a softer error (like where you exit with 2), that makes this a bit more complex.
Either you need to do 1, or add another return value ("success"?) to this function and do 3 (where the must function then understand the third return value).

cmd/peek-submission/main_test.go
```diff
@@ -1,94 +1,70 @@
...
 
+var testCases = []struct {
+	targetEmail        string
+	courseID           string
+	assignmentID       string
+	targetSubmission   string
+	expectedSubmission string
+	expectedExitCode   int
+	expectedStdout     string
+	expectedStderr     string
```
eriq-augustine, line 18: If you always expect it to be empty, then just don't have it here.

cmd/peek-submission/main_test.go
```diff
@@ -1,94 +1,70 @@
...
+		args := []string{
+			testCase.targetEmail,
+			testCase.courseID,
+			testCase.assignmentID,
+			testCase.targetSubmission,
+		}
 
-		{"course-admin@test.edulinq.org", "course101", "hw0", "", "", 0, ""},
-		{"course-student@test.edulinq.org", "course101", "hw0", "ZZZ", "", 0, ""},
+		commonCases := cmd.CommonCMDTestCases{
```
eriq-augustine, line 45: Can you just embed this in your test case struct?

cmd/user-list/main.go
```diff
@@ -1,94 +1,36 @@
...
-		fmt.Println(strings.Join(model.COURSE_USER_ROW_COLUMNS, "\t"))
-		for _, user := range users {
-			fmt.Println(strings.Join(user.MustToRow(), "\t"))
-		}
+	if args.Table {
+		cmd.MustHandleCMDRequestAndExitFull(`users/list`, request, users.ListResponse{}, args.CommonOptions, cmd.ListServerUsersTable)
 	} else {
-		fmt.Println(util.MustToJSONIndent(users))
+		cmd.MustHandleCMDRequestAndExitFull(`users/list`, request, users.ListResponse{}, args.CommonOptions, nil)
 	}
```
eriq-augustine, line 35: You generally don't want conditionals like this.
It's not the worst thing, but you generally want things to be as simple as you can.
Here, you are calling a decently complex function in each branch of the conditional,
and the code is mostly repeated (all except for the last param).
But, you could instead just have a simple conditional that changes the last parameter.

Like:
```
var printFunc any = nil;
if (args.Table) {
    printFunc = cmd.ListServerUsersTable;
}
```

Note that the conditional is simpler, has only one branch, and there is no repeated code.

cmd/user-list/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+		if testCase.table {
+			args = append(args, "--table")
+		}
+
+		_, _, _, err := cmd.RunCMDTest(test, main, args)
+		if err != nil {
+			test.Errorf("Case %d: CMD run returned an error: '%v'.", i, err)
+		}
+	}
+}
```
eriq-augustine, line 56: You don't need a function always testing verbose.
Since the verbose functionality is common code, you only need to test it in one place.

internal/cmd/request.go
```diff
@@ -0,0 +1,107 @@
...
+	"reflect"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/api/server"
+	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type CommonOptions struct {
+	Verbose bool `help:"Use verbose output to show full request/response without specific formatting." default:"false"`
```
eriq-augustine, line 15: "without specific formatting" doesn't really make sense.
You are outputting that data specifically with a specific format (JSON).
But what you want to say is that this may cause the output to not be a valid format/filetype.

internal/cmd/request.go
eriq-augustine, line 22: This print function should probably be a declared type.
E.g.:
https://github.com/edulinq/autograder-server/blob/main/internal/api/core/routing.go#L21

internal/cmd/request.go
```diff
@@ -0,0 +1,107 @@
...
+		fmt.Printf("\nAutograder Response:\n---\n%s\n---\n", util.MustToJSONIndent(response))
+	}
+
+	if customPrintFunc != nil {
+		customPrintFuncValue := reflect.ValueOf(customPrintFunc)
+
+		responseValue := reflect.ValueOf(response)
+		responseSlice := []reflect.Value{responseValue}
+
+		response := customPrintFuncValue.Call(responseSlice)[0]
```
eriq-augustine, line 100: You won't need this id you declare the function's type.

internal/cmd/test.go
```diff
@@ -135,3 +137,56 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+}
+
+func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonCases CommonCMDTestCases, prefix string) (string, string, int) {
+	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args)
+	if err != nil {
+		test.Errorf("%sCMD run returned an error: '%v'.", prefix, err)
+	}
+
+	if commonCases.ExpectedStderr != "" {
+		if commonCases.ExpectedStderr != stderr {
```
eriq-augustine, line 154: Why does stdout get all the special checks and stderr not?

internal/cmd/test.go
```diff
@@ -135,3 +137,56 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+	ExpectedStderr   string
+}
+
+func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonCases CommonCMDTestCases, prefix string) (string, string, int) {
+	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args)
+	if err != nil {
+		test.Errorf("%sCMD run returned an error: '%v'.", prefix, err)
+	}
+
+	if commonCases.ExpectedStderr != "" {
```
eriq-augustine, line 153: Why is this check necessary?
If we always just do the normal equality check, then we also don't need to length check below.
stderr will either be exactly as we want (which may be empty) or we fail.

internal/cmd/test.go
```diff
@@ -135,3 +137,56 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+	}
+
+	if !compareOutputs(commonCases.ExpectedStdout, stdout) {
+		test.Errorf("%sUnexpected output. Expected:\n'%s',\nActual:\n'%s'.", prefix, commonCases.ExpectedStdout, stdout)
+	}
+
+	return stdout, stderr, exitCode
+}
+
+func compareOutputs(expected string, actual string) bool {
```
eriq-augustine, line 174: Why all the work here?
Why not just a straight equality check?
Why break these into lines?

Are you trying to correct for line endings (newlines vs carriage returns)?
If so, then you should just do a global replace of carriage returns `\r` with empty strings.

internal/cmd/users.go
```diff
@@ -0,0 +1,38 @@
...
+		serverUsersTable.WriteString(user.Email)
+		serverUsersTable.WriteString("\t")
+		serverUsersTable.WriteString(user.Name)
+		serverUsersTable.WriteString("\t")
+		serverUsersTable.WriteString(user.Role.String())
+		serverUsersTable.WriteString("\t")
+		serverUsersTable.WriteString(coursesJSON)
+
+		// Only add a newline if this is not the last user
+		if i < len(responseContent.Users)-1 {
```
eriq-augustine, line 32: Missing space around binary operator (minus).

eriq-augustine, line 32: Instead of doing there here (and computing a len),
just do it before the row and check for `i > 0`.

eriq-augustine (changes requested):
Looking pretty good.

internal/cmd/request.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func MustHandleCMDRequestAndExitFull(endpoint string, request any, responseType any, options CommonOptions, customPrintFunc any) error {
+	response, err := SendCMDRequest(endpoint, request)
+	if err != nil {
+		return fmt.Errorf("Failed to send the CMD request: '%w'.", err)
+	}
+
+	if !response.Success {
+		fmt.Println(response.Message)
```
eriq-augustine, line 29: No, you can't do this.
This message needs to be accessible to someone.
When printed it's not on stderr (like the logs) and not accessible via an error,
so it's just gone to anyone who is not capturing stdout (which we already talked about how bad that is to do for most other code in most cases).

internal/cmd/users.go
OliverLok: Formatting removes the space. I could declare the length on the line before, but that feels a bit unnecessary.

cmd/peek-submission/expected_output_test.go
```diff
@@ -1,4 +1,4 @@
-package main
+package peek
 
 const latestSubmission = `{
```
eriq-augustine, line 3: For const, use the CORRECT_NAMING_CONVENTION.

cmd/peek-submission/expected_output_test.go
```diff
@@ -151,3 +151,7 @@ const incorrectSubmission = `{
     "submission-result": null
 }
 `
+
+const incorrectCourse = `{"message":"Could not find course: 'ZZZ'."}`
+
+const incorrectAssignment = `{"message":"Could not find assignment: 'zzz'."}`
```
eriq-augustine, line 157: Don't put such small test case responses in a file.
We put the big ones here because they would make the actual test hard to read,
but that adds overhead.
For small strings, it's not worth the overhead.
Depending on the size of your output, you may want to break your test case over multiple lines (which is super common).

cmd/peek-submission/main_test.go
```diff
@@ -42,29 +39,8 @@ func TestPeekBase(test *testing.T) {
...
-			testCase.assignmentID,
-			testCase.targetSubmission,
-			"--verbose",
-		}
-
-		_, _, _, err := cmd.RunCMDTest(test, main, args)
-		if err != nil {
-			test.Errorf("Case %d: CMD run returned an error: '%v'.", i, err)
-		}
+		cmd.RunVerboseCMDTests(test, main, args, fmt.Sprintf("Case %d: ", i))
```
eriq-augustine, line 44: No, make this another test.
(I have talked about this several times, if you don't understand talk to me.)

cmd/peek-submission/main.go
```diff
@@ -1,4 +1,4 @@
-package main
+package peek
```
eriq-augustine, line 1: I don't think this will even work.
Have you tried to actually run the CMD like this?

cmd/user-list/main_test.go
```diff
@@ -30,27 +28,8 @@ func TestServerUserListBase(test *testing.T) {
...
-		if testCase.table {
-			args = append(args, "--table")
-		}
+		cmd.RunCommonCMDTests(test, main, args, testCase.CommonCMDTestCases, fmt.Sprintf("Case %d: ", i))
 
-		_, _, _, err := cmd.RunCMDTest(test, main, args)
-		if err != nil {
-			test.Errorf("Case %d: CMD run returned an error: '%v'.", i, err)
-		}
+		cmd.RunVerboseCMDTests(test, main, args, fmt.Sprintf("Case %d: ", i))
```
eriq-augustine, line 33: !!!
I explicitly talked about this in the last round.

internal/cmd/request.go
```diff
@@ -8,34 +8,31 @@ import (
 	"github.com/edulinq/autograder/internal/api/core"
 	"github.com/edulinq/autograder/internal/api/server"
 	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/log"
 	"github.com/edulinq/autograder/internal/util"
 )
 
 type CommonOptions struct {
-	Verbose bool `help:"Use verbose output to show full request/response without specific formatting." default:"false"`
+	Verbose bool `help:"Use verbose output to show the full request/response. Output may not adhere to a valid format/filetype." default:"false"`
```
eriq-augustine, line 16: The wording here is strange.
When writing help for an option, we normally don't reference the option itself:
```
grep -R '`help:' internal cmd
```

I still don't think that the last part here would be clear to users.

internal/cmd/request.go
```diff
@@ -8,34 +8,31 @@ import (
...
 }
 
-func MustHandleCMDRequestAndExit(endpoint string, request any, responseType any) error {
-	return MustHandleCMDRequestAndExitFull(endpoint, request, responseType, CommonOptions{}, nil)
+type CMDPrintHandler func(response core.APIResponse) string
```
eriq-augustine, line 19: If this function is not actually printing (which I think it is a good pattern that it does not print),
then you should change the name (since it is not handling printing).
I would also drop the prefix, since it is already in the cmd package.

internal/cmd/request.go
```diff
@@ -85,20 +82,21 @@ func PrintCMDResponse(request any, response core.APIResponse, responseType any)
 
 // Print the CMD response in it's expected or custom format.
 // If verbose is true, it also displays the complete request and response.
```
eriq-augustine, line 84: You need to talk about the semantics of customPrintFunc (just one sentence).

internal/cmd/request.go
```diff
@@ -85,20 +82,21 @@ func PrintCMDResponse(request any, response core.APIResponse, responseType any)
...
 	}
 
-	if customPrintFunc != nil {
-		customPrintFuncValue := reflect.ValueOf(customPrintFunc)
+	if !response.Success {
+		log.Error("API response was unsuccessful.", log.NewAttr("message", response.Message))
 
-		responseValue := reflect.ValueOf(response)
-		responseSlice := []reflect.Value{responseValue}
+		util.Exit(2)
```
eriq-augustine, line 94: !!!
We talked about this in the last round.
Print function should NOT exit.

internal/cmd/test.go
```diff
@@ -135,3 +137,43 @@ func runCMD(mainFunc func(), args []string) (err error) {
 
 	return err
 }
+
+type CommonCMDTestCases struct {
```
eriq-augustine, line 141: Types should be declared at the top of the file (below consts, above funcs).

internal/cmd/test.go
```diff
@@ -135,3 +137,43 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+
+func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonCases CommonCMDTestCases, prefix string) (string, string, int) {
+	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args)
+	if err != nil {
+		test.Errorf("%sCMD run returned an error: '%v'.", prefix, err)
+	}
+
+	if len(stderr) > 0 && commonCases.ExpectedStderr == "" {
+		test.Errorf("%sCMD has content in stderr: '%s'.", prefix, stderr)
+	}
```
eriq-augustine, line 155: I mentioned this in my previous round.

Why do this check when we can just do an equality check on stderr?
If we expect stderr to by empty, we will pass an empty string
(which will effectively to the same check with less cases).

internal/cmd/test.go
```diff
@@ -135,3 +137,43 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+
+	if !strings.Contains(stderr, commonCases.ExpectedStderr) {
+		test.Errorf("%sUnexpected stderr. Expected substring: '%s', Actual stderr: '%s'.", prefix, commonCases.ExpectedStderr, stderr)
+	}
```
eriq-augustine, line 159: Why not a direct equality check?

I can see some more complex cases that could come up (you see these in the Python test infrastructure),
but I don't see those for peek and list.
(And this wouldn't be the way to handle it.)

internal/cmd/test.go
```diff
@@ -135,3 +137,43 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
 	return err
 }
+
+type CommonCMDTestCases struct {
+	ExpectedExitCode int
+	ExpectedStdout   string
+	ExpectedStderr   string
+}
+
+func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonCases CommonCMDTestCases, prefix string) (string, string, int) {
```
eriq-augustine, line 147: This function needs a final return, an "ok" boolean that signals if an error occurred.
This will allow tests to decide if they want to move on to the next test case.

internal/cmd/test.go
```diff
@@ -135,3 +137,43 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args)
+	if err != nil {
+		test.Errorf("%sCMD run returned an error: '%v'.", prefix, err)
```
eriq-augustine, line 150: We talked about this a previous round.

We don't want our test failures to cascade.
You need to return on every `test.Errorf()`.

internal/cmd/test.go
```diff
@@ -135,3 +137,43 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
+	}
+
+	if commonCases.ExpectedStdout != stdout {
+		test.Errorf("%sUnexpected output. Expected:\n'%s',\nActual:\n'%s'.", prefix, commonCases.ExpectedStdout, stdout)
+	}
+
+	return stdout, stderr, exitCode
+}
+
+func RunVerboseCMDTests(test *testing.T, mainFunc func(), args []string, prefix string) {
```
eriq-augustine, line 172: I don't think this is necessary.

eriq-augustine (changes requested):
This round was a bit disappointing.
There were several things that were already commented on in previous rounds.
Please be more careful in the future and ask me if you have questions.

cmd/peek-submission/expected_output_test.go
```diff
@@ -0,0 +1,153 @@
...
+            }
+        ],
+        "grading_start_time": 1697406273000,
+        "grading_end_time": 1697406273000,
+        "additional-info": null
+    }
+}
+`
+
+const SPECIFIC_SUBMISSION_SHORT = `{
```
eriq-augustine, line 49: Is this just duplicate data?
(If so, don't do that.)

Your variables should be based on output/content,
not the input that was used to query them.

You may as well just use one submission (1697406272) instead of multiple ones.
Remember, you are not testing the general peek functionality, you are testing just the CLI.

cmd/peek-submission/main_test.go
```diff
@@ -1,41 +1,88 @@
...
+		},
+		targetEmail:      "course-student@test.edulinq.org",
+		courseID:         "course101",
+		assignmentID:     "hw0",
+		targetSubmission: "ZZZ",
+	},
+	{
+		CommonCMDTestCases: cmd.CommonCMDTestCases{
+			ExpectedExitCode: 2,
+			ExpectedStderr:   `{"message":"Could not find course: 'ZZZ'."}`,
```
eriq-augustine, line 63: Just query for the string "Could not find course: 'ZZZ'.".
Including the rest makes it harder for people to see what the important part is.

internal/cmd/users.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"strings"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/api/users"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func ListServerUsersTable(response core.APIResponse) string {
```
eriq-augustine, line 11: Does anyone else us this function?
If not, then it should probably just be non-exported in users-list.

cmd/user-list/main_test.go
```diff
@@ -0,0 +1,53 @@
+package main
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/edulinq/autograder/internal/cmd"
+)
+
+var testCases = []struct {
```
eriq-augustine, line 10: Since this isn't shared, treat it like we do all our other test cases and declare it in the function it is used.

internal/cmd/request.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/log"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type CommonOptions struct {
+	Verbose bool `help:"Add the full request and response to the output. Be aware that the output will include extra details beyond the expected format." default:"false"`
+}
+
+type CustomFormat func(response core.APIResponse) string
```
eriq-augustine, line 19: The name needs work.
You want all your variable names to be accurate and precise.
Here, your name is lying (not accurate).
This variable is not representing a "format", it is representing a "formatter".

internal/cmd/request.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/api/server"
+	"github.com/edulinq/autograder/internal/common"
+	"github.com/edulinq/autograder/internal/log"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type CommonOptions struct {
+	Verbose bool `help:"Add the full request and response to the output. Be aware that the output will include extra details beyond the expected format." default:"false"`
```
eriq-augustine, line 16: s/details/text/

It is not the "details" that is the issue, it is the extra text that potentially breaks the format.
Don't let nice language make your statements incorrect.

internal/cmd/request.go
```diff
@@ -0,0 +1,104 @@
...
+	socketPath, err := common.GetUnixSocketPath()
+	if err != nil {
+		return core.APIResponse{}, fmt.Errorf("Failed to get the unix socket path: '%w'.", err)
+	}
+
+	connection, err := net.Dial("unix", socketPath)
+	if err != nil {
+		return core.APIResponse{}, fmt.Errorf("Failed to dial the unix socket: '%w'", err)
+	}
+
```
eriq-augustine, line 47: Lucas was incorrect here.
There should be no newline, since the defer is strongly related to the code above
(it is ensuring the opened connection is closed).

internal/cmd/request.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func PrintCMDResponse(request any, response core.APIResponse, responseType any) int {
+	return PrintCMDResponseFull(request, response, responseType, CommonOptions{}, nil)
+}
+
+// Print the CMD response in it's expected or custom format.
+// customPrintFunc defines a function that takes the CMD response and formats it into a custom output.
+// If verbose is true, it also displays the complete request and response.
+func PrintCMDResponseFull(request any, response core.APIResponse, responseType any, options CommonOptions, customPrintFunc CustomFormat) int {
```
eriq-augustine, line 84: I don't see a need to hand any return value for this function.
The caller passed in the response, so they can just check success on their own.

(And if we did want a return, we would just return a bool, not an int.)

internal/cmd/test.go
```diff
@@ -135,3 +170,28 @@ func runCMD(mainFunc func(), args []string) (err error) {
...
 }
+
+func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonCases CommonCMDTestCases, prefix string) (string, string, int, bool) {
+	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args)
+	if err != nil {
+		test.Errorf("%sCMD run returned an error: '%v'.", prefix, err)
+		return "", "", -1, false
+	}
+
+	if !strings.Contains(stderr, commonCases.ExpectedStderr) {
```
eriq-augustine, line 181: !!!
We explicitly talked about this!
This name is bad.

As a reviewer, the fact that you missed this is very concerning
(since we spent so much time talking about it).
Do what you need to do (take notes, record conversations, ask questions, etc) to never make a mistake like this again.

eriq-augustine (changes requested):
I feel like you need to slow down a bit and take some more care and consideration on this PR.
Aside from things that were mentioned previously, there are some design things (like ensuring the server has stopped) that I think may not have been thought through all the way.

internal/cmd/test.go
```diff
@@ -70,9 +104,6 @@ func RunCMDTest(test *testing.T, mainFunc func(), args []string) (string, string
 		log.SetBackendLevel(oldBackendLogLevel)
 	}()
 
-	// Force the log level to fatal.
-	args = append(args, "--log-level", log.LevelFatal.String())
```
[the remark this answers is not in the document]
OliverLok: I wanted to capture the stderr (log.Error) when the response isn't a success to make sure it was the correct error so setting the log level to fatal would suppress those logs.

internal/cmd/test.go
```diff
@@ -85,8 +116,12 @@ func RunCMDTest(test *testing.T, mainFunc func(), args []string) (string, string
 	oldStderr := os.Stderr
 	defer func() {
 		os.Stderr = oldStderr
+		log.SetTextWriter(oldStderr)
 	}()
 
+	stderrFile := util.MustCreateFile(stderrPath)
+	os.Stderr = stderrFile
+	log.SetTextWriter(stderrFile)
```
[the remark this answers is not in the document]
OliverLok: I'm not sure I understand.
Correct me if I'm wrong, but from what I understand, the actual writing of logs is done through the logToText function (writes to it's own textWriter variable).
The logging system is still writing to the original os.Stderr, not the file unless the textwriter is set to the file.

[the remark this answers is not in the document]
eriq-augustine: > Correct me if I'm wrong, but from what I understand, the actual writing of logs is done through the logToText function (writes to it's own textWriter variable).

Yeah, but that variable defaults to `os.Stderr`:
https://github.com/edulinq/autograder-server/blob/main/internal/log/logging.go#L37

So what you are trying to do is capture both logging output and stderr output to the same file, right?
You comments here.
Whenever you are doing anything that may be tricky, you should comment it!

internal/cmd/test.go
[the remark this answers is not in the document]
eriq-augustine: But you are not just doing that.
You are forcing all tests to either use the default logging level or set their own level.
I think you need to think more about how other tests will want to interact with this,
keeping in mind that overriding in an option.

internal/api/server/http_socket.go
```diff
@@ -14,9 +15,19 @@ import (
...
 
 func runAPIServer(routes *[]core.Route) (err error) {
+	var port = config.WEB_PORT.Get()
+
+	common.Lock(PORT_USAGE_CHECK_LOCK)
+	err = checkPortInUse(port)
+	if err != nil {
+		return err
+	}
+	common.Unlock(PORT_USAGE_CHECK_LOCK)
```
eriq-augustine, line 29: There are a couple issues here:
1) You will never unlock on an error.
2) Why not lock inside the function itself, and then you can defer an unlock?
3) Why use a different key than the rest of the file.

!!! (This part is very important.)
I think you may have misunderstood that I asked for before.
I said to error if the server is being used and someone calls runAPIServer(),
and I said to check this you would need to use a lock and manually unlock (not defer) around the check so that you allow people to finish any ongoing calls to stopAPIServer() before you get mad at them.
It is critically important that you don't just listen to what I say but REALLY understand it.
This is not a homework assignment where it is okay for you to miss a few points here and there,
this is a production system that real people rely on.
So it's vital that you can fully stand behind all your code and you can't do that if you don't understand the code and the reasons you are doing it.
The code you have here doesn't make sense on its own,
so my only real conclusion is that you misunderstood what I said and tried to just implement what you thought I said without understanding it.
Ask questions when you need them.

internal/cmd/test.go
```diff
@@ -26,17 +25,27 @@ const (
 	LOCAL_HOST      = "localhost"
 )
 
-type CommonCMDTestCases struct {
-	ExpectedExitCode int
-	ExpectedStdout   string
-	ExpectedStderr   string
+type CMDTestParameters struct {
```
eriq-augustine, line 28: I like the old name (except that it should be singular).

The log level affects the output and is therefore part of the test case.

internal/cmd/test.go
```diff
@@ -26,17 +25,27 @@ const (
...
 }
 
 // Common setup for all CMD tests that require a server.
 func CMDServerTestingMain(suite *testing.M) {
+	server.StopServer()
+
+	port, err := getUnusedPort()
+	if err != nil {
+		log.Error("Failed to get an unused port.", err)
+		os.Exit(1)
```
eriq-augustine, line 42: We don't use os.Exit() anymore, we use exit.Exit():
https://github.com/edulinq/autograder-server/blob/main/internal/exit/exit.go#L19

cmd/peek-submission/main_test.go
```diff
@@ -7,82 +7,82 @@ import (
...
-		targetSubmission: "ZZZ",
-	},
-	{
-		CommonCMDTestCases: cmd.CommonCMDTestCases{
-			ExpectedExitCode: 2,
-			ExpectedStderr:   `{"message":"Could not find course: 'ZZZ'."}`,
+		{
+			CMDTestParameters: cmd.CMDTestParameters{
+				ExpectedExitCode:        2,
+				ExpectedStderrSubstring: `"Could not find course: 'ZZZ'."`,
```
eriq-augustine, line 69: Don't include the outer double quotes,
it makes it look more confusing.

eriq-augustine (changes requested):
This only reviews new things,
this is not a full review.

internal/api/server/http_socket.go
eriq-augustine: Something else to note here, is that if the port is not available the function already returns an error on line 47.
So the port check on line 25 seems unnecessary.

internal/cmd/test.go
```diff
@@ -21,10 +24,28 @@ const (
...
+
 // Common setup for all CMD tests that require a server.
 func CMDServerTestingMain(suite *testing.M) {
-	// Run inside a func so defers will run before os.Exit().
+	server.StopServer()
+
+	port, err := getUnusedPort()
+	if err != nil {
+		log.Error("Failed to get an unused port.", err)
+		exit.Exit(1)
```
eriq-augustine, line 41: You can just log.Fatal() here (and another time below).

internal/api/server/http_socket.go
```diff
@@ -14,9 +14,16 @@ import (
...
-const API_SERVER_STOP_LOCK = "internal.api.server.API_STOP_LOCK"
+const API_SERVER_LOCK = "internal.api.server.API_SERVER_LOCK"
 
 func runAPIServer(routes *[]core.Route) (err error) {
+	common.Lock(API_SERVER_LOCK)
+	if apiServer != nil {
+		common.Unlock(API_SERVER_LOCK)
+		return fmt.Errorf("API server is already running.")
+	}
+	common.Unlock(API_SERVER_LOCK)
```
eriq-augustine, line 25: I know I asked for this to be first and all together, but I think we can make it a bit better.
I'm thinking:
 - Do the recover defer first in the function, followed by this.
 - Move the unlock until after you set apiServer (note that there are no possible errors or returns in that section.

This way there shouldn't be any holes where two people can call runServer().

cmd/user-list/main.go
```diff
@@ -1,94 +1,67 @@
...
 		}
 
-		users = newUsers
+		serverUsersTable.WriteString(user.Email)
+		serverUsersTable.WriteString("\t")
+		serverUsersTable.WriteString(user.Name)
+		serverUsersTable.WriteString("\t")
+		serverUsersTable.WriteString(user.Role.String())
+		serverUsersTable.WriteString("\t")
+		serverUsersTable.WriteString(util.MustToJSONIndent(user.Courses))
```
eriq-augustine, line 63: This breaks the output.
You wanna have one entry per row.

internal/cmd/request.go
```diff
@@ -0,0 +1,106 @@
...
+type CustomResponseFormatter func(response core.APIResponse) string
+
+func MustHandleCMDRequestAndExit(endpoint string, request any, responseType any) {
+	MustHandleCMDRequestAndExitFull(endpoint, request, responseType, CommonOptions{}, nil)
+}
+
+func MustHandleCMDRequestAndExitFull(endpoint string, request any, responseType any, options CommonOptions, customPrintFunc CustomResponseFormatter) {
+	response, err := SendCMDRequest(endpoint, request)
+	if err != nil {
+		log.Fatal("Failed to send the CMD request.", err)
```
eriq-augustine, line 29: Add the endpoint to the log.

internal/cmd/request.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	if !response.Success {
+		log.Error("API response was unsuccessful.", log.NewAttr("message", response.Message))
+
+		exit.Exit(2)
```
eriq-augustine, line 35: Go with 1 instead of 2.

internal/cmd/test.go
```diff
@@ -46,10 +72,20 @@ func CMDServerTestingMain(suite *testing.M) {
 		return suite.Run()
 	}()
 
-	os.Exit(code)
+	exit.Exit(code)
 }
 
-func RunCMDTest(test *testing.T, mainFunc func(), args []string) (string, string, int, error) {
+func getUnusedPort() (int, error) {
+	listener, err := net.Listen("tcp", ":0")
```
eriq-augustine, line 79: Make sure you some looping tests (just repeatedly calling real fast) on the code that uses this.
(Note that tests looping tests are not something included in your PR, just loke count=100 or the script we made.)
I just want you to make sure that we successfully give up this port that we bound to when checking that it is open.

eriq-augustine (changes requested):
Looking real good now,
just a few things.

cmd/user-list/main.go
OliverLok: By one entry do you mean one WriteString per row?

eriq-augustine: No.

The idea with `--table` is to write out a TSV file (as the help description says).
You currently write lines like:
```
11	12	13	14	15
21	22	{
    "23a": "23b"
}	24	25
31	32	33	34	35
```

Which, of course, is not proper TSV.

internal/cmd/request.go
```diff
@@ -76,6 +74,7 @@ func SendCMDRequest(endpoint string, request any) (core.APIResponse, error) {
 
 	var response core.APIResponse
 	util.MustJSONFromBytes(responseBuffer, &response)
+
 	if err != nil {
 		return core.APIResponse{}, fmt.Errorf("Failed to unmarshal the API response: '%w'.", err)
 	}
```
eriq-augustine, line 80: This check is incorrect.
err has not been changed since it was last checked.

internal/api/server/http_socket.go
```diff
@@ -33,6 +26,12 @@ func runAPIServer(routes *[]core.Route) (err error) {
 		err = errors.Join(err, fmt.Errorf("API server panicked: '%v'.", value))
 	}()
 
+	common.Lock(API_SERVER_LOCK)
```
eriq-augustine, line 29: Nit (I would normally not hold up an approval for this, but there is another change you need to make so I may as well include it).

Add a comment here saying that we don't defer the unlock, but we unlock on all possible code paths.

internal/cmd/request.go
```diff
@@ -26,15 +26,13 @@ func MustHandleCMDRequestAndExit(endpoint string, request any, responseType any)
...
 
 	if !response.Success {
-		log.Error("API response was unsuccessful.", log.NewAttr("message", response.Message))
+		log.Fatal("API Request was unsuccessful.", log.NewAttr("message", response.Message))
 
-		exit.Exit(2)
-
-		// Return after setting the exit code to avoid overwriting it during tests.
+		// Return after log.Fatal sets the exit code to 1 to avoid overwriting the exit code during tests.
```
eriq-augustine, line 35: Nit (I would normally not hold up an approval for this, but there is another change you need to make so I may as well include it).

We we mention functions, we usually add parens after (e.g. `log.Fatal()`).

eriq-augustine (changes requested):
Just one thing (and some nits).

eriq-augustine (approved):
Nice, lgtm.

PR #102, opened 2024-09-19
Added an API endpoint for getting course users.

Lucas-Ellenberger:
New API endpoint: courses/users/get.

This API endpoint allows users to get course information about themselves. Course admin or above can use the endpoint to get course information about other course users.

This PR must be reviewed with the associated python interface implementation, which can be found [here](https://github.com/edulinq/autograder-py/pull/31).

### DIFF
--- a/internal/api/courses/users/get.go
+++ b/internal/api/courses/users/get.go
@@ -1,0 +1,30 @@
+package users
+
+import (
+	"github.com/edulinq/autograder/internal/api/core"
+)
+
+type GetRequest struct {
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleOther
+
+	TargetCourseUser core.TargetCourseUserSelfOrAdmin `json:"target-email"`
+}
+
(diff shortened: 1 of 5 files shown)


internal/api/courses/users/get.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/edulinq/autograder/internal/api/core"
+)
+
+type GetRequest struct {
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleOther
+
+	TargetCourseUser core.TargetCourseUserSelfOrAdmin `json:"target-email"`
```
eriq-augustine, line 11: Move this down to `core.TargetCourseUserSelfOrGrader` (instead of admin),
since graders will sometimes need to do things like this.

eriq-augustine (changes requested):
Looking good, just one permissions level thing.

eriq-augustine (approved):
lgtm

PR #103, opened 2024-09-21
Updated the autograder temp directory from /tmp to /tmp/autograder-temp

BatuhanSA:


internal/util/dir.go
```diff
@@ -31,6 +32,10 @@ func MkDirTemp(prefix string) (string, error) {
 	tempDirMutex.Lock()
 	defer tempDirMutex.Unlock()
 
+	if tempDir != "" {
+		MkDir(tempDir)
+	}
+
```
eriq-augustine, line 38: Is this necessary?

eriq-augustine (changes requested):
There is not too much to do here.

The name of the PR is not accurate, go ahead and update it.

internal/util/dir.go
BatuhanSA: Unfortunately, we need to use MkDIr here because when we pass tempDir to os.MkdirTemp(tempDir, prefix), tempDir must already exist.

README.md
```diff
@@ -23,8 +23,8 @@ The autograding effort is broken into three main parts:
 ## Installation
 
 This project uses Go 1.21.
-Development and deployment are intended for a Linux environment,
-but all Go code is written in a platform independent fashion.
+Development and deployment are intended for a Linux environment.
+Temporary file management relies on POSIX (i.e Linux, MacOS, WSL) conventions
```
eriq-augustine, line 27: The extra details here (temp file management) are incorrect.
This is a case where you can just be more general, and create something that is both more accurate and generally better.

eriq-augustine, line 27: s/i.e/i.e.,/

eriq-augustine, line 27: Missing period.

eriq-augustine (changes requested):
Just some documentation issues.

README.md
```diff
@@ -23,8 +23,8 @@ The autograding effort is broken into three main parts:
 ## Installation
 
 This project uses Go 1.21.
-Development and deployment are intended for a Linux environment,
-but all Go code is written in a platform independent fashion.
+Development and deployment are intended for Linux environments,
```
eriq-augustine, line 26: This is not true anymore.
Development and deployment just rely on a POSIX system.

README.md
```diff
@@ -23,8 +23,8 @@ The autograding effort is broken into three main parts:
...
+with Go code relying on POSIX (e.g., Linux, MacOS, WSL) conventions.
```
eriq-augustine, line 27: "Conventions" is not correct, "system" is the correct word here.

eriq-augustine (changes requested):
Needs some more doc work.

eriq-augustine (approved):
lgtm

PR #107, opened 2024-09-23
Added a test token that is used for a deletion test on the python interface

Lucas-Ellenberger:
This PR adds a token to `course-admin@test.edulinq.org` so the python interface can verify an API with a successful token deletion.

The added test case caused issues for some internal tests, so I implemented a test function to clear all tokens on the server, which is used by the affected test cases.

This PR allows the associated python interface PR's tests to function, so they must be reviewed together. The PR can be found [here](https://github.com/edulinq/autograder-py/pull/32).


eriq-augustine (changes requested):
I don't know if I am a fan of this solution for the test cases.
My immediate thought would be to just have the token there and deal with it normally in the test cases.
What moved you away from that solution?

eriq-augustine (approved):
Nice, lgtm.

PR #110, opened 2024-10-14
Added a new CMD to describe the API endpoints on the server.

Lucas-Ellenberger:
This PR adds a new CMD that describes all API endpoints on the server by outputting the endpoints name, their description, request type, and response type.

This PR adds a `Route` interface and two structs that implement the interface, `BaseRoute` and `APIRoute`. Base routes are used for basic routes and redirects. API routes are used for all APIs on the server. The distinction allows us to use reflection to determine if we need to describe the endpoint during the new command.

API routes also store additional information that is useful for it's description. Currently, this only contains the reflection type of the request and response structs for the endpoint.


internal/api/core/routing.go
```diff
@@ -34,48 +34,104 @@ type ValidAPIHandler any
...
+	return route.Regex
+}
+
+func (route *APIRoute) GetSuffix() string {
+	return route.Suffix
+}
+
+func (route *APIRoute) Handle(response http.ResponseWriter, request *http.Request) error {
+	return route.Handler(response, request)
+}
```
eriq-augustine, line 100: Doesn't APIRoute already get most of these methods for free since it embeds BaseRoute?

internal/api/core/routing.go
```diff
@@ -34,48 +34,104 @@ type ValidAPIHandler any
...
+
+type APIRoute struct {
+	BaseRoute
+	Request  reflect.Type
+	Response reflect.Type
 }
 
 const MAX_FORM_MEM_SIZE_BYTES = 20 * 1024 * 1024 // 20 MB
 
+func (route *BaseRoute) GetDescription() string {
```
eriq-augustine, line 62: We typically use "this" for our receiver (context variable).

internal/api/core/routing.go
```diff
@@ -34,48 +34,104 @@ type ValidAPIHandler any
...
+type Route interface {
+	GetDescription() string
+	GetMethod() string
+	GetRegex() *regexp.Regexp
+	GetSuffix() string
+	Handle(response http.ResponseWriter, request *http.Request) error
+}
+
 // Inspired by https://benhoyt.com/writings/go-routing/
-type Route struct {
```
eriq-augustine, line 38: Move the route stuff to it's own file "route.go"?
(Leave routing stuff here.)

eriq-augustine (changes requested):
Looking pretty good.
The core stuff is here,
most the work needs to be around doing it in the proper Go way.

internal/api/core/route.go
```diff
@@ -0,0 +1,117 @@
...
+	BasePath string
+	Regex    *regexp.Regexp
+	Handler  RouteHandler
+}
+
+type APIRoute struct {
+	BaseRoute
+	Description string
+	Request     reflect.Type
+	Response    reflect.Type
```
eriq-augustine, line 30: These fields should be suffixed with "Type" (since they are types and not the values themselves).

internal/api/core/route.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return &BaseRoute{
+		Method:   method,
+		BasePath: basePath,
+		Regex:    regexp.MustCompile("^" + NewEndpoint(basePath) + "$"),
+		Handler:  redirectFunc,
+	}
+}
+
+func NewAPIRoute(basePath string, apiHandler any, description string) *APIRoute {
```
eriq-augustine, line 75: As per a comment last round, descriptions should eventually be parsed from comments and not passed directly.

I also said that you should move the descriptions from the passed-in strings to comments on the handlers,
but you just copied it...
You know how I generally feel about copied code ...
(Bad.)

internal/api/core/route.go
```diff
@@ -0,0 +1,117 @@
...
+		}()
+
+		endpointPath = request.URL.Path
+
+		err = handleAPIEndpoint(response, request, apiHandler)
+
+		return err
+	}
+
+	_, requestType, responseType, err := validateAPIHandler(endpointPath, apiHandler)
```
eriq-augustine, line 101: This is invalid.
`endpointPath` has not been populated.

internal/api/core/routing.go
```diff
@@ -272,50 +227,51 @@ func allocateAPIRequest(endpoint string, apiHandler ValidAPIHandler) (any, *APIE
 }
 
 // Reflexively ensure that the api handler is of the correct type/format (e.g. looks like APIHandler).
+// Returns a ValidAPIHandler, the reflect.Type of the input struct and first output struct, and an APIError.
```
eriq-augustine, line 230: You can be more clear here by writing more simply.
They can see the types in the signature, so you can leave them out.
Remember, documentation is not about repeating code.
It is about stating it in a different way.

So you can have something more clear that is like:
"Returns the handler, the type of ..., the type of ..., and an error.".

internal/api/routes.go
```diff
@@ -9,19 +9,30 @@ import (
...
-	core.NewRoute("GET", `/static`, static.Handle),
-	core.NewRoute("GET", `/static/.*`, static.Handle),
+	core.NewBaseRoute("GET", `/static`, static.Handle),
+	core.NewBaseRoute("GET", `/static/.*`, static.Handle),
 }
 
-func GetRoutes() *[]*core.Route {
-	routes := make([]*core.Route, 0)
+type APIDescription struct {
+	Endpoints map[string]map[string]EndpointDescription
```
eriq-augustine, line 25: Why the double map?
You have a type called endpoints which in JSON will have a single field called "Endpoints",
which is a map with a single field called "endpoints".
You can see how rediculious this is getting.

Just make it a single map `{basePath: description, ...}`.

eriq-augustine, line 25: Put some JSON info on these fields.

The idea is that we just convert the entire description to JSON.
This way, we can later just put this field in our API response type and we're done.

internal/api/routes.go
```diff
@@ -31,3 +42,33 @@ func GetRoutes() *[]*core.Route {
...
 	return &routes
 }
+
+func Describe() *APIDescription {
+	routes := GetRoutes()
+
+	endpointMap := make(map[string]EndpointDescription)
+	for _, route := range *routes {
+		apiRoute, ok := route.(*core.APIRoute)
+		if ok {
```
eriq-augustine, line 52: Invert this.

internal/api/routes.go
```diff
@@ -31,3 +42,33 @@ func GetRoutes() *[]*core.Route {
...
+	}
+
+	return &APIDescription{
+		Endpoints: map[string]map[string]EndpointDescription{
+			"endpoints": endpointMap,
+		},
+	}
+}
+
+func DescribeToJSON(description *APIDescription) (string, error) {
```
eriq-augustine, line 68: This function doesn't seem necessary.
Also, we usually don't want to let library functions (in this case just not top-level packages) do our formatting for us.

eriq-augustine (changes requested):
Getting closer.

internal/api/core/route.go
Lucas-Ellenberger: I thought this was a bridge step. I interpreted your original comment that said it can be done on a later PR to mean leave the passed in description and prep by also commenting the description on the handler.

I will resolve this comment after I update the code and you have seen it.

internal/api/core/route.go
```diff
@@ -0,0 +1,109 @@
...
+
+			err = sendAPIResponse(nil, response, nil, apiErr, false)
+		}()
+
+		err = handleAPIEndpoint(response, request, apiHandler)
+
+		return err
+	}
+
+	fullPath := NewEndpoint(basePath)
```
eriq-augustine, line 92: 1) I think we have an opportunity to make this function unexported
(which would be cool, since we generally don't want people mucking around with other paths).

2) We can also use this as a chance to change it's name to something better (maybe something like "makeFullAPIPath()"?).

3) This would also involve changing some test things (since the tests are using it).
I'm thinking that you could change SendTestAPIRequest*() to accept a basePath instead of the full path and then just build to full path in the test function.
(So we would pass something like `users/list` instead of `/api/v03/users/list`.)

internal/api/core/route.go
```diff
@@ -0,0 +1,109 @@
...
+
+	_, requestType, responseType, err := validateAPIHandler(fullPath, apiHandler)
+	if err != nil {
+		log.Warn("Error while validating API handler.", err, log.NewAttr("endpoint", fullPath))
```
eriq-augustine, line 96: Boost this to an error.

internal/api/core/routing.go
```diff
@@ -272,50 +227,51 @@ func allocateAPIRequest(endpoint string, apiHandler ValidAPIHandler) (any, *APIE
 }
 
 // Reflexively ensure that the api handler is of the correct type/format (e.g. looks like APIHandler).
+// Returns the handler, the type of the input struct, the type of the first output struct, and an error.
```
eriq-augustine, line 230: "the type of the input struct" and "the type of the first output struct" sounds a bit awkward (especially that "first" (which implies there is a second?)).
Once again, you can just make it a bit more clear by simplifying it.
Maybe something like "the type of the handler's input, the type of the hander's output".

eriq-augustine (changes requested):
Looking good, just a few things.

internal/api/core/test.go
```diff
@@ -57,7 +57,7 @@ func SendTestAPIRequest(test *testing.T, endpoint string, fields map[string]any)
 // Provided fields will override base fields.
 // If an email is provided without an "@", we will suffix the email with the common test domain.
 func SendTestAPIRequestFull(test *testing.T, endpoint string, fields map[string]any, paths []string, email string) *APIResponse {
```
eriq-augustine, line 59: The variable name and comment need updating since you changed the semantics of it.

internal/api/server/unix_socket.go
```diff
@@ -114,6 +114,8 @@ func handleUnixSocketConnection(connection net.Conn) error {
 		return fmt.Errorf("Failed to find the 'endpoint' key in the request.")
 	}
 
+	endpoint = core.MakeFullAPIPath(endpoint)
```
eriq-augustine, line 117: The variable name (the input here) needs to change with the semantics (the output can stay the same).
You also need to add a comment so people known that they don't need to pass a full API path.
I'm thinking you can make the comment where `ENDPOINT_KEY` is defined.

cmd/describe-api/main.go
```diff
@@ -30,10 +30,7 @@ func main() {
 }
 
 func run() int {
-	description, err := util.ToJSONIndent(api.Describe())
-	if err != nil {
-		log.Fatal("Unable to convert API description to JSON.", err)
-	}
+	description := util.MustToJSONIndent(api.Describe())
 
 	fmt.Print(description)
```
eriq-augustine, line 35: This is just a suggestion, not required.

You can do this in one line without the variable.

eriq-augustine (changes requested):
Pretty close, just some documentation things.

eriq-augustine (approved):
lgtm!

PR #112, opened 2024-10-22
Added a new API endpoint to describe all of the API endpoints on the server.

Lucas-Ellenberger:
This PR adds a new API Endpoint that describes all API endpoints on the server by outputting the endpoints name, their description, request type, and response type.

This PR adds a new API package `metadata` which will be used for endpoints that provide information about the server itself. To implement the API endpoint, some code was relocated for better structure. Additionally, `RunServer()` injects the routes to a global variable in `api/core` to avoid cyclic dependencies. Since the testing infrastructure works on a package basis, we are unable to provide the full routing information in the new APIs test.

However, the full functionality of the API endpoint can be seen when running the server and testing via the associated python interface. So, this PR depends on the corresponding python interface implementation, which can found [here](https://github.com/edulinq/autograder-py/pull/35).


internal/api/core/describe.go
eriq-augustine, line 1: I like having this stuff in another file, but it will probably need to be in api istead of api/core.
This is because here is where you need to do all the real testing for Describe().

internal/api/core/describe.go
```diff
@@ -0,0 +1,40 @@
...
+
+func SetServerRoutes(routes []Route) {
+	serverRoutes = routes
+}
+
+func GetServerRoutes() []Route {
+	return serverRoutes
+}
+
+func Describe(routes []Route) *APIDescription {
```
eriq-augustine, line 23: Add in caching.

internal/api/metadata/describe_test.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	var responseContent DescribeResponse
+	util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
+
+	expected := DescribeResponse{*core.Describe(core.GetServerRoutes())}
+	if !reflect.DeepEqual(expected, responseContent) {
+		test.Fatalf("Unexpected API description. Expected: '%s', actual: '%s'.",
+			util.MustToJSONIndent(expected), util.MustToJSONIndent(responseContent))
+	}
```
eriq-augustine, line 24: The real test for this will probably need to compare against an actual, hand-checked JSON object.
This is a design thing to think about: how will we maintain this object.
It could be nice to just keep a JSON file in the repo that is the output of cmd/describe.
This will allow things (like this test and the Python side) to just grab this file whenever it needs to.
But, then the file needs to be maintained or this test will fail.
Maybe we have a script to automatically updates the file?
(Note that the script would be something super simple like: `go run cmd/describe-api/main.go > api.json`.)

eriq-augustine (changes requested):
The issues you are having here stem from design issues around tests.
Remember, if something smells it may be time to throw it out.
In this case, the issue is the level of testing at the API endpoint level.
Tests for API endpoints are sometimes long, but at their core they are about testing the endpoint itself.
We should have other tests to test the functionality that the endpoint uses.

For example, we have the users/upsert API endpoint test, which is decently sized.
But then we have the procedures/users/upsert test which is HUGE.

In this case, you are trying to test Describe() in metadata/describe,
and it's causing you some issues.
Instead, the test for metadata/describe should be pretty trivial (since the endpoint is pretty trivial, we really just need to make sure the types validate).,
and then do the serious testing in describe_test.go.

scripts/update_api_description.sh
```diff
@@ -0,0 +1,28 @@
...
+	if [[ $# -ne 0 ]] ; then
+        echo "USAGE: $0"
+        exit 1
+    fi
+
+    trap exit SIGINT
+
+	cd "${ROOT_DIR}"
+
+	go run cmd/describe-api/main.go > api.json
```
eriq-augustine, line 16: I don't think I like this file just hanging out in the top-level directory.
Maybe we should make a directory like "resources" and put it in there.

If we do that, we should probably move the version file as well.
Be careful, this requires changes to docker material as well (you should search for the filename).

.github/workflows/main.yml
```diff
@@ -25,6 +25,9 @@ jobs:
       - name: Build
         run: ./scripts/build.sh
 
+      - name: Update API Description
+        run: ./scripts/update_api_description.sh
```
eriq-augustine, line 29: This is no good and defeats the purpose of testing the API description.
You are generating it here and then just testing it against the same generation in the unittest ...
Nowhere did you actually validate that the test is correct.

The whole idea is that we need a person to verify that the API description is correct.
This means that you need an API file to always live in the repo.

The general workflow for someone working on an endpoint would be:
 - Work on endpoint.
 - Run tests, which fail because they forgot to update the API description.
 - Look at changes to description (could be done directly through the unit test, since they will see a diff there).
 - User accepts changes to to the API (or finds an issue and goes back and fixes it).
 - User runs `./scripts/update_api_description.sh` to codify their changes.
 - Run tests again (which should pass).
 - Done.

internal/api/describe.go
```diff
@@ -0,0 +1,38 @@
...
+	endpointMap := make(map[string]core.EndpointDescription)
+	for _, route := range routes {
+		apiRoute, ok := route.(*core.APIRoute)
+		if !ok {
+			continue
+		}
+
+		var requestTypeStr, responseTypeStr string
+
+		if apiRoute.RequestType != nil {
```
eriq-augustine, line 17: I don't think this should be possible.
Seems like an error.

internal/api/describe_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestDescribeFull(test *testing.T) {
+	path := filepath.Join(util.RootDirForTesting(), "api.json")
```
eriq-augustine, line 13: Filename should be a constant.

internal/api/describe_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+func TestDescribeEmptyRoutes(test *testing.T) {
+	routes := []core.Route{}
+	description := Describe(routes)
+
+	if len(description.Endpoints) != 0 {
+		test.Errorf("Unexpected number of endpoints. Expected: '0', actual: '%d'.", len(description.Endpoints))
+	}
+}
+
```
eriq-augustine, line 37: The tests starting here don't really seem necessary.
I know it sounds strange for someone to say "test less",
but there is a maintenance cost associated with each test so we want to make sure each test has a purpose.

 - TestDescribeNonAPIRoute -- The full routes already has non-api routes.
 - TestDescribeMultipleAPIRoutes -- The full routes already has multiple API routes.
 - TestDescribeDuplicateBasePaths -- This is technically allowed, since they may have different HTTP methods.
 - TestDescribeEmptyRequestAndResponseTypes -- This should not be possible.

internal/api/metadata/describe_test.go
```diff
@@ -0,0 +1,37 @@
...
+	description := core.APIDescription{
+		Endpoints: map[string]core.EndpointDescription{
+			"metadata/describe": core.EndpointDescription{
+				RequestType:  "*DescribeRequest",
+				ResponseType: "*DescribeResponse",
+			},
+		},
+	}
+
+	core.SetAPIDescription(description)
```
eriq-augustine, line 22: Make sure to clean this up in a defer.

eriq-augustine (changes requested):
The code is looking good,
but I think there is some fundamental confusion about the `api.json` file (see the commend in the CI to start).

internal/api/describe.go
```diff
@@ -0,0 +1,25 @@
+package api
+
+import (
+	"github.com/edulinq/autograder/internal/api/core"
+)
+
+// Routes must be validated before calling Describe.
```
eriq-augustine, line 7: Am I missing something, I don't see any Validate() function.

This is a little tricky.
The core of this is in NewAPIRoute(), where we expect to be able to get the types.
So a route is effectively "validated" there, which is fine.
But an incorrectly formatted API handler can cause us to log an error (but not actually return one).
Then that would route would cause issues here.
Note that then the API endpoint will also not validate, so this is something we can guarantee will be caught in tests.

Maybe we need to change NewAPIRoute() to MustNewAPIRoute()?
Can you verify that this would be safe?

eriq-augustine, line 7: Nit: When referring to a function by their name, we usually end it with parens,
e.g. "Describe()".

internal/api/describe_test.go
```diff
@@ -0,0 +1,74 @@
...
+func TestDescribeDuplicateBasePaths(test *testing.T) {
```
eriq-augustine, line 41: My previous comment still stands here.
Duplicate paths for routes are allowed (since they can have a different HTTP method).

Also, this test seems off to me.
Who decided the semantics that the second path should survive?

I think there are two good options here:
1) Remove this test and don't encourage people to have duplicate API endpoints.
2) Formalize the functionality for duplicate routes and check for them.

internal/api/metadata/describe_test.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"reflect"
+	"testing"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestMetadataDescribe(test *testing.T) {
+	oldDescription := core.GetAPIDescription()
```
eriq-augustine, line 12: Move this line to right above where you set the description (line 24).
We want those three lines to be one visual block:
"set the description just for this test".

internal/common/resources.go
```diff
@@ -0,0 +1,29 @@
+package common
+
+import (
+	"os"
+	"path/filepath"
+)
+
+const (
+	API_DESCRIPTION_FILENAME = "api.json"
+	RESOURCES_OUTPUT_DIRNAME = "resources"
```
eriq-augustine, line 10: It's not an "output" dir, it's just a normal dir.

internal/common/resources.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+func GetResourcesDir() (string, error) {
+	cwd, err := os.Getwd()
```
eriq-augustine, line 14: We generally don't use os.Getwd().
Instead, we use common.ShouldGetCWD().

internal/common/resources.go
eriq-augustine, line 1: We already have common/paths.go, this should probably just be in there.

But also, you don't have any other dependencies in this file, maybe it can go down to util.
The main difference between util and common is that common can use config.
If you are not using config, util may be the place to be.

internal/common/resources.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return filepath.Join(cwd, "../../", RESOURCES_OUTPUT_DIRNAME), nil
```
eriq-augustine, line 19: !
Careful, subtle bug here.

You can't use "../.." in a path join, since you already joined two things in that string.
You need to join each "..".
For example:
https://github.com/edulinq/autograder-server/blob/main/internal/util/version.go#L53

internal/util/version.go
```diff
@@ -50,7 +50,7 @@ func readVersion() (Version, error) {
 		Api:     UNKNOWN_API,
 	}
 
-	versionPath := ShouldAbs(filepath.Join(ShouldGetThisDir(), "..", "..", VERSION_FILENAME))
+	versionPath := ShouldAbs(filepath.Join(ShouldGetThisDir(), "..", "..", "resources", VERSION_FILENAME))
```
eriq-augustine, line 53: WTF?
"What the funk?"

You make GetResourcesDir() and then don't use it?
Maybe it belongs here instead.

scripts/update_api_description.sh
```diff
@@ -0,0 +1,32 @@
...
+    fi
+
+    trap exit SIGINT
+
+    cd "${ROOT_DIR}"
+
+    mkdir -p "${OUT_DIR}"
+
+    go run cmd/describe-api/main.go > "${OUT_FILE}"
+
```
eriq-augustine, line 21: Remove this newline.
When you check (or store) the exit status, you usually want to do that right against the line creating the status.

scripts/update_api_description.sh
```diff
@@ -0,0 +1,32 @@
...
+    if [ $? -eq 0 ]; then
+        echo "API description successfully updated in api.json."
```
eriq-augustine, line 23: Use your variable in these strings.

scripts/update_api_description.sh
```diff
@@ -0,0 +1,32 @@
...
+    else
+        echo "Failed to update api.json."
+        return 1
```
eriq-augustine, line 26: Since there is a return here and the conditional branches are not symmetric (they both don't return),
I would invert this conditional and only have one case (which returns).
Then the success case is just the default.

eriq-augustine (changes requested):
There's a couple subtle things here.

internal/api/core/routing_test.go
```diff
@@ -1,20 +1,49 @@
...
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/edulinq/autograder/internal/common"
 	"github.com/edulinq/autograder/internal/util"
 )
 
+// Define all the malformed handlers.
+var handlerTestCases = []malformedHandlerTestCase{
```
eriq-augustine, line 19: Only one test function uses this, why is it not defined inside that function?

internal/api/core/routing_test.go
```diff
@@ -24,47 +53,84 @@ func TestAPIPanic(test *testing.T) {
...
 	response := SendTestAPIRequest(test, endpoint, nil)
 	if response.Locator != "-001" {
-		test.Fatalf("Response does not have panic locator of '-501', actual locator: '%s'.", response.Locator)
+		test.Fatalf("Response does not have panic locator of '-001', actual locator: '%s'.", response.Locator)
 	}
 }
 
 // Test handlers that do not have the correct signature.
 // Specifically, we are focusing on testing validateAPIHandler().
 func TestMalformedHandlers(test *testing.T) {
```
eriq-augustine, line 66: This test seems too complex in general.
Instead, why not just change log.Fatal():
https://github.com/edulinq/autograder-server/blob/main/internal/log/public.go#L35
It uses os.Exit(), but now-a-days everyone should be using util.Exit():
https://github.com/edulinq/autograder-server/blob/main/internal/util/exit.go#L14
Then you can just see if an exit happened.

The string/output checking seems a bit too complex as well.
We really just need to see if an exit happened.
Checking the string would be nice, but not if it makes the test complex.

eriq-augustine (changes requested):
I think a simplification would be nice.

cmd/peek-submission/main.go
```diff
@@ -89,7 +89,7 @@ func main() {
 
 		fmt.Println(output)
 
-		util.Exit(2)
+		log.Exit(2)
```
eriq-augustine, line 92: Leave this one as is.
Oliver already has changes lined up in #101, and we'll be nice and try to avoid conflicts for him.

internal/log/exit.go
```diff
@@ -12,7 +10,7 @@ var (
 )
 
 func Exit(code int) {
```
eriq-augustine, line 12: Add a comment about why this is here (and not just in util).

internal/log/exit.go
eriq-augustine, line 1: This should probably also exit in util.

It's a bit strange, but when designing projects we also want to think about how people (developers) will react when using our code.
If when people wanted to exit, they had to call into the `log` package, they would probably think that was strange.
You have a good reason for it, but it will seem strange without all the context.

So I suggest this:
 - Rename log.Exit() to something that encourages people not to mindlessly call it, like "ExitForUtil()" or "ExitJustForLoggingAndUtilYouProbablyShouldntBeCallingThis()".
 - Keep util.Exit() which just calls log.ExitForUtil().
 - Comment the behavior in log.
 - Encourage everyone to call util.Exit().

internal/api/core/routing_test.go
```diff
@@ -24,46 +25,58 @@ func TestAPIPanic(test *testing.T) {
...
+	// Suppress exits to capture exit codes.
+	log.ShouldExitForTesting = false
+	defer func() {
+		log.ShouldExitForTesting = true
+	}()
+
+	// Suppress stderr to ignore Fatal log messages.
+	oldTextWriter := log.GetTextWriter()
+	log.SetTextWriter(nil)
+	defer log.SetTextWriter(oldTextWriter)
```
eriq-augustine, line 48: Why not just set the text level to `log.LevelOff`?

internal/log/logging.go
```diff
@@ -60,6 +60,10 @@ func SetTextWriter(newTextWriter io.StringWriter) {
 	textWriter = newTextWriter
 }
 
+func GetTextWriter() io.StringWriter {
+	return textWriter
+}
+
```
eriq-augustine, line 66: I think this may not be necessary.
I commented on the only place this is used.

internal/api/core/route.go
```diff
@@ -93,7 +93,7 @@ func NewAPIRoute(basePath string, apiHandler any) *APIRoute {
 
 	_, requestType, responseType, err := validateAPIHandler(fullPath, apiHandler)
 	if err != nil {
-		log.Error("Error while validating API handler.", err, log.NewAttr("endpoint", fullPath))
+		log.FatalWithCode(3, "Error while validating API handler.", err, log.NewAttr("endpoint", fullPath))
```
eriq-augustine, line 96: I think your too deep to get to choose your exit status.
If you want, we can come up with a standard list.
Really, we would probably be following (sysexits.h)[https://man.freebsd.org/cgi/man.cgi?query=sysexits&sektion=3].

eriq-augustine (changes requested):
Nice!
This infrastructure is looking real strong, a good improvement to the overall system.

There's some movement that should be done, but nothing big.

internal/log/exit.go
```diff
@@ -0,0 +1,28 @@
...
+
+var (
+	lastExitCode         = 0
+	ShouldExitForTesting = true
+)
+
+// To properly test fatal log errors, we need to capture the last exit code.
+// Set ShouldExitForTesting to false and compare the last exit code.
+// This function ONLY exists to avoid a dependency cycles and should only be called within log.
+// EVERYONE should use util.Exit().
```
eriq-augustine, line 15: You hold off the spice a bit here and just do normal casing for "everyone".

internal/log/exit.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"os"
+)
+
+var (
+	lastExitCode         = 0
+	ShouldExitForTesting = true
+)
+
+// To properly test fatal log errors, we need to capture the last exit code.
```
eriq-augustine, line 12: This is not true.
It is not just fatal logs, any exit.

internal/util/exit.go
```diff
@@ -1,26 +1,18 @@
...
-
 	"github.com/edulinq/autograder/internal/log"
 )
 
-var (
-	lastExitCode         = 0
-	ShouldExitForTesting = true
-)
-
+// Everyone should use Exit() for coherent code.
```
eriq-augustine, line 7: "Coherent code" is a bit abstract.
I think something more direct would be better, like:
"Everyone should use Exit() instead of exiting via something like os.Exit().
This allows for easier testing."

eriq-augustine (changes requested):
All the core is here.
There are still some package design issues.

eriq-augustine (changes requested):
Just two small things.

eriq-augustine (approved):
lgtm!

PR #116, opened 2024-11-04
Added a cmd to remove a server user.

OliverLok:


cmd/user-remove/main.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/edulinq/autograder/internal/cmd"
+	"github.com/edulinq/autograder/internal/config"
+	"github.com/edulinq/autograder/internal/log"
+)
+
+var args struct {
+	config.ConfigArgs
+	cmd.CommonOptions
+
+	TargetEmail string `help:"Email of the user to remove." arg:""`
```
Lucas-Ellenberger, line 17: I am wondering if it's worthwhile to create common CLI options (similar to the [API config](https://github.com/edulinq/autograder-py/blob/9642fa80dd07925f94bed620fe12497524711f32/autograder/api/config.py#L264C1-L266C25) or [CLI config](https://github.com/edulinq/autograder-py/blob/main/autograder/cli/config.py) on the python side).

You could include more next to structs (similar to `CommonOptions`). I'm not sure if you can update the field tag (the help message) through a variable, so your help messages may be more generic (which would mirror the python interface).

Let me know if you think this is a good idea!

Lucas-Ellenberger (commented):
I left a note on a possible way to share code. This would apply to all PRs for this round, so I'll hold off reviewing the others.

cmd/user-remove/main.go
Lucas-Ellenberger: Oliver and I discussed this and have decided to hold off. Currently, only TargetEmail is shared enough to be worthwhile. We can reassess if we notice more CLI args are being shared between cmds.

Lucas-Ellenberger (approved):
lgtm!

eriq-augustine (approved):
lgtm

PR #121, opened 2024-11-07
Added the ability for CMDs to start and stop their own server.

OliverLok:


internal/cmd/test.go
```diff
@@ -83,6 +95,44 @@ func getUnusedPort() (int, error) {
...
+		if err != nil {
+			log.Fatal("Failed to start the server.", err)
+		}
+	}()
+
+	// Small sleep to allow the server to start up.
+	time.Sleep(100 * time.Millisecond)
+}
+
+func MustStopCMDServer() {
```
Lucas-Ellenberger, line 123: What makes this a `Must` function? It looks like we can log errors in `server.StopServer()`, but this cannot return errors (so it should never panic).

I think you want to mirror the function name of `MustStartCMDServer()` (which should be a `Must` because it can Fatal / panic). But, we probably shouldn't add the `Must` to `StopCMDServer()` since other people might get scared to call this code if they do not want to panic.

internal/cmd/request.go
```diff
@@ -24,11 +24,20 @@ func MustHandleCMDRequestAndExit(endpoint string, request any, responseType any)
 }
 
 func MustHandleCMDRequestAndExitFull(endpoint string, request any, responseType any, options CommonOptions, customPrintFunc CustomResponseFormatter) {
+	if !IsTesting() {
+		MustStartCMDServer()
```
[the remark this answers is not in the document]
OliverLok: Defers won't work since there is an exit.Exit() at the end which just terminates the program without letting defers run.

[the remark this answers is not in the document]
Lucas-Ellenberger: Right.

internal/cmd/request.go
```diff
@@ -24,11 +24,20 @@ func MustHandleCMDRequestAndExit(endpoint string, request any, responseType any)
...
+	}
+
 	response, err := SendCMDRequest(endpoint, request)
 	if err != nil {
+		MustStopCMDServer()
```
Lucas-Ellenberger, line 33: What happens if we call `MustStopCMDServer()` when we are not in testing mode (i.e. if this function did not call `MustStartCMDServer()`?

OliverLok: I'm not sure I understand. If we aren't in testing mode the only other way we'd get here is by running the cmd with go run (which would in turn call `MustStartCMDServer()`). I don't see anywhere else that would use these cmd start/stop servers, but even if there were, stopping the server is designed to be able to be called as many times as we want without causing an error/panic. You can see something similar being done [here](https://github.com/edulinq/autograder-server/blob/main/internal/cmd/test.go#L36) where we stop the server before starting it just in case the previous test wasn't able to stop it for some reason.

Lucas-Ellenberger: kk

Lucas-Ellenberger (approved):
lgtm!

internal/cmd/test.go
```diff
@@ -83,6 +94,46 @@ func getUnusedPort() (int, error) {
...
+		}
+	}()
+
+	serverStart.Wait()
+
+	// Small sleep to allow the server to start up.
+	time.Sleep(100 * time.Millisecond)
+}
+
+func StopCMDServer() {
```
eriq-augustine, line 124: On general, you don't want to export functions only used within the package.

eriq-augustine (changes requested):
I think there is some structural work to do here.
I see three main situations/cases:
1. A server is already running (already taken care of).
2. A server is not running and we want to start one then stop it after we are done.
3. A server is already running, but it was started by a CMD and we want to error out.

It seems you currently have two problems.

First (starting the server).
When starting the server, you need to distinguish between 1 and 3.
For 1 you want to connect to the server,
and for 3 you want to error out (fatal).
So you effectively need a way for either a server to reject connections from a CMD or for a CMD to recognize if it is allowed to connect to a server.

Second (stopping the sevrer).
When stopping the server, you need a way to distinguish between 1 and 2.
In 1, you just want to leave the server running.
This is like in your CMD unit tests where the individual tests (and therefor the CMD's main) doesn't need to worry about killing the server.
In 2, you need to remember that you created the server and need to stop it.
I have some relevant comments to this in this PR, but I think you can have a function invoked in MustHandleCMDRequestAndExitFull() that ensures the server is running and return a boolean indicating if a server was actually started (or one is already running).
Then you can just defer stopping the server based on the return.

Also, you will probably want at least one test here.
It's a bit tricky, but you probably need a test that calls a CMD without using CMDServerTestingMain() (since that already starts a server).
You'll need to think hard about this one.

internal/cmd/test.go
```diff
@@ -24,6 +25,11 @@ const (
 	STDERR_FILENAME = "stderr.txt"
 )
 
+var (
+	isTesting bool
```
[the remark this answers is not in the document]
OliverLok: If i used config.UNIT_TESTING_MODE to tell when the server shouldn't run for a test, I wouldn't be able to run a cmd regularly if I used --unit-testing which is why I did it like that. I could create another config value that only gets set in CMDServerTestingMain I suppose.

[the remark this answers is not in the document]
eriq-augustine: That seems like a fine reason not to use `config.UNIT_TESTING_MODE`, but that doesn't mean you need another config option.
If you take into account all the other comments, then you probably won't need an indicator like this.

internal/cmd/test.go
```diff
@@ -24,6 +26,10 @@ const (
 	STDERR_FILENAME = "stderr.txt"
 )
 
+var (
+	oldPort int
```
eriq-augustine, line 30: I don't think this is necessary as a package variable.
You always want to keep things at the smallest scope reasonable.

eriq-augustine (changes requested):
The core design work here is good.

But, there were some sloppy mistakes.
Don't let your guard down.

internal/cmd/test.go
OliverLok: I think I'm failing to see an effective alternative. One approach I thought of was getting the current/about to be old port and passing it as a parameter to stopCMDServer() and startCMDServer() [here](https://github.com/edulinq/autograder-server/pull/121/files#diff-1ed89e8c5cba2a2c529592a3018a85556e9f2f3f2f591e1301b6ac026da707ecR31-R32), but I feel like it's cleaner to use a variable. What do you think?

eriq-augustine: You already have code that ensures the server is running for CMDs, why can't that set the port and defer the reset back to the old port?

Or, if the server knows what the initiator is, why can't it just choose a random port there?

OliverLok: Ok so setting the port and deferring resetting it in the anonymous func? I can't set and defer it in startCMDServer() because the function returns (defers back to the old port) before the CMD gets to sending the request.

eriq-augustine: If you are going that route, you already do most of this:
https://github.com/edulinq/autograder-server/blob/a4ca15d6f837cc7354cdc008b6fcbf89f03a8f67/internal/cmd/request.go#L32

internal/procedures/server/start.go
```diff
@@ -46,10 +47,13 @@ func Start() (err error) {
 		go startCourse(course)
 	}
 
-	// Cleanup any temp dirs.
-	defer util.RemoveRecordedTempDirs()
+	// Keep temp dirs for the CMD testing infrastructure to access after this function ends.
+	if !config.UNIT_TESTING_MODE.Get() {
+		// Cleanup any temp dirs.
+		defer util.RemoveRecordedTempDirs()
+	}
```
eriq-augustine, line 54: Is this necessary?
You call pserver.RunSever(), not pserver.Start() when you start with `common.CMD_TEST_SERVER`.

If it is necessary, I think there is a better way.
You don't want to directly check `config.UNIT_TESTING_MODE` if you don't have to,
checking it represents a special case for testing.
But here, we can just pass a `cleanup` boolean argument which makes sense to have in any context.

eriq-augustine (changes requested):
The core design here all looks great.

internal/procedures/server/start.go
OliverLok: It is necessary for the one test where the cmd starts its own server. Should the comment better reflect this one case or is it fine to keep it saying it's for the CMD testing infrastructure because that one case also uses the infrastructure?

internal/cmd/cmd_test.go
```diff
@@ -0,0 +1,48 @@
...
+	for _, testCase := range expectedSubstrings {
+		if !strings.Contains(stderr, testCase.expectedSubstring) {
+			test.Errorf("Expected substring '%s' not found in stderr: '%s'.", testCase.expectedSubstring, stderr)
+		}
+	}
+}
+
+func userList() {
+	MustHandleCMDRequestAndExit(`users/list`, users.ListRequest{}, users.ListResponse{})
+}
```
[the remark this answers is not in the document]
OliverLok: That was my initial thought but go doesn't let you import main packages if I wanted to use that shared function here. I suppose I could have main() in cmd/users-list/main.go call a function from a different package that does everything main is doing so that function would be shared, but that doesn't follow the pattern for the rest of the CMDs.

internal/cmd/cmd_test.go
```diff
@@ -0,0 +1,48 @@
...
+	defer util.RemoveRecordedTempDirs()
+
+	expectedSubstrings := []struct {
+		expectedSubstring string
+	}{
+		{`API Server Started.`},
+		{`Unix Socket Server Started.`},
+		{`API Server Stopped.`},
+		{`Unix Socket Server Stopped.`},
+	}
```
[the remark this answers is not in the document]
OliverLok: Would it be safe to assume then that if userlist comes back with the expected list of users, that means it started it's own server? If that's the case then it may not be necessary to have these expected substrings, because as you said, they can easily change often.

cmd/user-auth/expected_output_test.go
eriq-augustine, line 1: I don't like this file.
We have files like this when we have a lot of output, no need for only a little output.

cmd/user-auth/main.go
```diff
@@ -1,22 +1,22 @@
...
 )
 
 var args struct {
 	config.ConfigArgs
+	cmd.CommonOptions
 
-	Email string `help:"Email for the user." arg:"" required:""`
-	Token string `help:"Token for the user." arg:"" required:""`
+	Email    string `help:"Email for the user." arg:"" required:""`
+	Password string `help:"Password for the user." arg:"" required:""`
```
eriq-augustine, line 19: Why the change?
"Tokens" is technically more correct:
https://github.com/edulinq/autograder-server/blob/main/internal/model/user_server.go#L36

cmd/user-auth/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+		return suite.Run()
+	}()
+
+	exit.Exit(code)
+}
+
+// This test ensures a CMD can start its own server.
+// It must remain or be replaced with an equivalent test if removed.
+// CMDs always succeed in user authentication, regardless of credentials, so only one test case is needed.
```
eriq-augustine, line 26: This comment belongs on the test case, not the test function.

cmd/user-auth/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+		defer db.CleanupTestingMain()
+
+		return suite.Run()
+	}()
+
+	exit.Exit(code)
+}
+
+// This test ensures a CMD can start its own server.
+// It must remain or be replaced with an equivalent test if removed.
```
eriq-augustine, line 25: I don't like this sentence.
You just say that the test needs be be here but not WHY it is important.
One of the main reasons for comments is to explain tricky things to readers,
this comment doesn't do that at all.

internal/procedures/server/start.go
```diff
@@ -74,3 +78,57 @@ func startCourse(course *model.Course) {
 		log.Error("Failed to update course.", err, course)
 	}
 }
+
+// Check to see if a server is running and start one if it's not.
+// Returns (false, 0) if a the primary server or cmd test server is already running
+// or (true, oldPort) if it started it's own server.
+func MustEnsureServerIsRunning() (bool, int) {
```
eriq-augustine, line 85: !!

I don't like this here, and I am concerned that you moved it here.
The previous comment on this function was that it should be in another file.
I didn't say another **package**.

This may seem like a small thing, but I see this as an orange flag (very close to red).
This function is not used by anything in this package and now has to be exported.
It's already a "must" function, so we would want to only export if absolutely necessary (which it is as it currently stands since it is in the wrong package).

internal/common/server_status.go
```diff
@@ -11,24 +11,33 @@ import (
...
+	CMD_SERVER      ServerInitiator = "cmd-server"
+	CMD_TEST_SERVER ServerInitiator = "cmd-test-server"
+)
+
 type StatusInfo struct {
-	Pid            int    `json:"pid"`
-	UnixSocketPath string `json:"unix_socket_path"`
+	Pid             int             `json:"pid"`
+	UnixSocketPath  string          `json:"unix-socket-path"`
+	ServerInitiator ServerInitiator `json:"server-initiator"`
```
eriq-augustine, line 31: Just call it "Initiator".

You already call it that elsewhere, and we try to avoid variables/members with the same names as types (if possible).

internal/common/server_status.go
```diff
@@ -48,31 +57,32 @@ func GetUnixSocketPath() (string, error) {
...
-	ok, err := checkAndHandleStalePid()
+	statusInfo, err := CheckAndHandleServerStatusFile()
 	if err != nil {
 		return err
 	}
 
-	if !ok {
-		return fmt.Errorf("Failed to create the status file '%s'.", statusPath)
+	if statusInfo != nil {
+		return fmt.Errorf("Failed to create the status file '%s' since another server is running.", statusPath)
```
eriq-augustine, line 73: This comment is inaccurate.
It didn't fail to create another status file, it refused to do so.
Be more direct here, like:
"Cannot start server, another server is running (pid %d).".

internal/procedures/server/start.go
```diff
@@ -74,3 +78,57 @@ func startCourse(course *model.Course) {
 		log.Error("Failed to update course.", err, course)
 	}
 }
+
+// Check to see if a server is running and start one if it's not.
+// Returns (false, 0) if a the primary server or cmd test server is already running
+// or (true, oldPort) if it started it's own server.
```
eriq-augustine, line 84: This function (and comment) is missing a case.
What happens if a CMD_SERVER is running?

eriq-augustine, line 84: "it" is ambiguous here.

eriq-augustine (changes requested):

internal/procedures/server/start.go
OliverLok: If a CMD_SERVER is running, another CMD_SERVER will not be able to start since it will get stopped from starting when it reads the status file and finds out the pid in the status file is still active (isAlive()). Should I add this as a comment even though this function isn't the one doing that checking?

eriq-augustine: You are relying on other code to set some external state and hopefully this function never gets called in that case.
You can see how this is pretty fragile.
Specifically, you are not [programming defensively](https://en.wikipedia.org/wiki/Defensive_programming), which you should generally be doing.
When programming defensively with enums, that usually means ensuring that all cases are accounted for.

cmd/user-auth/main.go
```diff
@@ -1,22 +1,22 @@
...
 )
 
 var args struct {
 	config.ConfigArgs
+	cmd.CommonOptions
 
-	Email string `help:"Email for the user." arg:"" required:""`
-	Token string `help:"Token for the user." arg:"" required:""`
+	Email  string `help:"Email for the user." arg:"" required:""`
+	Tokens string `help:"Tokens for the user." arg:"" required:""`
```
eriq-augustine, line 19: "Tokens" should be singular.

cmd/user-auth/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/edulinq/autograder/internal/cmd"
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/exit"
+)
+
+const SUCCESS_AUTH = `{
+    "success": true
+}
+`
```
eriq-augustine, line 15: Since this is expected test data and not a true constant, lower case it and put it at the bottom of the file.

Also change the name to indicate that it is expected output.

cmd/user-auth/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	exit.Exit(code)
+}
+
+// This test ensures a CMD can start its own server.
+// Since this is the only test that ensures a CMD can start its own server,
+// it must remain or be replaced with an equivalent test if removed.
+func TestUserAuthBase(test *testing.T) {
+	// CMDs always succeed in user authentication, regardless of credentials, so only one test case is needed.
+	testCase := []struct {
```
eriq-augustine, line 34: Since there is only one possible test case, we don't need a loop here.

internal/cmd/handler.go
eriq-augustine, line 1: I don't like the name "handler"
(since we already have a bunch of handlers for other things (like the API)).

internal/cmd/handler.go
```diff
@@ -99,3 +116,63 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+	if statusInfo != nil {
+		// Don't start the CMD server if the primary server or CMD test server is running.
+		if statusInfo.Initiator == common.PRIMARY_SERVER || statusInfo.Initiator == common.CMD_TEST_SERVER {
+			return false, 0
+		}
+
+		// log.Fatal() if another CMD server is running since they share the same working directory.
+		if statusInfo.Initiator == common.CMD_SERVER {
+			log.Fatal("Cannot start server, another CMD server is running.", log.NewAttr("PID", statusInfo.Pid))
+		}
```
eriq-augustine, line 139: We will want to change up the structure here.
Note that it looks like there is a way for the code to get past both the conditionals without exiting the function, but there is not.
The core issue is that you have two mutually exclusive cases not in the same if tree.

I would use a switch here (switches are usually a go-to for enums).

internal/cmd/handler.go
```diff
@@ -99,3 +116,63 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+
+	serverStart.Wait()
+
+	// Small sleep to allow the server to start up.
+	time.Sleep(100 * time.Millisecond)
+
+	return true, oldPort
+}
+
+func getUnusedPort() (int, error) {
```
eriq-augustine, line 170: Let's move this to util.

internal/common/server_status.go
```diff
@@ -11,24 +11,33 @@ import (
...
+	CMD_SERVER      ServerInitiator = "cmd-server"
+	CMD_TEST_SERVER ServerInitiator = "cmd-test-server"
+)
+
 type StatusInfo struct {
-	Pid            int    `json:"pid"`
-	UnixSocketPath string `json:"unix_socket_path"`
+	Pid            int             `json:"pid"`
+	UnixSocketPath string          `json:"unix-socket-path"`
+	Initiator      ServerInitiator `json:"server-initiator"`
```
eriq-augustine, line 31: nit

I would remove the "server-" prefix from the JSON name.

internal/common/server_status.go
```diff
@@ -48,31 +57,32 @@ func GetUnixSocketPath() (string, error) {
 	return statusJson.UnixSocketPath, nil
 }
 
-func WriteAndHandleStatusFile() error {
-	Lock(PID_SOCK_LOCK)
-	Unlock(PID_SOCK_LOCK)
+func WriteAndHandleStatusFile(initiator ServerInitiator) error {
+	Lock(SERVER_STATUS_LOCK)
+	Unlock(SERVER_STATUS_LOCK)
```
eriq-augustine, line 62: This should probably be defered.

internal/common/server_status.go
```diff
@@ -82,34 +92,33 @@ func WriteAndHandleStatusFile() error {
 	return nil
 }
 
-// Returns (true, nil) if it's safe to create the status file,
-// (false, nil) if another instance of the server is running,
-// or (false, err) if there are issues reading or removing the status file.
-func checkAndHandleStalePid() (bool, error) {
+// Returns (nil, nil) if the status file doesn't exist,
+// (&statusJson, nil) if the status file exists and another instance of the server is running,
+// or (nil, err) if there are issues reading or removing the status file.
```
eriq-augustine, line 97: You can streamline this a bit by just saying something like:
"When there is no error, ...".
This just assumes (fairly) that there will always be a nil returned for the first value on an error.

eriq-augustine (changes requested):
Making good progress, mainly small things.

cmd/user-auth/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+// This test ensures a CMD can start its own server.
+// Since this is the only test that ensures a CMD can start its own server,
+// it must remain or be replaced with an equivalent test if removed.
+func TestUserAuthBase(test *testing.T) {
+	// CMDs always succeed in user authentication, regardless of credentials, so only one test case is needed.
+	testCase := struct {
+		cmd.CommonCMDTestCase
+		targetEmail string
+		targetPass  string
+	}{
```
eriq-augustine, line 32: Why have this struct if there is only one test case?
This struct only makes sense if you are looping over them.

internal/common/server_status.go
```diff
@@ -11,24 +11,33 @@ import (
...
+	SERVER_STATUS_LOCK             = "internal.common.SERVER_STATUS_LOCK"
 	STATUS_FILENAME                = "status.json"
 	UNIX_SOCKET_RANDNUM_SIZE_BYTES = 32
 )
 
+const (
+	PRIMARY_SERVER  ServerInitiator = "primary-server"
+	CMD_SERVER      ServerInitiator = "cmd-server"
+	CMD_TEST_SERVER ServerInitiator = "cmd-test-server"
+)
```
eriq-augustine, line 26: Move this so that it is directly under the type declaration (with a newline between them).
That way, you declare the type and right away enumerate its possible values.

eriq-augustine (changes requested):

eriq-augustine (approved):
lgtm!

PR #122, opened 2024-11-22
Added support for globs in path-type FileSpecs.

OliverLok:


internal/common/filespec.go
```diff
@@ -47,6 +48,11 @@ func (this *FileSpec) Validate() error {
 		if this.Path == "" {
 			return fmt.Errorf("A path FileSpec cannot have an empty path.")
 		}
+
+		_, err := filepath.Glob(this.Path)
```
eriq-augustine, line 52: This is not good.
Although it technically works, it is excessive (and a small potential security issue).

First thing to note is that FileSpecs can be validated at any time, but that is not when they are executed/used (often inside of a Docker container).
Additionally many file specs can (and are required to be) relative paths.
So here you have no idea what paths is being traversed.
Although filepath.Glob() only returns errors on bad validation, you can be using a lot of time and memory traversing paths.
(This is where the security issue comes in, since this information could be used in a timing attack.)

Instead, go to the source and see how filepath.Glob() validates.

internal/common/filespec.go
```diff
@@ -47,6 +48,11 @@ func (this *FileSpec) Validate() error {
...
+		if err != nil {
+			return fmt.Errorf("Invalid glob pattern '%s': '%w'.", this.Path, err)
```
eriq-augustine, line 54: This error will be confusing for people who are not using (or don't know about) globs.
They didn't select a "glob-type" FileSpec.

internal/common/filespec.go
```diff
@@ -209,6 +215,7 @@ func (this *FileSpec) GetPath() string {
 // Copy the target of this FileSpec in the specified location.
 // If the FileSpec has a dest, then that will be the name of the resultant dirent within destDir.
 // If the filespec is a path, then copy a dirent.
+// If the path includes a glob pattern, copy all matching dirents.
```
eriq-augustine, line 218: I don't like how this looks like a different case than a vanilla (non-glob) path.

internal/common/filespec.go
```diff
@@ -219,7 +226,25 @@ func (this *FileSpec) CopyTarget(baseDir string, destDir string, onlyContents bo
...
+		}
+
+		files, err := this.matchFiles()
+		if err != nil {
+			return fmt.Errorf("Failed to resolve glob in path '%s': '%w'.", this.Path, err)
+		}
+
+		// Loop over each matched file and create a temporary FileSpec to copy it to the destination.
+		for _, file := range files {
+			tempFileSpec := &FileSpec{Type: FILESPEC_TYPE_PATH, Path: file}
```
eriq-augustine, line 240: Why have a temp FileSpec?

Just so you can get access to copyPath()?
If so, then what you are doing here should "smell" to you.
You are wrapping an object to call a method that is in the package/file/type that you are currently working on.
You are making a work around for your own code.
Just change it so you have access to the method you need.

eriq-augustine (changes requested):
A good start, but this only scratched the surface of tests.

This PR also needs to come with documentation updates.
We have existing documentation on FileSpecs, so it will need to be updated to talk about the globbing semantics:
https://github.com/edulinq/autograder-server/blob/main/docs/types.md#file-specification-filespec

docs/types.md
```diff
@@ -767,7 +767,7 @@ Submission limits are not enforced for course graders.
 ## File Specification (FileSpec)
 
 A file specification (FileSpec) defines how to access a specific file (or dir).
-At it easiest it is just the path to a file, but can also specify things like a URL or Git repository.
+At its easiest, it is just the path to a file, but can also specify things like a URL, a Git repository, or a glob pattern for matching multiple files or directories.
```
eriq-augustine, line 770: "glob pattern" should link to some documentation on globs (Wikipedia, probably).

docs/types.md
```diff
@@ -782,7 +782,7 @@ All types of FileSpecs share some common fields:
 
 ### FileSpec -- Path
 
-A FileSpec with `type` equal to `path` points to an absolute or relative path accessible from the current machine.
+A FileSpec with `type` equal to `path` points to an absolute or relative path accessible from the current machine. The absolute or relative path may also include a glob pattern to target multiple files or directories.
```
eriq-augustine, line 785: Watch the first rule of code style: be consistent with the code around you.
We break lines by sentence (or clause for long sentences).
This allows us to have better diffs (e.g., if you changed one work in a huge paragraph, you would get a hard to read diff).

eriq-augustine, line 785: s/absolute or relative path/path/

KISS
Documentation follows the same rule as code.

internal/common/filespec.go
```diff
@@ -208,7 +213,7 @@ func (this *FileSpec) GetPath() string {
 
 // Copy the target of this FileSpec in the specified location.
 // If the FileSpec has a dest, then that will be the name of the resultant dirent within destDir.
-// If the filespec is a path, then copy a dirent.
+// If the filespec is a path, then copy a single dirent or all matching dirents if a glob pattern was used.
```
eriq-augustine, line 216: I don't like the second part of this comment, it makes this seem like two cases,
but it is only one: "copy all matching dirents".

internal/common/filespec.go
```diff
@@ -219,7 +224,24 @@ func (this *FileSpec) CopyTarget(baseDir string, destDir string, onlyContents bo
 		// no-op.
 		return nil
 	case FILESPEC_TYPE_PATH:
-		return this.copyPath(baseDir, destDir, onlyContents)
+		if !strings.ContainsAny(this.Path, `[]\*^?`) {
```
eriq-augustine, line 227: I still don't like this.
You try to separate glob and non-glob paths, but then you include non-glob paths in the glob case, e.g., `/some/file/with_a_\*_star.txt`.
Making it seem like you separate the code into two cases when you don't is a recipe for trouble and hard to debug issues.

internal/common/filespec_test.go
```diff
@@ -146,6 +146,81 @@ func getCopyTestCases() []*testCaseCopy {
 			true,
 			"720b5768bf364f35c316976e549f1bcd",
 		},
+		&testCaseCopy{
+			FileSpec{Type: "path", Path: filepath.Join(config.GetTestdataDir(), "files", "*_test"), Dest: "dest"},
+			true,
+			"87efd29f124f9a5b617517041d4d9b32",
```
eriq-augustine, line 152: These tests are starting to get complex enough that I think we need another way to handle the expected output to make it more readable.
Think through ways you may want to do it.
It's a big hard because there can be multiple files and dirs.

Maybe we assume that the test dir is the base (relative dir), so we no longer have to supply it.
Then the expected could be a `map[string]any` (or JSON) that describes what the result should look like.
Dirs could be keys for their contents, and files could be keys for some constant that represents each file.
```
# Single file
map[string]any{
    "test.txt": SPEC_TXT,
}

# Directory with file.
map[string]any{
    "filespec_test": {
        "test.txt": SPEC_TXT,
    },
}
```

There's several ways you could make the expected output more clear,
the above is just one suggestion.

testdata/files/*filespec_test/*spec.txt
eriq-augustine, line 1: No need to use names that are close to the existing ones.

eriq-augustine (changes requested):
Test cases are looking better.

internal/common/filespec_test.go
```diff
@@ -93,21 +95,23 @@ func TestFileSpecCopy(test *testing.T) {
...
 			continue
 		}
 
-		md5, err := util.MD5FileHex(zipPath)
-		if err != nil {
-			test.Errorf("Case %d: Failed to get zip's MD5 hash: '%v'.", i, err)
-			continue
+		copiedDirents := []string{}
+		for _, dirent := range dirents {
+			parts := strings.SplitN(dirent, "dest/", 2)
```
eriq-augustine, line 106: Hardcoded path separator.

eriq-augustine (changes requested):
The structure of the tests is looking good.

You're missing some tests the represent the most common glob scenarios:
 - `*`
 - `*.ext` (replace "ext" with whatever extension).

docs/types.md
```diff
@@ -783,6 +784,7 @@ All types of FileSpecs share some common fields:
 ### FileSpec -- Path
 
 A FileSpec with `type` equal to `path` points to an absolute or relative path accessible from the current machine.
+The path may also include a glob pattern to target multiple files or directories.
```
eriq-augustine, line 787: Add in the semantics around multiple files and a missing dest somewhere.

internal/common/filespec.go
```diff
@@ -208,18 +213,61 @@ func (this *FileSpec) GetPath() string {
...
+
+		// Loop over each matched path and copy it to the destination.
+		for _, path := range paths {
+			err := copyPath(path, destPath, onlyContents)
+			if err != nil {
+				return fmt.Errorf("Failed to copy target '%s': '%w'.", path, err)
+			}
+		}
+
+		return nil
```
eriq-augustine, line 270: Move to a function.
I especially don't like putting a lot of code in a conditional (doubly so for a swtich/case (which just seems like a looser if).

You can also consider merging with matchTargets(), since that got smaller.

internal/common/filespec.go
```diff
@@ -208,18 +213,61 @@ func (this *FileSpec) GetPath() string {
...
+// |baseDir| provides the relative base.
 func (this *FileSpec) CopyTarget(baseDir string, destDir string, onlyContents bool) error {
 	switch this.Type {
 	case FILESPEC_TYPE_EMPTY, FILESPEC_TYPE_NIL:
 		// no-op.
 		return nil
 	case FILESPEC_TYPE_PATH:
-		return this.copyPath(baseDir, destDir, onlyContents)
+		if !filepath.IsAbs(this.Path) && (baseDir != "") {
+			this.Path = filepath.Join(baseDir, this.Path)
```
eriq-augustine, line 229: !
Potentially dangerous thing here.
You are joining the baseDir (which can change on every call to CopyTarget()) to this.Path (which lives forever with the filespec) and putting the result in this.Path.
This can create really tough bugs where the order a FileSpec is copied changes later invocations.

internal/common/filespec.go
```diff
@@ -208,18 +213,61 @@ func (this *FileSpec) GetPath() string {
...
+		}
+
+		destPath := ""
+		if onlyContents {
+			if this.Dest == "" {
+				destPath = destDir
+			} else {
+				destPath = filepath.Join(destDir, this.Dest)
+			}
+
```
eriq-augustine, line 239: Extra newline.

internal/common/filespec.go
```diff
@@ -208,18 +213,61 @@ func (this *FileSpec) GetPath() string {
...
+		paths, err := this.matchTargets()
+		if err != nil {
+			return fmt.Errorf("Failed to match target(s) for path '%s': '%w'.", this.Path, err)
+		}
+
+		// Create a new directory if multiple targets are matched.
+		if !util.PathExists(destPath) && len(paths) > 1 {
+			err := util.MkDir(destPath)
+			if err != nil {
+				return fmt.Errorf("Failed to make a directory: '%v'.", err)
```
eriq-augustine, line 258: Give just a little more in this error string (mention "FileSpec" somewhere).

internal/common/filespec_test.go
```diff
@@ -65,11 +67,13 @@ func TestFileSpecParseValidation(test *testing.T) {
...
 type testCaseCopy struct {
-	Spec         FileSpec
-	OnlyContents bool
-	ResultMD5    string
+	Spec                  FileSpec
+	OnlyContents          bool
+	ExpectedCopiedDirents []string
 }
 
+const PATH_SEPARATOR = "dest/"
```
eriq-augustine, line 75: !!
Still has a hard-coded path separator.

eriq-augustine (changes requested):
Tests are looking pretty good.

docs/types.md
```diff
@@ -783,6 +784,9 @@ All types of FileSpecs share some common fields:
...
+For glob patterns resolving to multiple files, files will be copied into the specified `dest`.
```
eriq-augustine, line 788: This description needs work.
I think you made it sound too fancy and missed a key point about dest (that it will be treated as a dir).
You need more about dest being treated as a dir, but start simple like "For paths that match multiple dirents, ...".

internal/common/filespec.go
```diff
@@ -208,18 +213,19 @@ func (this *FileSpec) GetPath() string {
...
 // and not the base dir itself is copied.
+// |baseDir| provides the relative base.
 func (this *FileSpec) CopyTarget(baseDir string, destDir string, onlyContents bool) error {
 	switch this.Type {
 	case FILESPEC_TYPE_EMPTY, FILESPEC_TYPE_NIL:
 		// no-op.
 		return nil
 	case FILESPEC_TYPE_PATH:
-		return this.copyPath(baseDir, destDir, onlyContents)
+		return this.matchAndCopyPaths(baseDir, destDir, onlyContents)
```
eriq-augustine, line 228: I'm not a can of the "match" part of the name.
I would just call it something simple like "copyPaths".

internal/common/filespec.go
```diff
@@ -251,15 +261,44 @@ func (this *FileSpec) copyPath(baseDir string, destDir string, onlyContents bool
...
+	if err != nil {
+		return fmt.Errorf("Failed to resolve the path pattern '%s': '%w'.", this.Path, err)
+	}
+
+	if len(paths) == 0 {
+		return fmt.Errorf("No targets found for the path '%s'.", this.Path)
+	}
+
+	// Create a new directory if multiple paths are matched.
+	if !util.PathExists(destPath) && len(paths) > 1 {
```
eriq-augustine, line 274: !
You have a bug that comes up because of a deficiency in your spec.
You make dest if it doesn't exist, but what if it is a file?

eriq-augustine (changes requested):
Looking mostly good, but has a bug.

docs/types.md
```diff
@@ -783,6 +784,9 @@ All types of FileSpecs share some common fields:
 ### FileSpec -- Path
 
 A FileSpec with `type` equal to `path` points to an absolute or relative path accessible from the current machine.
+The path may also include a glob pattern to target multiple files or directories.
+For paths that match multiple dirents, the dirents will be copied into `dest`.
+If `dest` does not exist, it will be automatically created as a directory.
```
eriq-augustine, line 789: What if dest is a file?

internal/common/filespec.go
```diff
@@ -251,15 +261,47 @@ func (this *FileSpec) copyPath(baseDir string, destDir string, onlyContents bool
...
+
+	if len(paths) == 0 {
+		return fmt.Errorf("No targets found for the path '%s'.", this.Path)
+	}
+
+	if util.IsFile(destPath) && len(paths) > 1 {
+		return fmt.Errorf("Cannot copy multiple targets into the existing file '%s'.", destDir)
+	}
+
+	if !util.PathExists(destPath) && len(paths) > 1 {
```
eriq-augustine, line 277: I think it would make sense to merge these
(an outter if for having multiple paths, then error quick if it's a file).

If you did have two, you should do the len check first, since it is quicker.

eriq-augustine (changes requested):
Just some small things.

docs/types.md
```diff
@@ -783,6 +784,10 @@ All types of FileSpecs share some common fields:
 ### FileSpec -- Path
 
 A FileSpec with `type` equal to `path` points to an absolute or relative path accessible from the current machine.
+The path may also include a glob pattern to target multiple files or directories.
+For paths that match multiple dirents, the dirents will be copied into `dest` only if `dest` is a directory.
+If `dest` does not exist, it will be automatically created as a directory.
+Attempting to copy multiple dirents into `dest` if `dest` is an existing file will result in an error.
```
eriq-augustine, line 790: Too complex, make it more simple.
(Remember, KISS.)

eriq-augustine (changes requested):
KISS!

docs/types.md
```diff
@@ -783,6 +784,9 @@ All types of FileSpecs share some common fields:
 ### FileSpec -- Path
 
 A FileSpec with `type` equal to `path` points to an absolute or relative path accessible from the current machine.
+The path may include a glob pattern to target multiple files/directories.
+`dest` must be a directory if multiple files/directories are being copied (a new directory will be created if `dest` doesn't exist).
+An error occurs if multiple files/directories try to get copied when `dest` is an existing file.
```
eriq-augustine, line 789: You can just drop this line.
Saying it "must" be a dir earlier implies an error on non-dir.

docs/types.md
```diff
@@ -783,6 +784,9 @@ All types of FileSpecs share some common fields:
 ### FileSpec -- Path
 
 A FileSpec with `type` equal to `path` points to an absolute or relative path accessible from the current machine.
+The path may include a glob pattern to target multiple files/directories.
+`dest` must be a directory if multiple files/directories are being copied (a new directory will be created if `dest` doesn't exist).
```
eriq-augustine, line 788: Use a comma instead of parenthetical (so it has more authority).

internal/common/filespec.go
```diff
@@ -251,15 +261,50 @@ func (this *FileSpec) copyPath(baseDir string, destDir string, onlyContents bool
...
+		return fmt.Errorf("Failed to resolve the path pattern '%s': '%w'.", this.Path, err)
+	}
+
+	if len(paths) == 0 {
+		return fmt.Errorf("No targets found for the path '%s'.", this.Path)
+	}
+
+	// Ensure destPath is a directory if there are multiple paths.
+	if len(paths) > 1 {
+		if util.IsFile(destPath) {
```
eriq-augustine, line 275: Use `!util.IsDir()`.
There are slight differences around links on different systems.

eriq-augustine (changes requested):

PR #123, opened 2024-11-23
Added a CMD that can take in most api endpoints with their parameters

OliverLok:
This PR aims to reduce the amount of CMDs by having a CMD that takes in an endpoint with the specific endpoints parameters.


cmd/api-endpoint/expected_output_test.go
```diff
@@ -0,0 +1,47 @@
+package main
+
+const SUBMISSION_1697406272 = `{
```
eriq-augustine, line 3: !
This is duplicated in cmd/peek-submission.

(Yellow flag, this is clearly duplicate code and you wrote both.)

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+	if args.Table {
+		switch args.Endpoint {
+		case usersList:
+			printFunc = cmd.ListServerUsersTable
+		case courseUserList:
+			printFunc = cmd.ListCourseUsersTable
+		default:
+			printFunc = nil
+		}
+	}
```
eriq-augustine, line 67: I don't like this.
Having endpoint-specific code diminishes the purpose of having a general purpose CMD.

You can consider alternatives for another PR
(I can think of at least one that would work fine).

cmd/api-endpoint/main.go
eriq-augustine, line 1: I don't like the name.

CMDs (and most CLI programs) generally have some sort of verb/action.
And "endpoint" is not any sort of action, it is just a thing that you are referencing.

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+	)
+
+	err := config.HandleConfigArgs(args.ConfigArgs)
+	if err != nil {
+		log.Fatal("Failed to load config options.", err)
+	}
+
+	var description core.EndpointDescription
+
+	describe := api.Describe(*api.GetRoutes())
```
eriq-augustine, line 39: Bad name.
"describe" is a verb, variable names (especially if they are not a function) are usually nouns.

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+	for endpoint, requestResponse := range describe.Endpoints {
+		if endpoint == args.Endpoint {
+			description = requestResponse
+			break
+		}
+	}
+
+	if description == (core.EndpointDescription{}) {
```
eriq-augustine, line 47: Use a pointer instead, so you can use nil.
It makes the check more obvious.

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+		log.Error("Failed to find the endpoint.", log.NewAttr("endpoint", args.Endpoint))
```
eriq-augustine, line 48: You don't have to interact with the user through logs.
We log.Fatal() in CMDs as a quick way of exiting,
but if you are interacting with a user just printing is the standard way.
For example:
https://github.com/edulinq/autograder-server/blob/main/cmd/assignment-report/main.go
Notice we logged to exit quickly, but just printed normal information.

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+
+		if !config.UNIT_TESTING_MODE.Get() {
+			fmt.Print(util.MustToJSONIndent(api.Describe(*api.GetRoutes())))
+		}
```
eriq-augustine, line 52: I don't see a reason for this distinction.
And remember, I mentioned before how we don't want to use this flippantly.
We only want to use this if we have to, we want our code to generally run the same in tests and production.
(The closer they are the less likely we are to have strange issues we only see in production).

I also question if you need this at all.
You can just point them to the API resource file in the help output.
(You could also enumerate them all in the help output, but I don't really see a need.)

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+			printFunc = cmd.ListServerUsersTable
+		case courseUserList:
+			printFunc = cmd.ListCourseUsersTable
+		default:
+			printFunc = nil
+		}
+	}
+
+	request := map[string]any{}
+
```
eriq-augustine, line 70: Remove this newline.
We want this block to be "init and fill the request parameters".

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+	for _, arg := range args.Parameters {
+		// Split the parameter into it's key and value.
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) == 1 {
+			log.Fatal("No colon provided in the parameter.", log.NewAttr("parameter", parts))
```
eriq-augustine, line 75: Make this log more friendly, i.e., reiterate the correct format.

cmd/api-endpoint/main.go
```diff
@@ -0,0 +1,85 @@
...
+			printFunc = nil
+		}
+	}
+
+	request := map[string]any{}
+
+	for _, arg := range args.Parameters {
+		// Split the parameter into it's key and value.
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) == 1 {
```
eriq-augustine, line 74: You may as well make this more strict/safe and to `!= 2`.
Because of SplitN(), what you have is equivalent, but this should make it more clear to readers who are not familiar with SplitN().

cmd/api-endpoint/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/edulinq/autograder/internal/cmd"
+)
+
+// Use the common main for all tests in this package.
+func TestMain(suite *testing.M) {
+	cmd.CMDServerTestingMain(suite)
+}
+
+// Test to ensure an API endpoint that requires parameters works correctly.
+// Since all CMDs use the same infrastructure, only one endpoint needs to be tested.
```
eriq-augustine, line 16: This sentence is not 100% true.
You have special cases for specific endpoints in your code.
I think those special cases should go away, but I would also drop this sentence.
Testing other endpoints would be useful later to try out other edge (but not hardcoded special) cases (like if we were sending a file).

internal/cmd/request.go
```diff
@@ -107,6 +107,8 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
 
 	if customPrintFunc != nil {
 		fmt.Println(customPrintFunc(response))
+	} else if responseType == "" {
```
eriq-augustine, line 110: Why `""`?
Why not `nil`, that's the standard empty value.

internal/cmd/table.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/edulinq/autograder/internal/api/core"
+	courseUser "github.com/edulinq/autograder/internal/api/courses/users"
+	"github.com/edulinq/autograder/internal/api/users"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func ListServerUsersTable(response core.APIResponse) string {
+	var responseContent users.ListResponse
+	util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
```
eriq-augustine, line 14: !
Unmarked Must function.

I don't like having these conversions in these functions.
Back when these functions were in their respective CMDs, then these conversions make sense (we know exactly what the types should be).
But here, you are letting anyone call these with any response types
(and thus can cause a panic on the conversions).

I think you should be sending in the users instead of the responses.

eriq-augustine (changes requested):
The core general arguments is good.

cmd/call-api-endpoint/expected_output_test.go
eriq-augustine, line 1: I'm not a fan of this workaround.
If you have a good reason why you should use this submission over the most recent one, then it could maybe fly.
But without that, do the right thing and put the test data in a common file.

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/edulinq/autograder/internal/config"
+	"github.com/edulinq/autograder/internal/log"
+)
+
+var args struct {
+	config.ConfigArgs
+	cmd.CommonOptions
+
+	Endpoint   string   `help:"Endpoint of the desired API." arg:""`
+	Parameters []string `help:"Parameter(s) for the endpoint" arg:"" optional:""`
```
eriq-augustine, line 20: It's already called "parameters", so no need for the optional "s" in the description.
I think the --help puts a "...", so you should be fine (check that).

eriq-augustine, line 20: Make sure to tell people the format (and any limitations).

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func main() {
+	kong.Parse(&args,
+		kong.Description("Execute an API request to the specified endpoint. For more information on available API endpoints, see the API resource file at: ../resources/api.json"),
```
eriq-augustine, line 25: Where are the user running this from?
(Keep that in mind while looking at this line.)

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,62 @@
...
+	)
+
+	err := config.HandleConfigArgs(args.ConfigArgs)
+	if err != nil {
+		log.Fatal("Failed to load config options.", err)
+	}
+
+	var description *core.EndpointDescription
```
eriq-augustine, line 33: Since you have "apiDescription", make this variable name symmetric.

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,62 @@
...
+		if endpoint == args.Endpoint {
+			description = &requestResponse
+			break
+		}
+	}
+
+	if description == nil {
+		log.Fatal("Failed to find the endpoint.", log.NewAttr("endpoint", args.Endpoint))
+
+		// Return after log.Fatal() sets the exit code to 1 to avoid proceeding during tests.
```
eriq-augustine, line 46: This is a case where you can have a better comment with less description.
By specifying "1" specifically, you are committing to keeping this comment up-to-date in case the default exit status for a log.Fatal() changes.
That would be a super hard thing to do (searching for "1" would not be very helpful).

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,62 @@
...
+		return
+	}
+
+	request := map[string]any{}
+	for _, arg := range args.Parameters {
+		// Split the parameter into it's key and value.
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) != 2 {
+			log.Fatal("Invalid parameter format: missing a colon. Expected format is 'key:value'. E.g. target-email:bob@test.edulinq.org", log.NewAttr("parameter", parts))
```
eriq-augustine, line 55: s/. E.g./, e.g., /g
"e.g." is typically followed be a comma (and usually doesn't start sentences).

s/target-email:bob@test.edulinq.org/'target-email:bob@test.edulinq.org'
Since you quote the template ('key:value'), quote the example.

s/$/./
Try for complete sentences.

eriq-augustine, line 55: You should have a test for this case (and then you will need a return).

cmd/call-api-endpoint/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/edulinq/autograder/internal/cmd"
+)
+
+// Use the common main for all tests in this package.
+func TestMain(suite *testing.M) {
+	cmd.CMDServerTestingMain(suite)
+}
+
+// Test to ensure an API endpoint that requires parameters works correctly.
```
eriq-augustine, line 15: I don't like this comment.
It's too conclusive (these tests are not conclusive).
I would just drop it, since this test is not tricky or special.

cmd/call-api-endpoint/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+func TestCallApiEndpointBase(test *testing.T) {
+	testCases := []struct {
+		cmd.CommonCMDTestCase
+		endpoint     string
+		targetEmail  string
+		courseID     string
+		assignmentID string
+		submissionID string
```
eriq-augustine, line 23: I really don't like this.
The whole point of the CMD is to handle general endpoints, but here you are limiting test cases to only these params (just like you previously had the CMD CLI arguments).

eriq-augustine (changes requested):
Getting closer.

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/edulinq/autograder/internal/config"
+	"github.com/edulinq/autograder/internal/log"
+)
+
+var args struct {
+	config.ConfigArgs
+	cmd.CommonOptions
+
+	Endpoint   string   `help:"Endpoint of the desired API." arg:""`
+	Parameters []string `help:"Parameter for the endpoint in the format 'key:value', where the first colon separates the key and value (e.g., 'target-email:bob@test.edulinq.org'). Additional colons are allowed in the value." arg:"" optional:""`
```
eriq-augustine, line 20: Make this text more terse (help prompts and man pages are typically pretty terse).
Everything past "'key:value'" seems kind of useless.
I could see giving a small example, but use shorter strings for it.

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func main() {
+	kong.Parse(&args,
+		kong.Description("Execute an API request to the specified endpoint. For more information on available API endpoints, see the API resource file at: /autograder-server/resources/api.json"),
```
eriq-augustine, line 25: The path is still off here.
Use an absolute path (and assuming the repo is at root) is no good.
Just use `resources/api.json` and assume you are at the repo's root (not filesystem root).

eriq-augustine (changes requested):
This is looking real good.
There is just one main thing left to resolve around testing.
It would be great to have more test cases.
I'm thinking three classes of test cases:
 - Errors and tricky failures.
 - Special functionality (like custom printers in the future).
 - Base functionality for each supported endpoint.

It's this last one that is tricky.
We don't need to do exhaustive tests (like the API tests),
but we want to make sure it runs (which is more similar to the tests we had before with these API CMDs).
But the real tricky part is where we put all the test data.
I guess it could just all live in this package (in the expected output file).
It would be big, but not too bad since it's just definitions.
I'm not sure, think about it a bit.

cmd/call-api-endpoint/main_test.go
```diff
@@ -0,0 +1,116 @@
...
+				ExpectedStderrSubstring: `Failed to find the endpoint. | {"endpoint":"ZZZ"}`,
+				ExpectedExitCode:        1,
+				ExpectEmptyStdout:       true,
+			},
+			endpoint: "ZZZ",
+			parameters: []string{
+				"target-email:course-student@test.edulinq.org",
+				"course-id:course101",
+				"assignment-id:hw0",
+				"target-submission:1697406256",
```
eriq-augustine, line 33: Since the endpoint doesn't exist, you don't need to pass any args.

cmd/call-api-endpoint/main.go
```diff
@@ -0,0 +1,65 @@
...
+		// Return to prevent further execution after log.Fatal().
+		return
+	}
+
+	request := map[string]any{}
+	for _, arg := range args.Parameters {
+		// Split the parameter into it's key and value.
+		parts := strings.SplitN(arg, ":", 2)
+		if len(parts) != 2 {
+			log.Fatal("Invalid parameter format: missing a colon. Expected format is 'key:value', e.g., 'target-email:bob@test.edulinq.org'.", log.NewAttr("parameter", parts))
```
eriq-augustine, line 55: Use the same example as the help.

cmd/call-api-endpoint/main_test.go
```diff
@@ -0,0 +1,116 @@
...
+			endpoint: "courses/assignments/submissions/fetch/user/peek",
+			parameters: []string{
+				"target-email:course-student@test.edulinq.org",
+				"course-idcourse101",
+				"assignment-id:hw0",
+				"target-submission:1697406256",
+			},
+		},
+
+		// Base functionality for each supported endpoint.
```
eriq-augustine, line 51: I feel like there are a good number of endpoints missing here.
Ones like a lot of the submission fetches and history.

internal/cmd/test.go
```diff
@@ -179,8 +180,13 @@ func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonTe
 		return "", "", -1, false
 	}
 
-	if commonTestCase.ExpectedStdout != stdout {
-		test.Errorf("%sUnexpected output. Expected: \n'%s', \n Actual: \n'%s'.", prefix, commonTestCase.ExpectedStdout, stdout)
+	if commonTestCase.ExpectEmptyStdout && len(stdout) > 0 {
```
eriq-augustine, line 183: !
You don't have a case for comparing stdout!

internal/cmd/test.go
```diff
@@ -27,6 +27,7 @@ const (
 type CommonCMDTestCase struct {
 	ExpectedExitCode        int
 	ExpectedStdout          string
+	ExpectEmptyStdout       bool
```
eriq-augustine, line 30: I don't think the semantics here work well.

You had to edit a good number of test cases and now anyone who wants to write a failing case (where the CMD fails and usually fatals) has to add this param
(which I don't think they would expect).

What you are really looking for is a way to skip the stdout check.
I don't think that behavior is intuitively achieved with this current option set.
You may just want some `IgnoreStdout` option instead.

eriq-augustine (changes requested):
The code code is looking good.

I think the test semantics are off and we'll need some more.

internal/cmd/test.go
```diff
@@ -179,7 +180,7 @@ func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonTe
 		return "", "", -1, false
 	}
 
-	if commonTestCase.ExpectedStdout != stdout {
+	if commonTestCase.ExpectedStdout != stdout && !commonTestCase.IgnoreStdout {
```
eriq-augustine, line 183: Swap this ordering.
When you are short-circuiting and one check (e.g., an unknown-length string comparison) can be long, do short checks (e.g., boolean checks first).

eriq-augustine (changes requested):
Just one small thing.
It doesn't effect functionality, but is a good thing to know about.

eriq-augustine (approved):
lgtm

PR #129, opened 2024-12-05
Added custom output support for the general API CMD.

OliverLok:
This uses a general formatting function that tries to convert a response content's output into a table. It defaults back to JSON if it can't convert it.


cmd/call-endpoint/main.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/edulinq/autograder/internal/util"
+)
+
+var args struct {
+	config.ConfigArgs
+	cmd.CommonOptions
+
+	Endpoint   string   `help:"Endpoint of the desired API." arg:""`
+	Parameters []string `help:"Parameter for the endpoint in the format 'key:value', e.g., 'id:123'." arg:"" optional:""`
+	Table      bool     `help:"Attempt to output data as a TSV. Will default to JSON." default:"false"`
```
eriq-augustine, line 25: The second sentence ("Will default to JSON.") needs some work.
The parameter itself defaults to `false`, so your wording could be a bit confusing.
Since you use the word "attempt" in the previous sentence, you want this one to tell what happens when that attempt can't happen.

cmd/call-endpoint/main.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+const (
+	USERS_KEY   = "users"
+	COURSES_KEY = "courses"
+)
```
eriq-augustine, line 31: Nah, you can't do this.
This is not too different than hard-coding the endpoints.
You have to inspect the response to see if you can tabularize it.

cmd/call-endpoint/main.go
```diff
@@ -0,0 +1,172 @@
...
+			log.Fatal("Invalid parameter format: missing a colon. Expected format is 'key:value', e.g., 'id:123'.", log.NewAttr("parameter", parts))
+
+			// Return to prevent further execution after log.Fatal().
+			return
+		}
+
+		request[parts[0]] = parts[1]
+	}
+
+	var printFunc cmd.CustomResponseFormatter
```
eriq-augustine, line 77: I prefer things like this to be more clear with a `= nil` at the end.

cmd/call-endpoint/main.go
```diff
@@ -0,0 +1,172 @@
...
+
+	apiDescription := api.Describe(*api.GetRoutes())
+	for endpoint := range apiDescription.Endpoints {
+		endpointList.WriteString(fmt.Sprintf("  - %s\n", endpoint))
+	}
+
+	return baseDescription + endpointList.String()
+}
+
+func printCMDResponseTable(response core.APIResponse) string {
```
eriq-augustine, line 99: Put this function in internal/cmd.

cmd/call-endpoint/main.go
```diff
@@ -0,0 +1,172 @@
...
+
+	var printFunc cmd.CustomResponseFormatter
+	if args.Table {
+		printFunc = printCMDResponseTable
+	}
+
+	cmd.MustHandleCMDRequestAndExitFull(args.Endpoint, request, nil, args.CommonOptions, printFunc)
+}
+
+func generateHelpDescription() string {
```
eriq-augustine, line 85: This is not in the scope of the this PR.

cmd/call-endpoint/main.go
```diff
@@ -0,0 +1,172 @@
...
+
+		var row []string
+		for _, key := range headerKeys {
+			switch value := userMap[key].(type) {
+			case string:
+				row = append(row, value)
+			case int:
+				row = append(row, strconv.Itoa(value))
+			case bool:
+				row = append(row, strconv.FormatBool(value))
```
eriq-augustine, line 158: I'm not a fan of these cases.
What about float?
Why do the non-JSON cases have to be treated differently?

internal/cmd/request.go
```diff
@@ -105,8 +105,13 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
 		fmt.Printf("\nAutograder Response:\n---\n%s\n---\n", util.MustToJSONIndent(response))
 	}
 
+	customOutput := ""
 	if customPrintFunc != nil {
-		fmt.Println(customPrintFunc(response))
+		customOutput = customPrintFunc(response)
```
eriq-augustine, line 110: What if empty string is valid output?
Usually better to return an error with this.

eriq-augustine (changes requested):
The core plan here needs some work.
Remember, we are trying to get away from hard-codes and special cases when possible.

cmd/call-endpoint/main.go
```diff
@@ -18,6 +18,7 @@ var args struct {
 
 	Endpoint   string   `help:"Endpoint of the desired API." arg:""`
 	Parameters []string `help:"Parameter for the endpoint in the format 'key:value', e.g., 'id:123'." arg:"" optional:""`
+	Table      bool     `help:"Attempt to output data as a TSV; defaults to JSON if there are issues." default:"false"`
```
eriq-augustine, line 21: Nope.
The word "default" is incorrect here.
"Default" in programming means an initial value used when no value is specified.
You are using "default" all "fallback".

eriq-augustine, line 21: I believe the semicolon is grammatically incorrect here.

cmd/course-user-list/main.go
```diff
@@ -45,7 +45,7 @@ func main() {
 	cmd.MustHandleCMDRequestAndExitFull(`courses/users/list`, request, users.ListResponse{}, args.CommonOptions, printFunc)
 }
 
-func listCourseUsersTable(response core.APIResponse) string {
+func listCourseUsersTable(response core.APIResponse) (string, bool) {
```
eriq-augustine, line 48: Why the change?
No one is using it.

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
 		fmt.Println(util.MustToJSONIndent(responseContent))
 	}
 }
+
+// Attempt to convert an API response to a TSV table.
+// Returns false if there are issues or the TSV table if it was successful.
+func PrintCMDResponseTable(response core.APIResponse) (string, bool) {
```
eriq-augustine, line 129: Name is misleading.

(I'm purposefully not telling you why the name is deceptive, so you have a chance to think about it.)

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
 		fmt.Println(util.MustToJSONIndent(responseContent))
 	}
 }
+
+// Attempt to convert an API response to a TSV table.
+// Returns false if there are issues or the TSV table if it was successful.
```
eriq-augustine, line 128: This statement is not correct.
You are returning two things, so this line cannot be correct.

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+	if !ok {
+		return "", false
+	}
+
+	// Don't try to format a response that has multiple keys.
+	if len(responseContent) != 1 {
+		return "", false
+	}
+
+	responseContentKey := reflect.ValueOf(responseContent).MapKeys()
```
eriq-augustine, line 140: Why not just range over the values (which you know there is only 1).

Reflection is super cool, but we should only do it when we need it.

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+
+	// Get the rows that will be added to the table.
+	entries, ok := responseContent[responseContentKey[0].String()].([]any)
+	if !ok {
+		return "", false
+	}
+
+	// Use the first entry to create the headers of the table.
+	firstEntry, ok := entries[0].(map[string]any)
```
eriq-augustine, line 149: Panic.

Add a test case and see the panic before you fix it.

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+	for key := range firstEntry {
+		headers = append(headers, key)
+	}
+
+	sort.Strings(headers)
+
+	var customTable strings.Builder
+	customTable.WriteString(strings.Join(headers, "\t"))
+
+	headerKeys := strings.Split(customTable.String(), "\t")
```
eriq-augustine, line 164: !
wtf? ("What the funk?")

You join and then re-split right away?
Are these values not all already in `headers`?

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+
+	customTable.WriteString("\n")
+
+	// Turn each entry into a row of the table.
+	for i, entry := range entries {
+		entryMap, ok := entry.(map[string]any)
```
eriq-augustine, line 170: If you are just going to do this check here, could you not do it on the parent on line 143?

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+	}
+
+	sort.Strings(headers)
+
+	var customTable strings.Builder
+	customTable.WriteString(strings.Join(headers, "\t"))
+
+	headerKeys := strings.Split(customTable.String(), "\t")
+
+	customTable.WriteString("\n")
```
eriq-augustine, line 166: In this case, you can just always just prefix this to the row (not the header).

internal/cmd/request.go
```diff
@@ -115,3 +123,71 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+			case map[string]any:
+				row = append(row, util.MustToJSON(value))
+			default:
+				row = append(row, fmt.Sprintf("%v", value))
+			}
+		}
+
+		customTable.WriteString(strings.Join(row, "\t"))
+
+		if i < len(entries)-1 {
```
eriq-augustine, line 187: I don't like the no space around binary operators (here, `-`) and having a non-leaf expression in a binary comparison (here, `<`) without parens.

By doing the latter, gofmt will let you do the former.

eriq-augustine (changes requested):
Looking better.

internal/cmd/request.go
```diff
@@ -105,8 +107,14 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
 	}
 
+	successfulConversion := false
+	customPrintOutput := ""
 	if customPrintFunc != nil {
-		fmt.Println(customPrintFunc(response))
+		customPrintOutput, successfulConversion = customPrintFunc(response)
+	}
+
+	if successfulConversion && customPrintFunc != nil {
```
eriq-augustine, line 116: Just print once after you collect the output from these conditionals.

internal/cmd/request.go
```diff
@@ -115,3 +123,75 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
 		fmt.Println(util.MustToJSONIndent(responseContent))
 	}
 }
+
+// Attempt to convert an API response to a TSV table.
+// Return ("", false) if there are issues converting the API response
+// or (customTable.String(), true) if the response got sucessfully converted.
+func AttemptApiResponseToTable(response core.APIResponse) (string, bool) {
```
eriq-augustine, line 130: Name still needs work.
Every function that returns an error (or success status) "attempts" to do something,
so we don't need the prefix here.

internal/cmd/request.go
eriq-augustine: You missed the important part of this comment.
I did not say "append".

eriq-augustine (changes requested):

eriq-augustine (changes requested):

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+	}
+
+	fmt.Println(output)
+}
+
+// Attempt to convert an API response into a TSV table.
+// If the response content has multiple keys or a single map, it creates a single row table.
+// If it has one key with a list, the values in each element becomes a row of the table.
+// Return the table and true on a successful conversion or false if there is an issue converting.
+func ConvertApiResponseToTable(response core.APIResponse) (string, bool) {
```
eriq-augustine, line 136: s/Api/API/

In this project we usually do all caps for acronyms.
This is a super project-by-project (and sometimes file-by-file) thing.
I didn't even know what we did and just grepped for 'Api' vs 'API'.

cmd/call-endpoint/main_test.go
```diff
@@ -40,6 +41,56 @@ func TestCallApiEndpointBase(test *testing.T) {
...
+			CommonCMDTestCase: cmd.CommonCMDTestCase{
+				ExpectedStdout: EXPECTED_COURSES_ASSIGNMENTS_TABLE,
+			},
+			endpoint: "courses/assignments/list",
+			parameters: []string{
+				"course-id:course101",
+				"--table",
+			},
+		},
+		// Single key with json.
```
eriq-augustine, line 56: Bad comment.
s/json/map// ?

cmd/call-endpoint/main_test.go
```diff
@@ -40,6 +41,56 @@ func TestCallApiEndpointBase(test *testing.T) {
...
+		{
+			CommonCMDTestCase: cmd.CommonCMDTestCase{
+				ExpectedStdout: EXPECTED_COURSES_ASSIGNMENTS_TABLE,
```
eriq-augustine, line 59: Same output?

Assuming they have the same output,
you should either generalize the variable name or alias it by creating a new variable that is initialized with the other variable.

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+	customPrintOutput := ""
 	if customPrintFunc != nil {
-		fmt.Println(customPrintFunc(response))
-	} else if responseType == nil {
-		fmt.Println(util.MustToJSONIndent(response.Content))
+		customPrintOutput, successfulConversion = customPrintFunc(response)
+	}
+
+	output := ""
+	if successfulConversion {
```
eriq-augustine, line 117: If you are going to use a nested if, then I don't think you need this case.
Just set `output` on the custom print call and only enter this conditional if the on `!successfulConversion`.

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+			output = util.MustToJSONIndent(response.Content)
+		}
+	}
+
+	fmt.Println(output)
+}
+
+// Attempt to convert an API response into a TSV table.
+// If the response content has multiple keys or a single map, it creates a single row table.
+// If it has one key with a list, the values in each element becomes a row of the table.
```
eriq-augustine, line 134: Put the list comment first, since it is more obvious.

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+	responseContent, ok := response.Content.(map[string]any)
+	if !ok {
+		return "", false
+	}
+
+	if len(responseContent) == 0 {
+		return "", false
+	} else if len(responseContent) > 1 {
+		return convertEntriesToTable([]any{responseContent})
+	} else {
```
eriq-augustine, line 146: I would break these up and remove the need for the last else.

I know this may sound confusing, given on how the last pass I told you that mutually exclusive cases should be in the same conditional tree.
But, the difference here is that you handle the zero case and fail fast.
Here is the comment for reference:

> These are mutually exclusive, they should be in the same tree.
> (If they both needed to exist.)
> 
> But, they probably don't all need to exist.
> Fail faster with zero length, then you don't need to have the second conditional.

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+// If the response content has multiple keys or a single map, it creates a single row table.
+// If it has one key with a list, the values in each element becomes a row of the table.
+// Return the table and true on a successful conversion or false if there is an issue converting.
+func ConvertApiResponseToTable(response core.APIResponse) (string, bool) {
+	responseContent, ok := response.Content.(map[string]any)
+	if !ok {
+		return "", false
+	}
+
+	if len(responseContent) == 0 {
```
eriq-augustine, line 142: Put a small comment on each case letting the reader know which it is.

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+		return "", false
+	} else if len(responseContent) > 1 {
+		return convertEntriesToTable([]any{responseContent})
+	} else {
+		var responseContentKey string
+		for key := range responseContent {
```
eriq-augustine, line 148: Note here how there is only one key/value to get.

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+			entries = value
+		default:
+			return "", false
+		}
+
+		return convertEntriesToTable(entries)
+	}
+}
+
+// Attempt to convert a slice of entries into a TSV table.
```
eriq-augustine, line 166: "slice of entries" is a bit ambiguous.
You can just give the type `[]map[string]any` since this is an unexported function.

internal/cmd/request.go
```diff
@@ -105,13 +107,107 @@ func PrintCMDResponseFull(request any, response core.APIResponse, responseType a
...
+			return "", false
+		}
+
+		var row []string
+		for _, key := range headers {
+			switch value := entryMap[key].(type) {
+			case map[string]any:
+				row = append(row, util.MustToJSON(value))
+			case []any:
+				row = append(row, util.MustToJSON(value))
```
eriq-augustine, line 203: I think you can double-up these cases:
https://go.dev/wiki/Switch#multiple-cases

scripts/update_api_description.sh
```diff
@@ -17,7 +17,7 @@ function main() {
 
     mkdir -p "${OUT_DIR}"
 
-    go run cmd/describe-api/main.go > "${OUT_FILE}"
+    go run cmd/call-endpoint/main.go metadata/describe > "${OUT_FILE}"
```
eriq-augustine, line 20: Just and FYI.
Hopefully github alerted you to this comment:
https://github.com/edulinq/autograder-server/pull/131#discussion_r1876974800

eriq-augustine (changes requested):
Nice.
The core here is looking pretty good, just minor things.

PR #130, opened 2024-12-06
Added a flag to the call-endpoint CMD to list all endpoints.

OliverLok:
This makes it easier to see a list of endpoints rather than having to go to an external file.


eriq-augustine (changes requested):
I am a bit hesitant about this one:
1) I am not convinced the the output will be clean enough to not just be a distraction. Copy example output (text, not image) into the PR conversation so we can see what it looks like when all finished.
2) What about endpoints we don't yet support? It feels a little disingenuous to list it and then smack down the user when they try and use it.

cmd/call-endpoint/main.go
```diff
@@ -16,14 +17,15 @@ var args struct {
 	config.ConfigArgs
 	cmd.CommonOptions
 
-	Endpoint   string   `help:"Endpoint of the desired API." arg:""`
+	Endpoint   string   `help:"Endpoint of the desired API." arg:"" optional:""`
```
eriq-augustine, line 20: Now that this is optional, there should be some sort of special case for handling when it is not specified.
Think about how you want to user to experience it.

cmd/call-endpoint/main.go
```diff
@@ -76,3 +84,14 @@ func main() {
...
 	cmd.MustHandleCMDRequestAndExitFull(args.Endpoint, request, nil, args.CommonOptions, printFunc)
 }
+
+func listAPIEndpoints() {
+	apiDescription, err := api.Describe(*api.GetRoutes())
+	if err != nil {
+		log.Fatal("Failed to describe API endpoints: '%w'.", err)
+	}
+
+	for endpoint := range apiDescription.Endpoints {
```
eriq-augustine, line 94: I don't like one-valued map range (I don't think it is as clear as using an empty value).
Put in an underscore for the value.

eriq-augustine (changes requested):
As I mentioned in my last round, past the output of a run in the comments so I can see what it looks like.

OliverLok:
> As I mentioned in my last round, past the output of a run in the comments so I can see what it looks like.

Output from using --list:
courses/users/get
courses/assignments/submissions/remove
courses/assignments/submissions/analysis/pairwise
courses/users/enroll
courses/assignments/submissions/fetch/user/attempt
courses/users/drop
system/stacks
users/password/change
courses/stats/query
courses/upsert/filespec
metadata/describe
courses/upsert/zip
lms/user/get
courses/admin/update
courses/assignments/list
courses/lms/scores/upload
users/auth
users/password/reset
courses/assignments/submissions/fetch/user/attempts
courses/assignments/submissions/fetch/user/history
courses/assignments/submissions/fetch/user/peek
users/get
users/list
users/upsert
courses/assignments/get
courses/assignments/submissions/submit
stats/system/query
users/tokens/create
courses/assignments/submissions/analysis/individual
lms/upload/scores
users/remove
logs/query
users/tokens/delete
courses/assignments/submissions/fetch/course/attempts
courses/assignments/submissions/fetch/course/scores
courses/users/list

cmd/call-endpoint/main.go
```diff
@@ -16,14 +17,15 @@ var args struct {
...
+	Endpoint   string   `help:"Endpoint of the desired API." arg:"" optional:""`
 	Parameters []string `help:"Parameter for the endpoint in the format 'key:value', e.g., 'id:123'." arg:"" optional:""`
 	Table      bool     `help:"Attempt to output data as a TSV. Fallback to JSON if the table conversion fails." default:"false"`
+	List       bool     `help:"List all API endpoints." default:"false"`
 }
 
 func main() {
 	kong.Parse(&args,
-		kong.Description("Execute an API request to the specified endpoint. For more information on available API endpoints, see the API resource file at: 'resources/api.json'."),
+		kong.Description("Execute an API request to a specified endpoint."),
```
eriq-augustine, line 28: I'm not a fan of this description.
"Executing" "to" something sounds strange.
"Sending" a request "to" and endpoint makes sense.
Or maybe use "call" to be symmetric with the CMD's name.

cmd/call-endpoint/main.go
```diff
@@ -34,13 +36,26 @@ func main() {
 		return
 	}
 
-	var endpointDescription *core.EndpointDescription
+	if args.List {
+		listAPIEndpoints()
+
+		// Return to prevent further execution after listing endpoints.
```
eriq-augustine, line 42: I don't think the comment (or space) are necessary here.
What you are doing is pretty clear.

cmd/call-endpoint/main.go
```diff
@@ -76,3 +91,14 @@ func main() {
 
 	cmd.MustHandleCMDRequestAndExitFull(args.Endpoint, request, nil, args.CommonOptions, printFunc)
 }
+
+func listAPIEndpoints() {
```
eriq-augustine, line 95: This should sort the endpoints.

cmd/call-endpoint/main_test.go
```diff
@@ -22,7 +22,7 @@ func TestCallEndpoint(test *testing.T) {
 		// Errors.
 		{
 			CommonCMDTestCase: cmd.CommonCMDTestCase{
-				ExpectedStderrSubstring: `Failed to find the endpoint. | {"endpoint":"ZZZ"}`,
+				ExpectedStderrSubstring: `Failed to find the endpoint. See --list to view all endpoints. | {"endpoint":"ZZZ"}`,
```
eriq-augustine, line 25: s/See/Use/

"See" really only makes sense if your are looking up or referencing something.
Here you are invoking the executable again with a new flag.

eriq-augustine (changes requested):
Now that I see the output, this generally makes sense to me.

OliverLok:
> As I mentioned in my last round, past the output of a run in the comments so I can see what it looks like.

Sorted list of endpoints:
courses/admin/update
courses/assignments/get
courses/assignments/list
courses/assignments/submissions/analysis/individual
courses/assignments/submissions/analysis/pairwise
courses/assignments/submissions/fetch/course/attempts
courses/assignments/submissions/fetch/course/scores
courses/assignments/submissions/fetch/user/attempt
courses/assignments/submissions/fetch/user/attempts
courses/assignments/submissions/fetch/user/history
courses/assignments/submissions/fetch/user/peek
courses/assignments/submissions/remove
courses/assignments/submissions/submit
courses/lms/scores/upload
courses/stats/query
courses/upsert/filespec
courses/upsert/zip
courses/users/drop
courses/users/enroll
courses/users/get
courses/users/list
lms/upload/scores
lms/user/get
logs/query
metadata/describe
stats/system/query
system/stacks
users/auth
users/get
users/list
users/password/change
users/password/reset
users/remove
users/tokens/create
users/tokens/delete
users/upsert

cmd/call-endpoint/main.go
```diff
@@ -16,14 +18,15 @@ var args struct {
...
+	Endpoint   string   `help:"Endpoint of the desired API." arg:"" optional:""`
 	Parameters []string `help:"Parameter for the endpoint in the format 'key:value', e.g., 'id:123'." arg:"" optional:""`
 	Table      bool     `help:"Attempt to output data as a TSV. Fallback to JSON if the table conversion fails." default:"false"`
+	List       bool     `help:"List all API endpoints." default:"false"`
 }
 
 func main() {
 	kong.Parse(&args,
-		kong.Description("Execute an API request to the specified endpoint. For more information on available API endpoints, see the API resource file at: 'resources/api.json'."),
+		kong.Description("Call an endpoint to perform an API request."),
```
eriq-augustine, line 29: This sill sounds a bit overly complicated.
Specifically, the last half sounds redundant (if you are calling an API endpoint you must be making a request).

cmd/call-endpoint/main.go
```diff
@@ -34,13 +37,24 @@ func main() {
...
-	var endpointDescription *core.EndpointDescription
+	if args.List {
+		listAPIEndpoints()
+		return
+	}
+
+	if args.Endpoint == "" {
+		log.Error("Please enter an endpoint. Use --list to view all endpoints.")
+
+		// Return to prevent further execution after log.Error().
```
eriq-augustine, line 48: Last round there was a comment that removed this same substring (these two lines) from the case above.
Remember that style comments generally apply to the entire PR.

cmd/call-endpoint/main.go
```diff
@@ -76,3 +90,21 @@ func main() {
...
+	if err != nil {
+		log.Fatal("Failed to describe API endpoints: '%w'.", err)
+	}
+
+	var endpoints []string
+	for endpoint, _ := range apiDescription.Endpoints {
+		endpoints = append(endpoints, endpoint)
+	}
+
+	sort.Strings(endpoints)
```
eriq-augustine, line 105: I would use slices.Sort() instead.
It's not a big deal, but modern Go (since generics) just calls the slices version.
https://pkg.go.dev/sort#Strings

eriq-augustine (changes requested):
Nice, just a few things.

eriq-augustine (approved):
lgtm!

PR #131, opened 2024-12-09
Added endpoint descriptions to the description of all API Endpoints.

Lucas-Ellenberger:
This PR automatically unpacks the description for API endpoints by finding the comment attached to the endpoint's handler.

The PR adds a strict directory convention to our API endpoints. A full description of the convention can be found in `docs/development.md`.

If a description cannot be found for an endpoint, a warning is logged. This approach was chosen over exiting with an error as some tests, such as `TestGoodPostFiles` create a new API Endpoint, but cannot follow the proposed endpoint structure.


docs/development.md
```diff
@@ -92,6 +92,22 @@ Within this range, each package in the API package is given a range of 100 locat
 
 ## API Notes
 
+### API File Structure
+
+All API endpoints must reside in the corresponding file path from `internal/api`.
```
eriq-augustine, line 97: This sentence could use some clarifying.
"corresponding" is doing too much work here.
I know you have an example below, but examples are meant to clarify not explain.

docs/development.md
```diff
@@ -92,6 +92,22 @@ Within this range, each package in the API package is given a range of 100 locat
...
+Additionally, the handler for the endpoint must be in a file with the same name.
+For example, the handler for `users/list` must be in a file named `list.go`.
+
+The handler function must contain the substring `Handle`.
+For our `users/list` example, a good handler name would be `HandleList`.
+
+Putting it all together for our `users/list` example, there should be a handler with `Handle` in it's name
+in the file `list.go`, which can be located in the directory `internal/api/users`.
+
+By following these conventions, we can properly find information about every endpoint.
```
eriq-augustine, line 109: There paraggraphs seem a little sparse (2 sentences max).
I think there can be some consolidation.
(Not that I like long paragraphs, but these seem a bit too short.)

internal/api/core/route.go
```diff
@@ -106,5 +111,28 @@ func MustNewAPIRoute(basePath string, apiHandler any) *APIRoute {
...
+		function, ok := decl.(*ast.FuncDecl)
+		if !ok {
+			continue
+		}
+
+		if strings.Contains(function.Name.Name, "Handle") && function.Doc != nil {
+			apiRoute.Description = strings.TrimSpace(function.Doc.Text())
+			break
+		}
+	}
```
eriq-augustine, line 135: Move this to another file.

The amount of code is small,
but we don't want readers to even have to think about parsing the source files if they don't need to.

internal/api/core/routes.go
```diff
@@ -17,3 +18,11 @@ func MakeFullAPIPath(suffix string) string {
 
 	return CURRENT_PREFIX + "/" + suffix
 }
+
+func makeAbsLocalAPIPath(suffix string) string {
```
eriq-augustine, line 22: I already mentioned in an earlier comment moving this to another file.
But you should also note that it seems strange that there is only one file that uses this and the function is not located in that file.

internal/api/core/route.go
```diff
@@ -106,5 +111,28 @@ func MustNewAPIRoute(basePath string, apiHandler any) *APIRoute {
...
 		RequestType:  requestType,
 		ResponseType: responseType,
+		Description:  "",
 	}
+
+	absPath := makeAbsLocalAPIPath(basePath) + ".go"
+	fset := token.NewFileSet()
+	node, parseErr := parser.ParseFile(fset, absPath, nil, parser.ParseComments)
+	if parseErr != nil {
+		log.Warn("Error while parsing file to get API description.", parseErr, log.NewAttr("endpoint", absPath))
```
eriq-augustine, line 121: Should this be an error?
Under what conditions can we fail a parse (given that we are running a binary that should have the compiled code in it)?

scripts/update_api_description.sh
```diff
@@ -17,7 +17,7 @@ function main() {
 
     mkdir -p "${OUT_DIR}"
 
-    go run cmd/describe-api/main.go > "${OUT_FILE}"
+    go run cmd/call-api-endpoint/main.go metadata/describe > "${OUT_FILE}"
```
eriq-augustine, line 20: Just an FYI, this will change soon.

It's a (non-competitive) race between @OliverLok (#129) and this for who will need to change this line.

resources/api.json
eriq-augustine, line 1: Add to the describe test to verify that all endpoints have a description.

internal/api/core/route.go
```diff
@@ -106,5 +111,28 @@ func MustNewAPIRoute(basePath string, apiHandler any) *APIRoute {
...
+		return &apiRoute
+	}
+
+	for _, decl := range node.Decls {
+		function, ok := decl.(*ast.FuncDecl)
+		if !ok {
+			continue
+		}
+
+		if strings.Contains(function.Name.Name, "Handle") && function.Doc != nil {
```
eriq-augustine, line 131: Could an anonymous function appear in the top-level declarations?
If so, would that have a nil name?

The `Name.Name` makes me a bit nervous about a panic.

internal/api/core/route.go
```diff
@@ -106,5 +111,28 @@ func MustNewAPIRoute(basePath string, apiHandler any) *APIRoute {
...
+			apiRoute.Description = strings.TrimSpace(function.Doc.Text())
```
eriq-augustine, line 132: Doc can be nil:
https://pkg.go.dev/go/ast#FuncDecl

eriq-augustine (changes requested):
Nice!
Grabbing the description comment was nice and clean.

I don't think your initial description of this PR is correct:
> If a description cannot be found for an endpoint, a warning is logged.

I see a warning if the parse fails, but not on an empty description.
In a comment, I recommended adding a check for non-empty desripiton in a test.

More about your initial PR comment confuses me:
> This approach was chosen over exiting with an error as some tests, such as TestGoodPostFiles create a new API Endpoint, but cannot follow the proposed endpoint structure.

Why are those test API endpoints ever involved?
Shouldn't they only exist for the scope of a single test?
(If not, they really should.)

Lucas-Ellenberger:
> Nice! Grabbing the description comment was nice and clean.
Thanks!
> > This approach was chosen over exiting with an error as some tests, such as TestGoodPostFiles create a new API Endpoint, but cannot follow the proposed endpoint structure.
> 
> Why are those test API endpoints ever involved? Shouldn't they only exist for the scope of a single test? (If not, they really should.)

Tests create API endpoints and they only exist in the scope of a single test. But, they add the API endpoint by calling `MustNewAPIRoute`, which is were the magic of unpacking the description from the endpoints path happens.

I debated moving the describe logic out of `MustNewAPIRoute` and building the API description later, perhaps inside a function like `Describe()`. My issue with putting it into another function with the potential to error seems sketchier. The current setup is only called when starting the server and can be verified to work through a single test. I could rename `Describe()` to `MustDescribe()` to ensure other developers know to be wary.

I'll resolve these issues and see if I can find a clean way of organizing everything.

docs/development.md
```diff
@@ -92,6 +92,19 @@ Within this range, each package in the API package is given a range of 100 locat
...
 ## API Notes
 
+### API File Structure
+
+All API endpoints must reside in a directory that follows the same structure as the endpoint's base path from `internal/api`.
+For example, the endpoint `users/list` must reside in a directory called `internal/api/users`.
+Additionally, the handler for the endpoint must be in a file with the same name.
+For example, the handler for `users/list` must be in a file named `list.go`.
+Finally, the handler function must contain the substring `Handle`.
+For our `users/list` example, a good handler name would be `HandleList`.
```
eriq-augustine, line 102: This paragraph is a bit "stuttery",
in that it gives an example and then lists another requirement,
and then repeats this.

Instead, try having the description first then then the example.
For example:
> The structure of API endpoints files must follow these rules ...
> \1) ...
> ...
> For example, the handler for `users/list` is called ... and located at ...

(Ignore the backslash above.)

internal/api/core/describe.go
```diff
@@ -19,3 +31,45 @@ func SetAPIDescription(description APIDescription) {
 func GetAPIDescription() APIDescription {
 	return apiDescription
 }
+
+func GetDescriptionFromHandler(basePath string) (string, error) {
```
eriq-augustine, line 35: Move the bulk of this low-level code to another file (more-or-less by itself) so that other people don't have to worry about AST (which are pretty cool).

internal/api/core/describe.go
```diff
@@ -19,3 +31,45 @@ func SetAPIDescription(description APIDescription) {
...
+
+		if strings.Contains(function.Name.Name, "Handle") {
+			return strings.TrimSpace(function.Doc.Text()), nil
+		}
+	}
+
+	return "", fmt.Errorf("Unable to find a description for an endpoint: '%s'.", basePath)
+}
+
+func makeAbsLocalAPIPath(suffix string) string {
```
eriq-augustine, line 69: Maybe instead call this something like "getHandlerFilePath()" and the parameter could be the base path.
Then you can add on the ".go" here.

internal/api/core/describe.go
```diff
@@ -19,3 +31,45 @@ func SetAPIDescription(description APIDescription) {
...
+			continue
+		}
+
+		if function.Doc == nil {
+			continue
+		}
+
+		if strings.Contains(function.Name.Name, "Handle") {
+			return strings.TrimSpace(function.Doc.Text()), nil
+		}
```
eriq-augustine, line 63: When you move this to util, makes some test cases for non-existent and empty (whitespce) comments.

internal/api/describe_test.go
```diff
@@ -14,23 +14,36 @@ func TestDescribeFull(test *testing.T) {
...
+	if err != nil {
+		test.Fatalf("Failed to describe endpoints: '%v'.", err)
+	}
+
+	for endpoint, description := range descriptions.Endpoints {
+		if description.Description == "" {
+			test.Errorf("Describe found an empty description. Endpoint: '%s'.", endpoint)
+			continue
+		}
+	}
```
eriq-augustine, line 33: I think this should be in another test (as opposed to being grouped in this other test).
Maybe "TestDescribeEmpty"?

eriq-augustine (changes requested):
Looking pretty good, just some smaller things.

docs/development.md
```diff
@@ -92,6 +92,16 @@ Within this range, each package in the API package is given a range of 100 locat
...
 
+### API File Structure
+
+The structure of API endpoint files must follow these rules:
+
+1. The file must reside in a directory that follows the same structure as the endpoint's base path from `internal/api`.
+2. The handler for the endpoint must be in a file with the same name.
+3. The handler function must contain the substring `Handle`.
+
+For example, the handler for `users/list` is called `HandleList` and located in the file `List.go` within the filepath `internal/api/users`.
```
eriq-augustine, line 103: s/List.go/list/go

eriq-augustine, line 103: The word "filepath" is a reference to the `path/filepath` Go library.
"file path" would be the correct wording.
(I know this can be confusing, since "filename" is real and different from "file name".)

eriq-augustine, line 103: I would group these last two parts of the sentence together and say something like:
"and located in the file ..." (which includes the path and name).

internal/api/core/describe.go
```diff
@@ -19,3 +29,26 @@ func SetAPIDescription(description APIDescription) {
...
 }
+
+func GetDescriptionFromHandler(basePath string) (string, error) {
+	absPath := getHandlerFilePath(basePath)
+	if !util.IsFile(absPath) {
+		return "", fmt.Errorf("Unable to find file path to API Handler. Endpoint: '%s'. Expected path: '%s'.", basePath, absPath)
+	}
+
+	re := regexp.MustCompile(`Handle`)
+	description, err := util.GetDescriptionFromFunction(absPath, re)
```
eriq-augustine, line 40: You can just pass the compile expression directly as the argument.

If you don't want to do that, then rename the variable to something more descriptive like "handlePattern".

internal/api/server/server.go
```diff
@@ -19,7 +20,12 @@ func RunServer(initiator common.ServerInitiator) (err error) {
 		return err
 	}
 
-	core.SetAPIDescription(*api.Describe(*api.GetRoutes()))
+	apiDescription, errs := api.Describe(*api.GetRoutes())
+	if errs != nil {
+		log.Warn("Found errors when describing the API endpints.", errs)
+	}
```
eriq-augustine, line 26: What errors are you thinking of that we could encounter, but still run the server?

I know an error in the description may not be something that actually stops the server from running, but we could force it to stop so people resolve the issue.

internal/util/parser.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"regexp"
+	"strings"
+)
+
+func GetDescriptionFromFunction(absPath string, re *regexp.Regexp) (string, error) {
```
eriq-augustine, line 12: `re` is not descriptive enough.

eriq-augustine, line 12: `absPath` is a bad name.
I don't think the path needs to be absolute.

eriq-augustine, line 12: This function needs a comment (especially with the pattern argument).

internal/util/parser.go
```diff
@@ -0,0 +1,43 @@
...
+	if !IsFile(absPath) {
+		return "", fmt.Errorf("Unable to find file path to function. Expected path: '%s'.", absPath)
+	}
+
+	fset := token.NewFileSet()
```
eriq-augustine, line 17: I don't like the name "fset".
In general use more descriptive names.
We don't need to go overboard, but we also don't need to worry about the size of our source files.

internal/util/parser.go
```diff
@@ -0,0 +1,43 @@
...
+	node, parseErr := parser.ParseFile(fset, absPath, nil, parser.ParseComments)
```
eriq-augustine, line 18: Stick with `err` for errors, unless you have some special situation.

internal/util/parser_test.go
```diff
@@ -0,0 +1,80 @@
+package util
+
+import (
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
```
eriq-augustine, line 9: I think you can either do all these in one function (with the classes testCases pattern (note that the strings could be placed in variables/constants at the end of the file to make it look cleaner),
or you can use a common function that does most of this common code.

(The first option sounds like the better one to me.)

eriq-augustine (changes requested):
The core code is looking good.
Mainly needs some work on docs and tests.

internal/util/parser_test.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		if description != testCase.expectedDescription {
+			test.Errorf("Case %d: Unexpected function description. Expected: '%s', actual: '%s'.", i, testCase.expectedDescription, description)
+			continue
+		}
+	}
+}
+
+const base_description = `package temp
```
eriq-augustine, line 61: Since these are const, make them all upper case.
I know that will also export them, but this is the convention we use elsewhere in this project.
(The way I think about this is that constants can't be changed anyways, so it's not a problem if they are exported.)

internal/util/parser.go
```diff
@@ -0,0 +1,46 @@
...
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"regexp"
+	"strings"
+)
+
+// This function takes a file path and a pattern that mathes the name of the target function.
+// Returns the comment attached to the target function.
+// Errors occur when the file or target function cannot be found or if the target function does not have a comment attached.
```
eriq-augustine, line 14: I'm not convinced about the error semantics here.
Returning an error on no comment (especially when we can return an empty comment (on whitespace only)) seems pretty harsh.
I think that we either need to return an empty string if we find the function (but not the comment),
or return a special error that the caller can check for to see this case
(like [io.EOF](https://pkg.go.dev/io#pkg-variables)).

Handling this can also make your tests for this function a little smoother.

(Note that this is a hard design decision, so there is no clear answer.)

eriq-augustine (changes requested):
Pretty close.
There are some design semantics to think about.

eriq-augustine (approved):
lgtm!

PR #133, opened 2024-12-20
Require acknowledgement from users when submitting assignments late.

Lucas-Ellenberger:
This PR adds a new parameter to the `courses/assignments/submissions/submit` API endpoint to signal if the user acknowledges their submission is late. It works as follows:
* If the assignment does not have a due date, either option works.
* If the student submits before the due date, either options works.
* But if the student is submitting late and does NOT acknowledge, the submission will be rejected.
* If the student is submitting late and DOES acknowledge, normal grading proceeds.

To facilitate proper testing for any interface that uses this feature, check for late submissions will run in unit testing mode. So, submission tests must supply the `allow-late` flag as needed.

This PR must be reviewed with the associated [python interface PR](https://github.com/edulinq/autograder-py/pull/36).


internal/api/courses/assignments/submissions/submit_test.go
```diff
@@ -120,3 +123,53 @@ func TestRejectSubmissionMaxAttempts(test *testing.T) {
...
+func TestRejectLateSubmission(test *testing.T) {
+	db.ResetForTesting()
+	defer db.ResetForTesting()
+
+	course := db.MustGetCourse("course101")
+	dueDate := timestamp.Zero()
+	course.Assignments["hw0"].DueDate = &dueDate
+	db.MustSaveCourse(course)
+
+	// Note that we are using a submission from a different assignment.
```
eriq-augustine, line 136: Is this comment accurate?

internal/api/courses/assignments/submissions/submit_test.go
```diff
@@ -120,3 +123,53 @@ func TestRejectSubmissionMaxAttempts(test *testing.T) {
...
 }
+
+func TestRejectLateSubmission(test *testing.T) {
+	db.ResetForTesting()
+	defer db.ResetForTesting()
+
+	course := db.MustGetCourse("course101")
+	dueDate := timestamp.Zero()
+	course.Assignments["hw0"].DueDate = &dueDate
+	db.MustSaveCourse(course)
```
eriq-augustine, line 134: This pattern is a bit strange (fetching/saving the course instead of the assignment).
I see that SaveAssignment() exists for DB backends, but is not exposed in internal/db like all the other functions.
Looks like an oversight to me.
I would just fix that and then use  db.MustGetTestAssignment() and db.SaveAssignment().

internal/api/courses/assignments/submissions/submit_test.go
```diff
@@ -24,6 +26,7 @@ func TestSubmit(test *testing.T) {
 		fields := map[string]any{
 			"course-id":     testSubmission.Assignment.GetCourse().GetID(),
 			"assignment-id": testSubmission.Assignment.GetID(),
+			"allow-late":    true,
```
eriq-augustine, line 29: I'm not a fan of this change,
but I will let it stand as long as it is commented.

However, the test for allow-late should be more targeted and specific
(this this more blanket approach).
I recommend modifying TestRejectLateSubmission() to loop over test cases where the due date is an option for the test case.

internal/grader/grade.go
```diff
@@ -18,19 +18,21 @@ var submissionLocks sync.Map
 type GradeOptions struct {
 	NoDocker     bool
 	LeaveTempDir bool
+	AllowLate    bool
 }
 
-func GetDefaultGradeOptions() GradeOptions {
+func GetDefaultGradeOptions(allowLate bool) GradeOptions {
```
eriq-augustine, line 24: This argument breaks the pattern.
(It's not much of a default if you have to select an option.)

Instead, callers should get the default option and then set AllowLate if need be.

internal/grader/reject.go
```diff
@@ -45,10 +45,48 @@ func (this *RejectWindowMax) fullString(now timestamp.Timestamp) string {
 		nextTime.SafeMessage(), deltaString)
 }
 
-func checkForRejection(assignment *model.Assignment, submissionPath string, user string, message string) (RejectReason, error) {
+type RejectLateWithoutAllow struct {
```
eriq-augustine, line 48: Change name to "RejectLate".
This could be used for more in the future, and I don't want to keep it too specific.

internal/grader/reject.go
```diff
@@ -45,10 +45,48 @@ func (this *RejectWindowMax) fullString(now timestamp.Timestamp) string {
...
+	AssignmentName string
+	DueDate        timestamp.Timestamp
+}
+
+func (this *RejectLateWithoutAllow) String() string {
+	return this.fullString(timestamp.Now())
+}
+
+func (this *RejectLateWithoutAllow) fullString(now timestamp.Timestamp) string {
```
eriq-augustine, line 57: RejectWindowMax.fullString() exists for tests.
If you aren't using this in tests, then only String() should exist.

internal/grader/reject.go
```diff
@@ -45,10 +45,48 @@ func (this *RejectWindowMax) fullString(now timestamp.Timestamp) string {
...
+	deltaMS := now.ToMSecs() - this.DueDate.ToMSecs()
+	deltaString := time.Duration(deltaMS * int64(time.Millisecond)).String()
+
+	return fmt.Sprintf("Attempting to submit assignment (%s) late without 'allow late'."+
```
eriq-augustine, line 61: s/'allow late'/the 'allow late' option/

internal/grader/reject.go
```diff
@@ -45,10 +45,48 @@ func (this *RejectWindowMax) fullString(now timestamp.Timestamp) string {
...
+		" It was due on %s (which was %s ago). Include the 'allow late' flag to submit an assignment late."+
```
eriq-augustine, line 62: Break this line on the sentence.

This is just a personal nit, but we may as well do it while we are here.

eriq-augustine, line 62: You can't use the word "flag" (that implies a CLI).
You can use "option".

eriq-augustine, line 62: s/Include/Use/g

This is more general.
(Include isn't used for things like CLI.)

internal/grader/reject.go
```diff
@@ -45,10 +45,48 @@ func (this *RejectWindowMax) fullString(now timestamp.Timestamp) string {
...
+		" See your interface's help section for more information.",
```
eriq-augustine, line 63: s/help section/documentation/

internal/grader/reject.go
```diff
@@ -45,10 +45,48 @@ func (this *RejectWindowMax) fullString(now timestamp.Timestamp) string {
...
+	reason := checkLateSubmission(assignment, allowLate)
+	if reason != nil {
+		return reason, nil
+	}
+
 	return checkSubmissionLimit(assignment, user)
 }
 
+func checkLateSubmission(assignment *model.Assignment, allowLate bool) RejectReason {
+	now := timestamp.Now()
```
eriq-augustine, line 77: Do the nil check first.
(It's small, but generally we want to see if we have to do work before we do it.)

eriq-augustine (changes requested):
The core here is looking pretty good.

internal/api/courses/assignments/submissions/submit_test.go
```diff
@@ -120,3 +125,95 @@ func TestRejectSubmissionMaxAttempts(test *testing.T) {
...
+				i, responseContent.Rejected, testCase.expectReject, util.MustToJSONIndent(responseContent))
+			continue
+		}
+
+		if testCase.expectReject {
+			expected := (&grader.RejectLate{assignment.Name, *assignment.DueDate}).String()
+			expected = timeDeltaPattern.ReplaceAllString(expected, timeDeltaReplacement)
+			actual := timeDeltaPattern.ReplaceAllString(responseContent.Message, timeDeltaReplacement)
+			if expected != actual {
+				test.Fatalf("Case %d: Did not get the expected rejection reason. Expected: '%s', Actual: '%s'.",
```
eriq-augustine, line 208: s/Fatalf/Errorf/

Put in a continue after as well.

internal/db/assignment.go
```diff
@@ -42,3 +42,23 @@ func GetAssignment(rawCourseID string, rawAssignmentID string) (*model.Assignmen
 
 	return assignment, nil
 }
+
+func MustSaveAssignment(assignment *model.Assignment) {
+	err := SaveAssignment(assignment)
+	if err != nil {
+		log.Fatal("Failed to save course.", err)
```
eriq-augustine, line 49: s/course/assignment/

eriq-augustine (changes requested):
Looking great, just a couple small things.

eriq-augustine (approved):
lgtm

PR #134, opened 2024-12-21
Fix for TestCMDStartsOwnServer.

OliverLok:
It seems like procedures/server/start.go's defer of removing temp directories would occasionally get executed while another test would run. Since TestCMDStartsOwnServer already deals with removing temp dirs, procedures doesn't need to when unit testing.

I also noticed in ~1/100 tests that the server wouldn't have enough time to start up for the test, so I increased its sleep to 200 milliseconds.

After these changes, I ran TestCMDStartsOwnServer a few hundred times with no issues.

### DIFF
--- a/internal/api/server/server.go
+++ b/internal/api/server/server.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"os"
 	"os/signal"
-	"sync"
 	"syscall"
 
 	"github.com/edulinq/autograder/internal/api"
@@ -13,12 +12,20 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
-// FinishCleanup ensures cleanup tasks execute before the server fully stops.
-// Add(1) to the WaitGroup for each defered cleanup task and call Done() when finished.
-var FinishCleanup sync.WaitGroup
+type APIServer struct {
+	errorsChan     chan error
+	shutdownSignal chan os.Signal
+}
+
+func NewAPIServer() *APIServer {
+	return &APIServer{
+		errorsChan:     make(chan error),
+		shutdownSignal: make(chan os.Signal),
+	}
+}
 
 // Run the autograder server and listen on an http and unix socket.
-func RunServer(initiator common.ServerInitiator) (err error) {
+func (this *APIServer) RunServer(initiator common.ServerInitiator) (err error) {
 	err = common.WriteAndHandleStatusFile(initiator)
 	if err != nil {
 		return err
@@ -31,50 +38,39 @@ func RunServer(initiator common.ServerInitiator) (err error) {
 
 	core.SetAPIDescription(*apiDescription)
 
-	FinishCleanup.Add(1)
 	defer func() {
 		err = errors.Join(err, util.RemoveDirent(common.GetStatusPath()))
-		FinishCleanup.Done()
-	}()
-
-	errorsChan := make(chan error, 2)
-
-	go func() {
-		errorsChan <- runAPIServer(api.GetRoutes())
 	}()
 
 	go func() {
-		errorsChan <- runUnixSocketServer()
+		this.errorsChan <- runAPIServer(api.GetRoutes())
+	}()
+
+	go func() {
+		this.errorsChan <- runUnixSocketServer()
 	}()
 
 	// Gracefully shutdown on Control-C (SIGINT).
-	shutdownSignal := make(chan os.Signal, 1)
-	signal.Notify(shutdownSignal, syscall.SIGINT, syscall.SIGTERM)
+	signal.Notify(this.shutdownSignal, syscall.SIGINT, syscall.SIGTERM)
 	go func() {
-		<-shutdownSignal
-		signal.Stop(shutdownSignal)
-		StopServer(true)
+		<-this.shutdownSignal
+		signal.Stop(this.shutdownSignal)
+		this.StopServer()
 	}()
 
 	// Wait for at least one error (or nil) to stop both servers,
 	// then wait for the next error (or nil).
-	err = errors.Join(err, <-errorsChan)
+	err = errors.Join(err, <-this.errorsChan)
 	// Stop server without waiting to ensure cleanup tasks get executed.
-	StopServer(false)
-	err = errors.Join(err, <-errorsChan)
+	this.StopServer()
+	err = errors.Join(err, <-this.errorsChan)
 
-	close(errorsChan)
+	close(this.errorsChan)
 
 	return err
 }
 
-// waitForCleanup should always be set to true when stopping a server
-// to ensure the server is fully cleaned up by the time it's stopped.
-func StopServer(waitForCleanup bool) {
+func (this *APIServer) StopServer() {
 	stopUnixSocketServer()
 	stopAPIServer()
-
-	if waitForCleanup {
-		FinishCleanup.Wait()
-	}
 }
--- a/internal/api/server/server.go
+++ b/internal/api/server/server.go
@@ -25,7 +25,7 @@ func NewAPIServer() *APIServer {
 }
 
 // Run the autograder server and listen on an http and unix socket.
-func (this *APIServer) RunServer(initiator common.ServerInitiator) (err error) {
+func (this *APIServer) Run(initiator common.ServerInitiator) (err error) {
 	err = common.WriteAndHandleStatusFile(initiator)
 	if err != nil {
 		return err
@@ -55,14 +55,14 @@ func (this *APIServer) RunServer(initiator common.ServerInitiator) (err error) {
 	go func() {
 		<-this.shutdownSignal
 		signal.Stop(this.shutdownSignal)
-		this.StopServer()
+		this.Stop()
 	}()
 
 	// Wait for at least one error (or nil) to stop both servers,
 	// then wait for the next error (or nil).
 	err = errors.Join(err, <-this.errorsChan)
 	// Stop server without waiting to ensure cleanup tasks get executed.
-	this.StopServer()
+	this.Stop()
 	err = errors.Join(err, <-this.errorsChan)
 
 	close(this.errorsChan)
@@ -70,7 +70,7 @@ func (this *APIServer) RunServer(initiator common.ServerInitiator) (err error) {
 	return err
 }
 
-func (this *APIServer) StopServer() {
+func (this *APIServer) Stop() {
 	stopUnixSocketServer()
 	stopAPIServer()
 }
--- a/internal/api/server/server.go
+++ b/internal/api/server/server.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 
 	"github.com/edulinq/autograder/internal/api"
@@ -11,6 +12,8 @@ import (
 	"github.com/edulinq/autograder/internal/common"
 	"github.com/edulinq/autograder/internal/util"
 )
+
+var FinishCleanup sync.WaitGroup
 
 // Run the autograder server and listen on an http and unix socket.
 func RunServer(initiator common.ServerInitiator) (err error) {
--- a/internal/api/server/server.go
+++ b/internal/api/server/server.go
@@ -13,6 +13,8 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
+// FinishCleanup ensures cleanup tasks execute before the server fully stops.
+// Add(1) to the WaitGroup for each defered cleanup task and call Done() when finished.
 var FinishCleanup sync.WaitGroup
 
 // Run the autograder server and listen on an http and unix socket.
@@ -29,8 +31,10 @@ func RunServer(initiator common.ServerInitiator) (err error) {
 
 	core.SetAPIDescription(*apiDescription)
 
+	FinishCleanup.Add(1)
 	defer func() {
 		err = errors.Join(err, util.RemoveDirent(common.GetStatusPath()))
+		FinishCleanup.Done()
 	}()
 
 	errorsChan := make(chan error, 2)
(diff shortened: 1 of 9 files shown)


internal/cmd/test.go
```diff
@@ -165,23 +165,33 @@ func runCMD(mainFunc func(), args []string) (err error) {
 
 func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonTestCase CommonCMDTestCase, prefix string) (string, string, int, bool) {
 	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args, commonTestCase.LogLevel)
+
+	logOutputs := func() {
+		test.Logf("Stdout: '%s'.", stdout)
+		test.Logf("Stderr: '%s'.", stderr)
+	}
```
eriq-augustine, line 172: Make this a real function.

internal/cmd/test.go
```diff
@@ -165,23 +165,33 @@ func runCMD(mainFunc func(), args []string) (err error) {
 
 func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonTestCase CommonCMDTestCase, prefix string) (string, string, int, bool) {
 	stdout, stderr, exitCode, err := RunCMDTest(test, mainFunc, args, commonTestCase.LogLevel)
+
+	logOutputs := func() {
+		test.Logf("Stdout: '%s'.", stdout)
+		test.Logf("Stderr: '%s'.", stderr)
```
eriq-augustine, line 171: Change the formatting so it is more clear what is in stdout/stderr.
Like put markers around it and start it on a new line.

internal/cmd/server.go
```diff
@@ -55,7 +55,7 @@ func mustEnsureServerIsRunning() (bool, int) {
 	serverStart.Wait()
 
 	// Small sleep to allow the server to start up.
-	time.Sleep(100 * time.Millisecond)
+	time.Sleep(200 * time.Millisecond)
```
eriq-augustine, line 58: Is this necessary?

internal/procedures/server/start.go
```diff
@@ -48,7 +48,9 @@ func Start(initiator common.ServerInitiator) (err error) {
 	}
 
 	// Cleanup any temp dirs.
-	defer util.RemoveRecordedTempDirs()
+	if !config.UNIT_TESTING_MODE.Get() {
```
eriq-augustine, line 51: Add a comment as to why you are doing this.

eriq-augustine (changes requested):
You have correctly identified the process causing some of the issues (removing temp dirs out from under another procedure because the server cleanup routines are leaking into other tests).
Good work, that is very tricky.

However, you have not yet addressed the core issue:
the server is not fully cleaned up when other code is called.
What you have here is more of a band-aid.
Think over what a full solution may be and chat with me.

eriq-augustine:
Once this gets in, we can merge https://github.com/edulinq/autograder-py/pull/36 .

internal/api/server/server.go
```diff
@@ -12,6 +13,8 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
+var FinishCleanup sync.WaitGroup
```
eriq-augustine, line 16: Comment this.
This should probably be a couple sentences to describe what it is, why we need it, and how people should use it.

internal/cmd/test.go
```diff
@@ -167,23 +167,32 @@ func RunCommonCMDTests(test *testing.T, mainFunc func(), args []string, commonTe
...
+		logOutputs(test, stdout, stderr)
 		return "", "", -1, false
 	}
 
 	return stdout, stderr, exitCode, true
 }
+
+func logOutputs(test *testing.T, stdout string, stderr string) {
+	test.Logf("\n--- START OF STDOUT ---\n%s\n--- END OF STDOUT---", stdout)
+	test.Logf("\n--- START OF STDERR ---\n%s\n--- END OF STDERR ----", stderr)
```
eriq-augustine, line 197: I'm not a fan of the formatting here:
 - Starting with a newline is a bit strange.
 - The "start of"/"end of" wording seems a bit unnecessary (this is the output of a failing test, not something sent to a normal user).
 - Missing space after "END OF STDOUT".

We already have an example for this stdout/stderr pattern in a test:
https://github.com/edulinq/autograder-server/blob/main/internal/grader/grade_test.go#L82

internal/procedures/server/start.go
```diff
@@ -29,8 +29,11 @@ func Start(initiator common.ServerInitiator) (err error) {
 		return fmt.Errorf("Failed to open the database: '%w'.", err)
 	}
 
+	// Ensure the database closes before a CMD that started a server finishes its execution.
```
eriq-augustine, line 32: I wouldn't comment this, since there are multiple places this pattern will be used.

internal/procedures/server/start.go
```diff
@@ -47,8 +50,15 @@ func Start(initiator common.ServerInitiator) (err error) {
 		go startCourse(course)
 	}
 
-	// Cleanup any temp dirs.
-	defer util.RemoveRecordedTempDirs()
+	// Don't remove temp dirs during unit testing since they are needed after this function finishes.
+	if !config.UNIT_TESTING_MODE.Get() {
+		// Ensure temp dirs are removed before a CMD that started a server finishes its execution.
```
eriq-augustine, line 55: This comment seems overly specific (and therefore incorrect).

internal/procedures/server/start.go
```diff
@@ -47,8 +50,15 @@ func Start(initiator common.ServerInitiator) (err error) {
 		go startCourse(course)
 	}
 
-	// Cleanup any temp dirs.
-	defer util.RemoveRecordedTempDirs()
+	// Don't remove temp dirs during unit testing since they are needed after this function finishes.
+	if !config.UNIT_TESTING_MODE.Get() {
```
eriq-augustine, line 54: I would like to avoid special cases for testing if we can
(that was a large chunk of our last chat).

Can this be handled with an argument instead (cleanupTempDirs?)?

internal/cmd/request.go
```diff
@@ -34,6 +34,9 @@ func MustHandleCMDRequestAndExitFull(endpoint string, request any, responseType
 	func() {
 		startedCMDServer, oldPort := mustEnsureServerIsRunning()
 		if startedCMDServer {
+			// Wait for the server to finish cleaning up before this func finishes.
+			defer server.FinishCleanup.Wait()
```
eriq-augustine, line 38: What about for cmd/server?
Shouldn't this be in api.server.StopServer()?

eriq-augustine (changes requested):
The core seems pretty sound,
but it seems like this my be overly specific.
You may have missed the forest for the trees a bit.

You have code (and comments) specifically for the case where a server is started from a (non cmd/sever) CMD,
but what about cmd/Sevrer?
Should it also clean up properly?

eriq-augustine (changes requested):
We discussed these in more detail synchronously.

internal/cmd/test.go
```diff
@@ -79,7 +87,13 @@ func RunCMDTest(test *testing.T, mainFunc func(), args []string, logLevel log.Lo
 	exit.SetShouldExitForTesting(false)
 	defer exit.SetShouldExitForTesting(true)
 
+	// Prevent the cmd testing dir from automatic cleanup to use during testing.
+	util.SetShouldRemoveTempDirs(false)
+	defer util.SetShouldRemoveTempDirs(true)
+
 	tempDir := util.MustMkDirTemp("autograder-testing-cmd-")
+	defer util.RemoveRecordedTempDirs()
```
eriq-augustine, line 95: ! Bug

First, before the bug, this line should be grouped with 91/92.
All these are about manipulating how temp dirs are removed during this test.
94 is just a normal line that creates a temp dir using the standard procedure,
it should not be grouped into the "doing unusual things with temp dirs" group.

Now the bug.
Remember that defers are kept as a stack, so later ones happen first.
This will run before the defer on 92.

internal/procedures/server/start.go
```diff
@@ -14,9 +14,13 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
-func Start(initiator common.ServerInitiator) (err error) {
-	defer server.StopServer()
+type ProcedureServer struct {
+	apiServer *server.APIServer
+}
+
+var serverInstance *ProcedureServer
```
eriq-augustine, line 21: When defining variables like this, I prefer explicitly setting their value (`= nil`).
It does not change the program, but makes it more clear.

internal/procedures/server/start.go
```diff
@@ -14,9 +14,13 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
-func Start(initiator common.ServerInitiator) (err error) {
-	defer server.StopServer()
+type ProcedureServer struct {
```
eriq-augustine, line 17: Because your start server function is not-exported,
I don't see too much use of this struct.
Unlike the API server, the functionality for working with the server in his package are functions (which are not all exported) and not methods.

I think the pattern here generally makes sense (and it makes sense that api/server and procedures/server have different patterns),
just get rid of this struct and store the api server instance instead.

internal/api/server/server.go
```diff
@@ -12,8 +12,20 @@ import (
...
+func NewAPIServer() *APIServer {
+	return &APIServer{
+		errorsChan:     make(chan error),
+		shutdownSignal: make(chan os.Signal),
+	}
+}
+
 // Run the autograder server and listen on an http and unix socket.
-func RunServer(initiator common.ServerInitiator) (err error) {
+func (this *APIServer) RunServer(initiator common.ServerInitiator) (err error) {
```
eriq-augustine, line 28: Since this is now a method, drop "Server" from the method name.

internal/procedures/server/start.go
```diff
@@ -47,15 +47,49 @@ func Start(initiator common.ServerInitiator) (err error) {
...
+
+	serverInstance.apiServer.StopServer()
+
+	serverInstance.apiServer = nil
+	serverInstance = nil
+
+	return err
+}
+
+func RunAndBlockServer(initiator common.ServerInitiator, skipSetup bool) (err error) {
```
eriq-augustine, line 69: Given the name of this package, I would drop "Server" from these function names.

internal/procedures/server/start.go
```diff
@@ -47,15 +47,49 @@ func Start(initiator common.ServerInitiator) (err error) {
...
 
-	// Cleanup any temp dirs.
-	defer util.RemoveRecordedTempDirs()
+	return nil
+}
 
-	err = server.RunServer(initiator)
-	if err != nil {
-		return fmt.Errorf("Error during server startup sequence: '%w'.", err)
+func CleanupAndStopServer() (err error) {
```
eriq-augustine, line 53: These is a race condition here.
Pretty much on anything that touches serverInstance/serverInstance.apiServer.

Make a normal lock and use it normally on cleanup (lock the full function).
But, you will have to be more specific for setup
(mainly because you assign serverInstance.apiServer and do setup in different functions).
I recommend making another function that assigns serverInstance.apiServer, takes skipSetup, and calls setup();
then you can lock that entire function instead of locking just chunks of code in RunAndBlock() (which you would have to do now).

internal/procedures/server/start.go
```diff
@@ -47,15 +47,49 @@ func Start(initiator common.ServerInitiator) (err error) {
...
+
+		if !skipSetup {
+			err = setupServer(initiator)
+			if err != nil {
+				err = fmt.Errorf("Failed to setup the server: '%w'.", err)
+				return
+			}
+		}
+
+		err = serverInstance.apiServer.RunServer(initiator)
```
eriq-augustine, line 88: You need a comment here noting that serverInstance may be nil when this call completes.

eriq-augustine (changes requested):
The core here is looking good.
There's a timing bug and race condition to resolve.

internal/cmd/test.go
```diff
@@ -74,12 +82,19 @@ func CMDServerTestingMain(suite *testing.M) {
 	exit.Exit(code)
 }
 
-func RunCMDTest(test *testing.T, mainFunc func(), args []string, logLevel log.LogLevel) (string, string, int, error) {
+func RunCMDTest(test *testing.T, mainFunc func(), args []string, logLevel log.LogLevel, removeTempDirsAfterEachTest bool) (string, string, int, error) {
```
eriq-augustine, line 85: The name "removeTempDirsAfterEachTest" is not good.
This function runs one test, but the name implies multiple tests.

internal/cmd/test.go
```diff
@@ -74,12 +82,19 @@ func CMDServerTestingMain(suite *testing.M) {
...
 	exit.SetShouldExitForTesting(false)
 	defer exit.SetShouldExitForTesting(true)
 
+	// Retain temp dirs to use after tests complete.
 	tempDir := util.MustMkDirTemp("autograder-testing-cmd-")
+	util.SetShouldRemoveTempDirs(false)
+	if removeTempDirsAfterEachTest {
+		defer util.RemoveRecordedTempDirs()
+	}
+	defer util.SetShouldRemoveTempDirs(true)
```
eriq-augustine, line 96: This is a very special situation that only one test needs.
We can just set the value in that one test.

internal/procedures/server/start.go
```diff
@@ -47,19 +45,69 @@ func Start(initiator common.ServerInitiator) (err error) {
...
+		err := setup(initiator)
+		if err != nil {
+			return fmt.Errorf("Failed to setup the server: '%w'.", err)
+		}
+	}
+
+	return nil
+}
+
+func RunAndBlock(initiator common.ServerInitiator, skipSetup bool) (err error) {
```
eriq-augustine, line 87: Our typical pattern is to hide advanced options like skipSetup() using a *Full() function.

internal/util/dir.go
```diff
@@ -62,16 +63,24 @@ func RemoveRecordedTempDirs() error {
...
 		errs = errors.Join(errs, RemoveDirent(dir))
 	}
 
-	ClearRecordedTempDirs()
+	clearRecordedTempDirs()
 
 	return errs
 }
 
+func SetShouldRemoveTempDirs(shouldRemove bool) {
```
eriq-augustine, line 80: Race condition on shouldRemoveTempDirs.

eriq-augustine (approved):
In general, it looks good.
There are still minor issues, but I am going to accept and fix them on my side because of the time constraints.

PR #136, opened 2025-01-03
Initial Stats Infrastructure

eriq-augustine:
- Added some initial infrastructure for stats.
 - Added collection for system stats.
 - Added an API endpoint for querying system stats.

Sibling Python PR: https://github.com/edulinq/autograder-py/pull/37


internal/db/stats_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		test.Fatalf("Failed to fetch stats: '%v'.", err)
+	}
+
+	if len(records) != 1 {
+		test.Fatalf("Did not get the correct number of records. Expected: 1, Actual: %d.", len(records))
+	}
+
+	if *records[0] != testRecord {
+		test.Fatalf("Did not get the expected record back. Expected '%s', Actual: '%s'.",
```
OliverLok, line 42: Missing colon after Expected

internal/stats/query.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/edulinq/autograder/internal/timestamp"
+)
+
+type Query interface {
+	// Does this metric match the filtering conditions of this query.
+	Match(record Metric) bool
+}
+
+// The base for a stats query.
+// Note that this the semantics of this struct mean that times before UNIX epoch (negative times)
```
OliverLok, line 15: Typo. Remove the first "this".

internal/stats/query.go
```diff
@@ -0,0 +1,95 @@
...
+// must be offset by at least one MS (as a zero value is treated as the end of time).
+type BaseQuery struct {
+	// Limit the number of results.
+	// Take this number of results from the top.
```
OliverLok, line 19: s/this/the

internal/stats/query.go
```diff
@@ -0,0 +1,95 @@
...
+	sortOrder := baseQuery.Sort
+	if sortOrder < 0 {
+		sortOrder = -1
+	} else if sortOrder > 0 {
+		sortOrder = 1
+	}
+
+	results := make([]T, 0, len(metrics))
+
+	// First filter.
```
OliverLok, line 70: s/First/First,

internal/stats/system.go
```diff
@@ -0,0 +1,148 @@
...
+	ctx = nil
+	cancelFunc = nil
+	cancelWait = nil
+}
+
+// Get the system metrics.
+// Specific metrics have quirks the caller should know about.
+//
+// CPU:
+// The get CPU statistics, we have to make an internal query, wait, make another internal query, and compare the results.
```
OliverLok, line 96: s/The get/To get

OliverLok (changes requested):
Just some small typos.

internal/stats/query.go
eriq-augustine: "This" is actually correct in this case.
("This" refers to the variable, so "take this number of results from the top".)

internal/stats/system.go
```diff
@@ -0,0 +1,148 @@
...
+		return
+	}
+
+	// Signal that we have fully stopped collecting stats.
+	cancelWait.Add(1)
+	defer cancelWait.Done()
+
+	for {
+		select {
+		case <-ctx.Done():
```
eriq-augustine, line 40: A comment here about the context cancellation and timing control would be nice.

eriq-augustine (commented):
One issue, one case of sharing code, and one missing comment.

internal/db/disk/stats.go
```diff
@@ -0,0 +1,37 @@
...
+	return util.AppendJSONLFile(this.getSystemStatsPath(), record)
+}
+
+func (this *backend) GetSystemStats(query stats.Query) ([]*stats.SystemMetrics, error) {
+	this.statsLock.RLock()
+	defer this.statsLock.RUnlock()
+
+	path := this.getSystemStatsPath()
+	if !util.PathExists(path) {
+		return make([]*stats.SystemMetrics, 0), nil
```
OliverLok, line 25: Why not return nil, nil?

OliverLok (changes requested):
One small thing.

internal/db/disk/stats.go
eriq-augustine: So that it outputs as `[]` instead of `null` in JSON.

PR #145, opened 2025-02-06
Added stats for how much time was spent on tasks. 

OliverLok:


internal/stats/course.go
```diff
@@ -12,9 +12,12 @@ type CourseMetricType string
 const (
 	CourseMetricTypeUnknown          CourseMetricType = ""
 	CourseMetricTypeGradingTime                       = "grading-time"
+	CourseMetricTypeTaskDuration                      = "task-duration"
```
eriq-augustine, line 15: Maybe use `task-time` to match the other metric types.

internal/stats/course.go
```diff
@@ -12,9 +12,12 @@ type CourseMetricType string
...
 	CourseMetricTypeCodeAnalysisTime                  = "code-analysis-time"
 )
 
+const TASK_KEY = "task"
```
eriq-augustine, line 19: Since you have a const for this, you should make one for "anslysis-type" (in internal/analysis/common.go) as well.

Then you can also have a more descriptive name like "ATTRIBUTE_KEY_TASK" (and a similar name for the analysis type).

internal/tasks/core.go
```diff
@@ -80,15 +81,18 @@ func runNextTask() {
...
-	now := timestamp.Now()
-	if task.NextRunTime > now {
+	startTimestamp := timestamp.Now()
+	if task.NextRunTime > startTimestamp {
 		return
 	}
 
 	log.Debug("Task started.", task)
 	runTask(task)
+	endTimestamp := timestamp.Now()
```
eriq-augustine, line 91: I wouldn't bother with this here.
You can just use `timestamp.Now()` in the call to stats.AsyncStoreCourseTaskDuration().
It would only be adding in the time of a log, which is pretty small.

eriq-augustine (approved):
LGTM!

internal/tasks/core.go
```diff
@@ -80,15 +81,17 @@ func runNextTask() {
...
+	startTimestamp := timestamp.Now()
+	if task.NextRunTime > startTimestamp {
 		return
 	}
 
 	log.Debug("Task started.", task)
 	runTask(task)
 	log.Debug("Task finished.", task)
 
+	stats.AsyncStoreCourseTaskTime(startTimestamp, timestamp.Now(), task.CourseID, task.AssignmentID, task.UserEmail, task.Name)
```
eriq-augustine, line 93: You should use the task's type instead of name.
The type will always exist, but the name can be empty.

eriq-augustine (changes requested):
Sorry, I gave an lgtm too early.

eriq-augustine (approved):
lgtm

PR #151, opened 2025-02-20
Added request stat metric storing

OliverLok:


internal/stats/request.go
```diff
@@ -0,0 +1,47 @@
...
+type RequestMetric struct {
+	BaseMetric
+
+	Endpoint     string `json:"endpoint"`
+	Locator      string `json:"locator,omitempty"`
+	CourseID     string `json:"course,omitempty"`
+	AssignmentID string `json:"assignment,omitempty"`
+	UserEmail    string `json:"user,omitempty"`
+	IPAddress    string `json:ip-address`
+	Value        uint64 `json:"duration"`
```
eriq-augustine, line 18: Why call this "Value"?
For things like course metrics, we don't know what will be stored in the field.
But here, we know what it will be.

internal/util/string.go
```diff
@@ -18,3 +18,11 @@ func BaseString(obj any) string {
 func JoinStrings(delim string, parts ...string) string {
 	return strings.Join(parts, delim)
 }
+
+func GetFirstNonEmptyString(primary string, fallback string) string {
```
eriq-augustine, line 22: This name is a bit off.
What if both are empty?

This concept is often called a "default",
so maybe "GetStringWithDefault()"?

eriq-augustine (changes requested):
What is here is pretty good, but there are two big missing pieces:
 - Fetching the metric from the DB.
   - You store it, but don't have anything the read it.
 - An API endpoint to fetch the data.

internal/api/core/response.go
```diff
@@ -70,3 +71,56 @@ func getRequestInfo(request ValidAPIRequest) (string, timestamp.Timestamp) {
...
+		assignmentID = util.GetStringWithDefault(assignmentID, apiError.AssignmentID)
+		userEmail = util.GetStringWithDefault(userEmail, apiError.UserEmail)
+		locator = apiError.Locator
+	}
+
+	return endpoint, userEmail, courseID, assignmentID, locator
+}
+
+// Reflexively get the endpoint, userEmail, courseID, and assignmentID from a ValidAPIRequest.
+func getValidAPIRequestInfoForStats(request ValidAPIRequest) (string, string, string, string) {
```
eriq-augustine, line 93: I don't like this name.
The "valid" part is tough.
Although it is in the type name of the argument, it is also very overloaded.

Maybe something like: "getBasicAPIRequestInfo".

internal/api/core/response.go
```diff
@@ -70,3 +71,56 @@ func getRequestInfo(request ValidAPIRequest) (string, timestamp.Timestamp) {
 
 	return id, startTime
 }
+
+// Get the endpoint, userEmail, courseID, assignmentID, and locator
+// from a ValidAPIRequest and an APIError, both of which may be nil.
+func getRequestInfoForStats(request ValidAPIRequest, apiError *APIError) (string, string, string, string, string) {
```
eriq-augustine, line 77: I don't think the "ForStats" part of the name is necessary.
There's nothing here that is specific to stats.

internal/api/core/routing.go
```diff
@@ -146,10 +150,16 @@ func sendAPIResponse(apiRequest ValidAPIRequest, response http.ResponseWriter,
 
 			payload, _ = util.ToJSON(apiResponse)
 		} else {
-			return sendAPIResponse(apiRequest, response, nil, apiErr, true)
+			return sendAPIResponse(apiRequest, response, nil, apiErr, true, startTime, sender)
 		}
 	}
 
+	// Don't log request stats when in testing mode.
```
eriq-augustine, line 157: Why disable the stat collection?
All our other stats log during testing.
https://github.com/edulinq/autograder-server/blob/main/internal/grader/grade.go#L116

They may do it synchronously, but they still collect.

internal/api/stats/request/routes.go
```diff
@@ -0,0 +1,15 @@
+package request
+
+// All the API endpoints handled by this package.
+
+import (
+	"github.com/edulinq/autograder/internal/api/core"
+)
+
+var routes []core.Route = []core.Route{
+	core.MustNewAPIRoute(`stats/request/query`, HandleQuery),
```
eriq-augustine, line 10: I don't like the name "request" for this endpoint.
It is way too general.
"api" would be fine.

internal/api/stats/request/query_test.go
```diff
@@ -0,0 +1,113 @@
...
+		email          string
+		permError      bool
+		query          stats.BaseQuery
+		expectedValues []int
+	}{
+		{"server-admin", false, stats.BaseQuery{}, []int{100, 200, 300}},
+		{"server-admin", false, stats.BaseQuery{Sort: 1}, []int{300, 200, 100}},
+		{"server-admin", false, stats.BaseQuery{After: timestamp.FromMSecs(150)}, []int{200, 300}},
+
+		{"server-user", true, stats.BaseQuery{}, []int{}},
```
eriq-augustine, line 27: Use nil for the last value.

internal/db/database.go
eriq-augustine, line 1: You're missing StoreRequestMetric().

internal/stats/request_test.go
```diff
@@ -0,0 +1,53 @@
...
+	if count != 1 {
+		test.Fatalf("Got an unexpected number of metrics. Expected: 1, Actual: %d.", len(typedBackend.request))
+	}
+
+	// Compare the stored metric with the expected one.
+	if !reflect.DeepEqual(expectedMetric, typedBackend.request[0]) {
+		test.Fatalf("Stored metric is not as expected. Expected: '%s', Actual: '%s'.",
+			util.MustToJSONIndent(expectedMetric), util.MustToJSONIndent(typedBackend.request[0]))
+	}
+
```
eriq-augustine, line 52: Extra newline.

eriq-augustine (changes requested):
Generally looking good.

But, you're missing a test here.
You have tested the API endpoint, but you need to test the deeper functionality that requests are actually properly logged.
You need to have a test that makes requests (of different types/status) and ensure they are logged properly.

internal/stats/request.go
eriq-augustine, line 1: I'm still not a fan of the "request" name for this file, it's very overloaded.
Maybe "apirequest"?

internal/api/stats/api/query_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/edulinq/autograder/internal/stats"
+	"github.com/edulinq/autograder/internal/timestamp"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestQuery(test *testing.T) {
+	db.ResetForTesting()
+	defer db.ResetForTesting()
+
+	testCases := []struct {
```
eriq-augustine, line 17: There will need to be more query-focused test cases.
See: internal/api/courses/stats/query_test.go

eriq-augustine (changes requested):
I do not like the structure of the `request_log_test.go` tests.
All these should be in a single test case, not scattered throughout several different directories.
This is something you should really think over.
Having almost identical copies of testing infrastructure over several files should smell to you.
Change the location if you need the routes.

The backend work is looking good.

internal/stats/metrics.go
```diff
@@ -15,6 +15,12 @@ type BaseMetric struct {
 	Attributes map[string]any `json:"attributes,omitempty"`
 }
 
+type CourseAssignmentEmailMetric struct {
```
eriq-augustine, line 18: I don't like this name, this is not a "metric".

internal/api/server/apirequest_log_test.go
eriq-augustine, line 1: The records we are keeping are not "logs", they are "stats" or "metrics".
"logs" are typically for the things handled by internal/log.

internal/api/server/apirequest_log_test.go
```diff
@@ -0,0 +1,180 @@
...
+			},
+		},
+	}
+
+	port, err := util.GetUnusedPort()
+	if err != nil {
+		test.Fatalf("Failed to get an unused port: '%v'.", err)
+	}
+
+	oldPort := config.WEB_HTTP_PORT.Get()
```
eriq-augustine, line 128: Can you just use the standard test server like most of the API endpoint tests?
server_test.go needs very low level control,
but I don't think you need to do that here.

internal/api/server/apirequest_log_test.go
```diff
@@ -0,0 +1,180 @@
...
+	defer config.WEB_HTTP_PORT.Set(oldPort)
+
+	// Adjust the server address for test client.
+	oldURL := core.SetTestServerURL(fmt.Sprintf("http://127.0.0.1:%d", port))
+	defer core.SetTestServerURL(oldURL)
+
+	for i, testCase := range testCases {
+		db.ResetForTesting()
+
+		runServerTestBaseFull(test, testCase.endpoint, testCase.fields, testCase.email, testCase.expectedLocator, fmt.Sprintf("Case %d: ", i))
```
eriq-augustine, line 139: I don't think you will need to use this function (see above comment),
but if you do then you need to modify it so you can test test.Errorf() instead of test.Fatalf().
Otherwise you will be skipping test cases on each error.

internal/stats/backend.go
```diff
@@ -9,7 +9,13 @@ var backendLock sync.RWMutex
 
 type StorageBackend interface {
 	StoreSystemStats(record *SystemMetrics) error
+	GetSystemStats(query Query) ([]*SystemMetrics, error)
+
 	StoreCourseMetric(record *CourseMetric) error
+	GetCourseMetrics(query CourseMetricQuery) ([]*CourseMetric, error)
+
+	StoreAPIRequestMetric(record *APIRequestMetric) error
+	GetAPIRequestMetrics(query APIRequestMetricQuery) ([]*APIRequestMetric, error)
```
eriq-augustine, line 18: Flip the pair's ordering (get then store).

internal/stats/metrics.go
```diff
@@ -15,6 +15,12 @@ type BaseMetric struct {
...
+	CourseID     string `json:"course,omitempty"`
+	AssignmentID string `json:"assignment,omitempty"`
+	UserEmail    string `json:"user,omitempty"`
+}
```
eriq-augustine, line 22: Abstracting this out seemed to be a pretty big pain to several places that created stats (or queries).

Of course, abstraction is usually a good thing, because it makes our future lives easier.
But here, I'm not sure what we gained from it.
It was embedded into a few structs, but then never pulled out of the struct for commons use, e.g.,
```
someFunc(myCourseMetric.CourseAssignmentEmailMetric)
someFunc(myAPIRequestMetric.CourseAssignmentEmailMetric)
```

I would think about it and see if you think we should keep the abstraction or just put the fields in directly.

eriq-augustine (changes requested):
The tests are looking much better.
There are still some testing things to look into, but not as many.

internal/api/stats/apirequest/query_test.go
```diff
@@ -0,0 +1,135 @@
...
+				if testCase.permErrorLocator != response.Locator {
+					test.Errorf("Case %d: Incorrect locator on perm error. Expected: '%s', Actual: '%s'.", i, testCase.permErrorLocator, response.Locator)
+				}
+			} else {
+				test.Errorf("Case %d: Response is not a success when it should be: '%v'.", i, response)
+			}
+
+			continue
+		}
+
```
eriq-augustine, line 76: Missing the case where `response.Success && (testCase.permErrorLocator == "")`.

We see a similar pattern when we use `errorSubstring`:
https://github.com/edulinq/autograder-server/blob/main/internal/util/fileop_test.go#L236

internal/stats/apirequest.go
```diff
@@ -0,0 +1,95 @@
...
+		return false
+	}
+
+	if (this.Endpoint != "") && (this.Endpoint != record.Endpoint) {
+		return false
+	}
+
+	if (this.Locator != "") && (this.Locator != record.Locator) {
+		return false
+	}
```
eriq-augustine, line 64: I would change the order of these so that the most specific cases get checked first (this way we can best take advantage of short circuiting).

eriq-augustine (changes requested):
Looking pretty good, just a few small things.

With this set, you should code up the Python side (which should be pretty easy).
I can't remember if you have used the verify test data script (on the Python side) recently,
if not let me know and we can walk through it.

eriq-augustine (approved):
lgtm!

We'll merge once we get the Python PR ready.

Issue #147, opened 2025-02-10
JSON5 Support


eriq-augustine:
One of the most common errors that course builders have is that their JSON files are not properly formatted.
It's usually not something major, but something like a trailing comma.
Of course, [JSON5](https://en.wikipedia.org/wiki/JSON#JSON5) is a bit more flexible and could solve some of these issues.

Support for JSON5 would be nice, but we are also a bit limited because of how we unmarshal objects using encoding/json (which directly unmarshals into structs).
A potential way around this would be to read a file, parse as JSON5, convert to JSON, then unmarshal into a struct.
Since we mainly use our own JSON interfaces in internal/util/json.go, this would not be hard to do.
We should make sure a conversion like this does not have too large a performance impact.
We can also only check for JSON5 on specific files,
since we only allow that on files that come from users (as opposed to files that are in the DB, for example).

jayantpranjal0:
Hi @eriq-augustine I would like to work on this issue. Can you assign this to me?

eriq-augustine:
Hey @jayantpranjal0,

I don't assign issues until I see some work on them (otherwise it is quite common for young students to claim issues prematurely).

I don't have a specific library in mind, but am fairly picky about the libraries we bring in.
So you should make sure to take a good look around.

Good luck.

jayantpranjal0:
Yeah, sure. There doesn't seem to be an official implementation of json5 in golang. However, I found https://github.com/yosuke-furukawa/json5 which is still not complete implementation of json5, but still is used by google's chromium. 
Also, I went over some other alternatives for extension of json. I thin hjson and hujson are also good, whose implementation are also there for golang.
For hujson: https://github.com/tailscale/hujson
For hjson: https://github.com/hjson/hjson-go

PR #152, opened 2025-02-22
Added HuJson support

jayantpranjal0:
Fixes: #147

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/edulinq/autograder
 
-go 1.22.0
+go 1.23
 
-toolchain go1.23.1
+toolchain go1.23.2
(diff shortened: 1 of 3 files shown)


eriq-augustine (changes requested):
I am afk, so these comments will be terse and not include line-specific comments.

You should take another read of the issue.
We specifically talk about how we don't need to decide json5 (jwcc in this case), in all situations.

You are missing tests.

Why did you change the base go version?

How does this affect performance?
The issue explicitly mentions this.
Do you have any benchmarks?
They don't need to be fancy, but we should check.

eriq-augustine:
@jayantpranjal0 
Any progress on this?

PR #153, opened 2025-02-26
Added support for globs in the source of mv or cp file-ops.

Lucas-Ellenberger:
This PR adds support for globs during file-ops.

This change allows users to leverage the power of globs when to simplify file operations. Currently, only the source of either `move` or `copy` ops are supported. To ensure file ops remain with the local directory, we simplify the source operand before evaluating the glob. This behavior is noted in `types.md`.

This PR should be reviewed with the associated PR on the canonical python interface, which can be found [here](https://github.com/edulinq/autograder-py/pull/39).


docs/types.md
```diff
@@ -1045,6 +1045,23 @@ Below are the currently supported file operations:
 | `make-dir` | `mkdir`    | path                   | `mkdir -p`       | Create a directory at the given path. The path make already exist. Any required parent directories will be created. |
 | `remove`   | `rm`       | path                   | `rm -rf`         | Remove the given path. The path may be a file or directory. The path may not exist. |
 
+Any `move` or `copy` command may include globs in the source path.
```
eriq-augustine, line 1048: `remove` makes sense for glob support as well.

docs/types.md
```diff
@@ -1045,6 +1045,23 @@ Below are the currently supported file operations:
...
+We support the [Go standard for matching globs](https://pkg.go.dev/path/filepath#Match).
```
eriq-augustine, line 1049: s/matching globs/globs/

docs/types.md
eriq-augustine, line 1048: Add a link to the glob wiki page when you first mention them.

docs/types.md
```diff
@@ -1045,6 +1045,23 @@ Below are the currently supported file operations:
...
+We use lazy evaluation for globs, so we reduce the source path before expanding globs.
```
eriq-augustine, line 1050: The "lazy" part is not exactly correct, since they are cannot be evaluated when they are first parsed.
I would be more direct and say that paths with globs simplified as if the globs were normal names before they are expanded.

docs/types.md
```diff
@@ -1045,6 +1045,23 @@ Below are the currently supported file operations:
...
+When given the following example to move contents into a directory called `temp`:
```
eriq-augustine, line 1051: This wording is too flowery/indirect.
Just say something like:
"A glob path like ... will be simplified into ...".

internal/util/fileop.go
```diff
@@ -166,22 +166,50 @@ func (this *FileOperation) Exec(baseDir string) error {
 	parts := []string(*this)
 	command := parts[0]
 
+	var errs error = nil
 	if command == FILE_OP_LONG_COPY {
```
eriq-augustine, line 170: These cases are getting too large, move them to a function.

internal/util/fileop.go
```diff
@@ -166,22 +166,50 @@ func (this *FileOperation) Exec(baseDir string) error {
...
+		sourcePaths, err := prepForGlobs(sourcePathGlob, destPath)
+		if err != nil {
+			return fmt.Errorf("Failed to prep globs '%s' for a move: '%w'.", sourcePathGlob, err)
+		}
+
+		for _, sourcePath := range sourcePaths {
+			finalDestPath := destPath
+			if IsDir(destPath) {
+				finalDestPath = filepath.Join(destPath, filepath.Base(sourcePath))
+			}
```
eriq-augustine, line 204: This seems like something MoveDirent() should handle.

internal/util/fileop.go
```diff
@@ -166,22 +166,50 @@ func (this *FileOperation) Exec(baseDir string) error {
...
 		destPath := resolvePath(parts[2], baseDir, false)
-		if sourcePath == destPath {
-			return nil
+
+		sourcePaths, err := prepForGlobs(sourcePathGlob, destPath)
+		if err != nil {
+			return fmt.Errorf("Failed to prep globs '%s' for a move: '%w'.", sourcePathGlob, err)
+		}
+
+		for _, sourcePath := range sourcePaths {
```
eriq-augustine, line 200: (sourcePath == destPath) is not handled.
Make sure there is a test case for this as well.

internal/util/fileop.go
```diff
@@ -166,22 +166,50 @@ func (this *FileOperation) Exec(baseDir string) error {
...
+			finalDestPath := destPath
+			if IsDir(destPath) {
+				finalDestPath = filepath.Join(destPath, filepath.Base(sourcePath))
+			}
+
+			err = MoveDirent(sourcePath, finalDestPath)
+			if err != nil {
+				errs = errors.Join(errs, err)
+			}
 		}
```
eriq-augustine, line 210: With some of my above comments about the different cases, copy and move may be able to share some common code (with a passed in function to handle the actual file operation).

internal/util/fileop.go
```diff
@@ -234,3 +262,23 @@ func resolvePath(path string, baseDir string, forceUnix bool) string {
...
 	return path
 }
+
+func prepForGlobs(globPath string, destPath string) ([]string, error) {
+	sourcePaths, err := filepath.Glob(globPath)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to resolve globs '%s': '%w'.", globPath, err)
+	}
+
+	if sourcePaths == nil {
```
eriq-augustine, line 272: Use `(len(sourcePaths) == 0)` instead.
This is more flexible.

internal/util/path.go
```diff
@@ -108,6 +108,21 @@ func IsSymLink(path string) bool {
...
+		}
+	}
+
+	if !IsDir(path) {
+		return fmt.Errorf("Path ('%s') does not exist or is not a dir.", path)
+	}
+
+	return nil
+}
+
```
eriq-augustine, line 125: Doesn't MkDir already effectively do this?

internal/util/fileop_test.go
```diff
@@ -217,6 +263,11 @@ func TestFileOpValidateBase(test *testing.T) {
 			nil,
 			"cannot point just to the current directory",
 		},
+		{
+			NewFileOperation([]string{"copy", "*/..", "b"}),
+			nil,
+			"cannot point just to the current directory",
+		},
 	}
```
eriq-augustine, line 271: What about malformed glob patterns?

internal/util/fileop.go
eriq-augustine, line 1: You're missing validation checks for the glob patterns.
See how FileSpec does it:
https://github.com/edulinq/autograder-server/blob/main/internal/util/filespec.go#L87

internal/util/fileop_test.go
```diff
@@ -326,33 +377,86 @@ func TestFileOpCopyBase(test *testing.T) {
...
+			"does not exist or is not a dir",
+		},
 	}
 
 	for i, testCase := range testCases {
 		rawOperation := []string{"cp", testCase.source, testCase.dest}
 
 		runFileOpExecTest(test, fmt.Sprintf("Case %d", i), rawOperation, testCase.errorSubstring, func(tempDir string) {
-			expectedSource := filepath.Join(tempDir, testCase.source)
+			expectedSourceGlob := filepath.Join(tempDir, testCase.source)
```
eriq-augustine, line 442: This is getting pretty complex.
Maybe the globs should have their own test function or you should change this one to have a []string for the expected dirents.

eriq-augustine (changes requested):
A pretty good start.

internal/util/dirent.go
```diff
@@ -20,6 +20,10 @@ func RemoveDirent(path string) error {
 	return err
 }
 
+func CopyDirentDefault(source string, dest string) error {
```
eriq-augustine, line 23: I would flip this and instead use `CopyDirent(source, dest)` and `CopyDirentFull(source, dest, onlyContents)`.
This would be nice, since most people won't have to think about onlyContents.

internal/util/dirent.go
```diff
@@ -48,6 +52,10 @@ func CopyLink(source string, dest string) error {
 		dest = filepath.Join(dest, filepath.Base(source))
 	}
 
+	if source == dest {
```
eriq-augustine, line 55: These need to be normalized before they are compared.
I would just do ShouldAbs() to normalize them (filepath.Abs() calls filepath.Clean()).
Note that we can abs here because we are just about to do the operation on them
(whereas FileOps and FileSpecs have to hang around for a while until they can be evaluated).

internal/util/fileop.go
```diff
@@ -102,6 +102,11 @@ func (this *FileOperation) Validate() error {
 			return fmt.Errorf("Argument at index %d ('%s') cannot point just to the current directory. File operation paths must point to a dirent inside the current directory tree.", i, parts[i])
 		}
 
+		_, err := filepath.Match(path, "")
```
eriq-augustine, line 105: A non-glob (e.g., dest) can have a path that is valid but looks like an invalid glob pattern.

internal/util/fileop.go
```diff
@@ -234,3 +233,67 @@ func resolvePath(path string, baseDir string, forceUnix bool) string {
 
 	return path
 }
+
+func handleGlobFileOperation(sourceGlob string, dest string, operation func(string, string) error) error {
+	sourcePaths, err := prepForGlobs(sourceGlob, dest)
+	if err != nil {
+		return fmt.Errorf("Failed to prep globs '%s' for a move: '%w'.", sourceGlob, err)
```
eriq-augustine, line 240: "move" is incorrect for a copy.

internal/util/fileop.go
```diff
@@ -234,3 +233,67 @@ func resolvePath(path string, baseDir string, forceUnix bool) string {
...
+
+func prepForGlobs(globPath string, destPath string) ([]string, error) {
+	sourcePaths, err := filepath.Glob(globPath)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to resolve globs '%s': '%w'.", globPath, err)
+	}
+
+	if len(sourcePaths) == 0 {
+		return nil, fmt.Errorf("Unable to find source path: '%s'.", globPath)
+	}
```
eriq-augustine, line 285: Remove doesn't have a similar error condition (which makes sense, `rm -rf` doesn't complain if nothing is there).

What about copy and move.
I could see it going either way.
I would think about it, make sure the code matches, and update the docs to match.

internal/util/fileop.go
```diff
@@ -234,3 +233,67 @@ func resolvePath(path string, baseDir string, forceUnix bool) string {
...
+
+	if len(sourcePaths) > 1 {
```
eriq-augustine, line 287: This guard seems unnecessary, given the return above.

internal/util/fileop.go
```diff
@@ -234,3 +233,67 @@ func resolvePath(path string, baseDir string, forceUnix bool) string {
...
+		err = MkDir(destPath)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to create dest dir '%s': '%w'.", destPath, err)
+		}
+
+		if !IsDir(destPath) {
```
eriq-augustine, line 293: How could this happen given the line above?
Doesn't MkDir() error if a non-dir is there?
And if it doesn't we can just beef-up MkDir().

internal/util/fileop_test.go
```diff
@@ -217,6 +263,16 @@ func TestFileOpValidateBase(test *testing.T) {
 			nil,
 			"cannot point just to the current directory",
 		},
+		{
+			NewFileOperation([]string{"copy", "*/..", "b"}),
+			nil,
+			"cannot point just to the current directory",
+		},
+		{
+			NewFileOperation([]string{"copy", "*/[a-z", "b"}),
```
eriq-augustine, line 272: Add a test case where this invalid pattern is used as a dest (and is therefore okay).
Currently this will fail, but we want it to pass
(which makes it a nice test).

internal/util/fileop_test.go
```diff
@@ -358,6 +414,185 @@ func TestFileOpCopyBase(test *testing.T) {
 	}
 }
 
+func TestFileOpCopyGlob(test *testing.T) {
+	testCases := []struct {
+		source          string
+		dest            string
+		expectedSources []string
+		expectedDests   []string
```
eriq-augustine, line 422: I could be missing something, but I'm not sure why there are two here.
Are the expected sources and dests treated differently, or can we just combine them?

If this is for symmetry with the move test, maybe just have expectedExists and expectedNotExists slices.

eriq-augustine (changes requested):
Looking pretty good.
Some small things and test things.

internal/util/fileop.go
Lucas-Ellenberger: For a source path of length 1, the destination path is allowed to be a file.

This guard is here to make sure a source glob that matches multiple files are not clobbering each other over a single file destination (by forcing them to target a directory).

eriq-augustine: oic, makes snese.

internal/util/fileop.go
```diff
@@ -105,6 +112,13 @@ func (this *FileOperation) Validate() error {
 		parts[i] = path
 	}
 
+	if supportsSourceGlobs[command] {
+		_, err := filepath.Match(parts[1], "")
+		if err != nil {
+			return fmt.Errorf("Argument at index 1 ('%s') contains an invalid path pattern: '%w'.", parts[1], err)
```
eriq-augustine, line 118: s/path pattern/glob pattern/

internal/util/fileop.go
```diff
@@ -234,3 +242,63 @@ func resolvePath(path string, baseDir string, forceUnix bool) string {
 
 	return path
 }
+
+func handleGlobFileOperation(sourceGlob string, dest string, operation func(string, string) error) error {
+	sourcePaths, err := prepForGlobs(sourceGlob, dest)
+	if err != nil {
+		return fmt.Errorf("Failed to prep globs '%s' for an operation: '%w'.", sourceGlob, err)
```
eriq-augustine, line 249: s/globs/glob path/
or
s/globs/glob pattern/

I would pick one and generally apply it throughout this PR.

internal/util/fileop.go
```diff
@@ -234,3 +242,63 @@ func resolvePath(path string, baseDir string, forceUnix bool) string {
...
+	sourcePaths, err := filepath.Glob(globPath)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to resolve globs '%s': '%w'.", globPath, err)
+	}
+
+	if len(sourcePaths) == 0 {
+		return nil, fmt.Errorf("Unable to find source path: '%s'.", globPath)
+	}
+
+	if len(sourcePaths) > 1 {
```
eriq-augustine, line 296: Put a small comment here describing why a dir is made.

internal/util/fileop_test.go
```diff
@@ -358,6 +424,154 @@ func TestFileOpCopyBase(test *testing.T) {
...
+			[]string{
+				alreadyExistsFileRelpath,
+				alreadyExistsFileAltRelpath,
+				filepath.Join(alreadyExistsDirname, "*.txt", alreadyExistsFilename),
+				filepath.Join(alreadyExistsDirname, "*.txt", alreadyExistsFilenameAlt),
+			},
+			"",
+		},
+		{
+			"*",
```
eriq-augustine, line 535: Add a similar case to this.
You have `copy * a`,
add `copy * <alreadyExistsDirname>`.
This will specifically test ignoring a self copy when doing globs.

Make a move version of this as well.

docs/types.md
```diff
@@ -1045,6 +1045,13 @@ Below are the currently supported file operations:
 | `make-dir` | `mkdir`    | path                   | `mkdir -p`       | Create a directory at the given path. The path make already exist. Any required parent directories will be created. |
 | `remove`   | `rm`       | path                   | `rm -rf`         | Remove the given path. The path may be a file or directory. The path may not exist. |
```
eriq-augustine, line 1046: What do you think about changing the relevant arguments here from "path" to "glob path" or "glob pattern"?

It may be a little more confusing for those that don't know what globs are, but your next paragraph should clear it up.

I have a comment further down that talks about the choice between "glob pattern" and "glob path".

docs/types.md
```diff
@@ -1045,6 +1045,13 @@ Below are the currently supported file operations:
...
 
+Any `move`, `copy`, or `remove` command may include [globs](https://en.wikipedia.org/wiki/Glob_%28programming%29) in the source path.
+We support the [Go standard for globs](https://pkg.go.dev/path/filepath#Match).
+All paths are simplified before globs are expanded.
+A glob path like `*/*/..` will be simplified into `*`.
```
eriq-augustine, line 1051: I would join this into the previous sentence (so it's really clear that they are linked).

eriq-augustine (changes requested):
Code's looking real good.
Only some small wording and testing changes.

eriq-augustine (approved):
lgtm

Issue #137, opened 2025-01-17
Send Course Emails


eriq-augustine:
We have all the infrastructure necessary to send emails to students/roles from the server,
e.g. if an instructor wants to remind all the students about an assignment deadline.

We need to decide if it should be graders or admins with this ability.

Something to think about that will not be part of the base functionality (or this issue) is having advanced queries for the email recipients,
e.g., "send to all students who have not submitted this assignment"
or "send to all students who scored below this value on this assignment".
Using `--dry-run` you should be able to see who would be emailed without actually emailing them.

eriq-augustine:
We actually now have the functionality for sending to course roles and excluding specific users from the recipients:
https://github.com/edulinq/autograder-server/blob/main/docs/types.md#course-email-specification-courseemailspec

eriq-augustine:
Complete by @jkelman1324 in 0d4e11443ad7f4451b42523e74bdde39839701a4.

PR #154, opened 2025-03-02
Issue #137 Send Course Emails

jkelman1324:


internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+		// Dry run.
+		{
+			UserEmail: "course-grader",
+			Message: email.Message{
+				To:      []string{"*"},
+				Subject: "Email for All Course Users",
+			},
+			DryRun: true,
+		},
```
eriq-augustine, line 126: Missing some test cases.
E.g.,
 - To: course-student (role), -course-student@test.edulinq.org
 - CC
 - BCC
 - Body
 - HTML

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,166 @@
...
+		if !response.Success {
+			if testCase.PermError {
+				if response.Locator != testCase.Locator {
+					test.Errorf("Case %d: Incorrect error returned. Expected '%s', found '%s'.",
+						i, testCase.Locator, response.Locator)
+				}
+			} else if testCase.Message.Subject == "" {
+				continue
+			} else if (len(testCase.Message.To) + len(testCase.Message.CC) + len(testCase.Message.BCC)) == 0 {
+				continue
```
eriq-augustine, line 147: KISS

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,166 @@
...
+			continue
+		}
+
+		if testCase.PermError {
+			test.Errorf("Case %d: Did not get an expected permissions error.", i)
+			continue
+		}
+
+		if testCase.DryRun {
+			if email.GetTestMessages() != nil {
```
eriq-augustine, line 161: Check len() == 0 instead.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,166 @@
...
+				test.Errorf("Case %d: Emails were sent when they should not have.", i)
```
eriq-augustine, line 162: Missing a continue after.
It is not strictly necessary, but it's nice to put here so we don't accidentally leave it off when/if we add more checks.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,166 @@
...
+			}
+		}
```
eriq-augustine, line 164: Missing any checks on the response.

You don't need to check the full contents of the To/CC/BCC (since that would require more fields in the test case that I don't think are necessary),
instead just check that the total number of recipients is > 0.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,166 @@
...
+	}
```
eriq-augustine, line 165: Missing checks against actual emails that were sent (or captured in unit testing).

internal/api/courses/admin/routes.go
```diff
@@ -8,6 +8,7 @@ import (
 
 var routes []core.Route = []core.Route{
 	core.MustNewAPIRoute(`courses/admin/update`, HandleUpdate),
+	core.MustNewAPIRoute(`courses/admin/email`, HandleEmail),
```
eriq-augustine, line 11: Put these endpoints in lexicographic order.

eriq-augustine (changes requested):
Looking pretty good.
There are definitely things to go through, but the core is looking pretty strong.

internal/api/courses/admin/email.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, core.NewBadRequestError("-627", &request.APIRequest, "No email subject provided.")
+	}
+
+	var errs error
+
+	request.To, err = db.ResolveCourseUsers(request.Course, request.To)
+	if err != nil {
+		err = fmt.Errorf("Failed to resolve 'to' email addresses.")
+		errs = errors.Join(errs, err)
+	}
```
eriq-augustine, line 42: You missed the point of an errors.Join() a bit.
You don't need to make a new error or have a conditional when using it.

If you really want the specific error message you can use the conditionals,
but then specific internal errors would make more sense.

internal/api/courses/admin/email.go
```diff
@@ -0,0 +1,76 @@
...
+
+	request.BCC, err = db.ResolveCourseUsers(request.Course, request.BCC)
+	if err != nil {
+		err = fmt.Errorf("Failed to resolve 'bcc' email addresses.")
+		errs = errors.Join(errs, err)
+	}
+
+	if errs != nil {
+		return nil, core.NewInternalError("-628", &request.APIRequestCourseUserContext, errs.Error())
```
eriq-augustine, line 57: Generally, we want to attach errors with Err() instated of calling Error() when forming an APIError.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+	for i, testCase := range testCases {
+		email.ClearTestMessages()
+
+		var fields map[string]any
+		util.MustJSONFromString(util.MustToJSON(testCase.Message), &fields)
+		fields["dry-run"] = testCase.DryRun
+
+		response := core.SendTestAPIRequestFull(test, "courses/admin/email", fields, nil, testCase.UserEmail)
+
```
eriq-augustine, line 158: Remove this newline.
Since we are checking the result of a call (like when we check for an error),
we generally want the error checking code right up against what may produce the error.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,199 @@
...
+		if !response.Success {
+			if response.Locator != testCase.Locator {
+				test.Errorf("Case %d: Incorrect error returned. Expected '%s', found '%s'.",
+					i, testCase.Locator, response.Locator)
+			}
+
+			continue
+		}
+
```
eriq-augustine, line 167: Missing a check for a locator when there is no error.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,199 @@
...
+		var responseContent EmailResponse
+		util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
+
+		if (len(responseContent.To) + len(responseContent.CC) + len(responseContent.BCC)) == 0 {
+			test.Errorf("Case %d: Email was sent without any recipients.", i)
+			continue
+		}
+
+		if testCase.Message.Body != email.GetTestMessages()[0].Body {
```
eriq-augustine, line 189: Just fetch the message once instead of calling GetTestMessage() several times.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,199 @@
...
+				test.Errorf("Case %d: Email was sent when it should not have.", i)
+			}
+
+			continue
+		} else {
+			if len(email.GetTestMessages()) == 0 {
+				test.Errorf("Case %d: Email was not sent when it should have.", i)
+				continue
+			}
+		}
```
eriq-augustine, line 179: Instead of these nested if's, I would prefer if you check check for dry run (which will return) and then (in a separate if) check for this.

eriq-augustine (changes requested):
This is improving.
The error handling and testing needs some more work.

internal/api/courses/admin/email.go
```diff
@@ -0,0 +1,66 @@
...
+	errs = errors.Join(errs, err)
+
+	request.CC, err = db.ResolveCourseUsers(request.Course, request.CC)
+	errs = errors.Join(errs, err)
+
+	request.BCC, err = db.ResolveCourseUsers(request.Course, request.BCC)
+	errs = errors.Join(errs, err)
+
+	if errs != nil {
+		return nil, core.NewInternalError("-628", &request.APIRequestCourseUserContext, "Failed to resolve course users.").Err(errs)
```
eriq-augustine, line 47: This error message is a bit off.
An end user does not know what you mean by "course users".
You can say something they would understand like "email recipients".

internal/api/courses/admin/email.go
```diff
@@ -0,0 +1,66 @@
...
+type EmailResponse struct {
+	To  []string `json:"to"`
+	CC  []string `json:"cc"`
+	BCC []string `json:"bcc"`
+}
+
+// Send an email to course users.
+func HandleEmail(request *EmailRequest) (*EmailResponse, *core.APIError) {
+	response := EmailResponse{}
+	var err error
```
eriq-augustine, line 29: Move this definition to right above errs (since you don't use it until then).

internal/api/courses/admin/email.go
```diff
@@ -0,0 +1,66 @@
...
+
+type EmailResponse struct {
+	To  []string `json:"to"`
+	CC  []string `json:"cc"`
+	BCC []string `json:"bcc"`
+}
+
+// Send an email to course users.
+func HandleEmail(request *EmailRequest) (*EmailResponse, *core.APIError) {
+	response := EmailResponse{}
```
eriq-augustine, line 28: Since you don't use this until then end, I would just define it as a literal and set To, CC, and BCC at the same time.
(Don't do it in the return, that looks messy with multiple returns.)

Generally, we want to declare variable right before they are used.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,205 @@
...
+			if response.Locator != testCase.Locator {
+				test.Errorf("Case %d: Incorrect error returned. Expected '%s', found '%s'.",
+					i, testCase.Locator, response.Locator)
+			}
+
+			continue
+		}
+
+		if response.Locator != "" {
+			test.Errorf("Case %d: Unexpected error. Expected '%s', found '%s'.", i, testCase.Locator, response.Locator)
```
eriq-augustine, line 168: This is a not clear to the reader.
You don't expected an error and response.Locator is empty,
so drop response.Locator from your message.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,205 @@
...
+			continue
+		}
+
+		testMessages := email.GetTestMessages()
+		if testCase.DryRun {
+			if len(testMessages) != 0 {
+				test.Errorf("Case %d: Email was sent when it should not have.", i)
```
eriq-augustine, line 175: Put in the error how many emails were found.
You don't need the messages themselves, but since the above guard checks the length, you may as well include it in the output.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+
+			continue
+		}
+
+		if len(testMessages) == 0 {
```
eriq-augustine, line 181: What about if 2 emails were sent?
I would change this to `!= 1` and update the message accordingly.

internal/api/courses/admin/email_test.go
```diff
@@ -0,0 +1,205 @@
...
+			test.Errorf("Case %d: Email was not sent when it should have.", i)
+			continue
+		}
+
+		var responseContent EmailResponse
+		util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
+		testMessage := testMessages[0]
```
eriq-augustine, line 188: I don't this this line should be in this vertical block.
Lines 186 and 187 are all about decoding the response,
but then you just pull out the test message here.

We want vertical block to reflect one "action" or "thing",
so it makes it easier for people reading your code.

eriq-augustine (changes requested):
Getting pretty close, just some small things.

eriq-augustine (approved):
lgtm!

We'll merge this in after the Python side gets approved.

PR #155, opened 2025-03-05
Added metric aggregation.

OliverLok:


internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/edulinq/autograder/internal/log"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type AggregationQuery struct {
+	EnableAggregation bool     `json:"enable-aggregation,omitempty"`
```
eriq-augustine, line 14: I'm pretty hesitant about this field.
What if the other fields are non-empty, but this is false?
Can't we check check if the other fields are empty?

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+	GroupByFields     []string `json:"group-by,omitempty"`
+	AggregateField    string   `json:"aggregate,omitempty"`
+}
+
+type QueryResponse struct {
+	Response []map[string]any `json:"response"`
+}
```
eriq-augustine, line 21: Seems like either a strange location or name for this struct.
This file should generally just be about aggregation.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+
+const (
```
eriq-augustine, line 23: Generally, consts before types.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+	COUNT_KEY  = "count"
+	MAX_KEY    = "max"
+	MEAN_KEY   = "mean"
+	MEDIAN_KEY = "median"
+	MIN_KEY    = "min"
```
eriq-augustine, line 28: These seem strange.
Why are we pulling the values out by key instead of just putting the result (util.AggregateValues) directly in our result?

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+		jsonTag := util.JSONFieldName(field)
+
+		if jsonTag != "" {
+			jsonTags[jsonTag] = true
+		}
+	}
+}
+
+// Ensure the groupByFields and aggregateField exist in the metricType.
+func validate(metricType any, groupByFields []string, aggregateField string) error {
```
eriq-augustine, line 79: Make a more specific name.
Validate functions are typically methods, so they don't need more specific names.
But you are using a function.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+const (
+	COUNT_KEY  = "count"
+	MAX_KEY    = "max"
+	MEAN_KEY   = "mean"
+	MEDIAN_KEY = "median"
+	MIN_KEY    = "min"
+)
+
+// Aggregate metrics based on groupByFields and the aggregateField.
+func ApplyAggregation(metrics []map[string]any, metricType any, groupByFields []string, aggregateField string) ([]map[string]any, error) {
```
eriq-augustine, line 32: Using `metrics` and `metricType` here seems a bit strange to me.
You may be converting to JSON to early.
If you instead passed in a []any or []Metric, then you wouldn't have to have two fields here.

eriq-augustine, line 32: Why is AggregationQuery not passed in?
Instead you pass its core component in.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+
+		if result != nil {
+			aggregatedResults = append(aggregatedResults, result)
+		}
+	}
+
+	return aggregatedResults, nil
+}
+
+func extractJSONTags(reflectType reflect.Type, jsonTags map[string]bool) {
```
eriq-augustine, line 61: I would take a minute and think about if you really want to do this reflexively instead of via a JSON conversion.
You already did a JSON conversion to generalize the input to aggregation,
so then mixing reflection seems a bit janky.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+		if !hasAllNeededFields(metric, groupByFields, aggregateField) {
+			continue
+		}
+
+		var groupFieldValues []string
+		for _, field := range groupByFields {
+			groupFieldValues = append(groupFieldValues, fmt.Sprintf("%v", metric[field]))
+		}
+
+		bucketKey := strings.Join(groupFieldValues, "|")
```
eriq-augustine, line 113: Use a less likely value (like "::") and make it a constant.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+	}
+
+	if totalCount == 0 {
+		return nil, nil
+	}
+
+	aggregatedResults[COUNT_KEY] = totalCount
+
+	// Only add additional stats if aggregation was done on numeric values.
```
eriq-augustine, line 181: Just use different functions for string and float aggregation.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+// Calculate aggregations for a group of metrics.
+func computeGroupAggregation(groupedMetrics []map[string]any, aggregateField string, groupByFields []string) (map[string]any, error) {
+	if len(groupedMetrics) == 0 {
+		return nil, nil
+	}
+
+	aggregatedResults := make(map[string]any)
+
+	// Use the first metric in the group to extract the group-by field values.
+	firstMetric := groupedMetrics[0]
```
eriq-augustine, line 145: Put a newline after this.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, field := range groupByFields {
+		value, exists := firstMetric[field]
+		if !exists {
```
eriq-augustine, line 148: Is this possible?
Didn't you ensure this earlier?

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		totalCount++
+
+		aggregateValueString := fmt.Sprintf("%v", aggregateValue)
+		numericValue, err := util.StrToFloat(aggregateValueString)
+		// Only append to values if the aggregateValue was a number.
+		if err == nil {
+			values = append(values, numericValue)
+		}
```
eriq-augustine, line 172: I don't like this conversion to a string and back.
Instead, I would just track for each bucket if it has a string value (thereby forcing all values to be strings).

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+		aggregatedResults[MEAN_KEY] = stats.Mean
+		aggregatedResults[MEDIAN_KEY] = stats.Median
+		aggregatedResults[MIN_KEY] = stats.Min
+		aggregatedResults[MAX_KEY] = stats.Max
+	}
+
+	return aggregatedResults, nil
+}
+
+func QuerySortFuncForTesting(recordA, recordB any) int {
```
eriq-augustine, line 194: Why put this outside of a test file?

eriq-augustine, line 194: We do not share type declarations in this project.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,212 @@
...
+	jsonA, err := json.Marshal(recordA)
```
eriq-augustine, line 195: Why direct marshals instead of one of our util functions?

internal/stats/apirequest.go
```diff
@@ -22,12 +22,28 @@ type APIRequestMetric struct {
...
+	Locator      string `json:"include-locator,omitempty"`
+}
+
+type ExcludeAPIRequestMetricField struct {
+	Sender       string `json:"exclude-sender,omitempty"`
+	Endpoint     string `json:"exclude-endpoint,omitempty"`
+	UserEmail    string `json:"exclude-user,omitempty"`
+	CourseID     string `json:"exclude-course,omitempty"`
+	AssignmentID string `json:"exclude-assignment,omitempty"`
+	Locator      string `json:"exclude-locator,omitempty"`
```
eriq-augustine, line 46: This seems bananas to me.

Why do we need include/exclude?

And if we do need it, why not just use strings like `GroupByFields`?

internal/util/json.go
```diff
@@ -83,6 +84,30 @@ func JSONMapFromString(data string) (map[string]any, error) {
...
+func MustToJSONMapSlice(data any) []map[string]any {
+	result, err := ToJsonMapSlice(data)
+	if err != nil {
+		log.Fatal("Failed to convert object to a slice of maps.", log.NewAttr("data", data), err)
+	}
+
+	return result
+}
+
+func ToJsonMapSlice(data any) ([]map[string]any, error) {
```
eriq-augustine, line 96: I'm not a fan of these.

Something like MustToGenericJSON() makes sense to me since we pass in a slice and expect a slice.
Here we don't even know we have a slice/list coming in.

internal/util/json.go
```diff
@@ -239,3 +264,12 @@ func MustFormatJSONObjectIndent(text string) string {
 	MustJSONFromString(text, &object)
 	return MustToJSONIndent(object)
 }
+
+func MustToGenericJSON(input []any, sortFunc func(a any, b any) int) []map[string]any {
```
eriq-augustine, line 268: The name should indicate a slice/list.

eriq-augustine (changes requested):
There are still some core concepts (like having the aggregation work for all the queries) that you have still not addressed.

internal/stats/aggregate.go
OliverLok: So it doesn't have to get redefined in each package that uses it (other packages can't access this exported function if it's in a test file).

eriq-augustine: A couple things about this:
1) The name of this function is off,
    this function is not a sorting function it's a comparison function.
2) This function has no record-specific functionality, so I would just put this in util/json.go.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,279 @@
...
+		for field, value := range metric {
+			stringMetric[field] = fmt.Sprintf("%v", value)
+		}
+
+		var groupFieldValues []string
+		for _, field := range query.GroupByFields {
+			groupFieldValues = append(groupFieldValues, fmt.Sprintf("%v", metric[field]))
+		}
+
+		bucketKey := strings.Join(groupFieldValues, "::")
```
eriq-augustine, line 168: Make the separator a constant.

eriq-augustine, line 168: Make the separator a constant.

internal/stats/aggregate.go
```diff
@@ -0,0 +1,279 @@
...
+	for aggregateKey, aggregateValue := range statsJSONMap {
+		statsMap[aggregateKey] = aggregateValue
+	}
+
+	aggregateMap[STATS] = statsMap
+
+	return aggregateMap, nil
+}
+
+func aggregateNonNumeric(groupedMetrics []map[string]string, aggregateMap map[string]any) (map[string]any, error) {
```
eriq-augustine, line 272: This should call into util, so we can expand it in the future (e.g., most common string, length stats, etc).

internal/stats/aggregate.go
```diff
@@ -0,0 +1,279 @@
...
+
+		// Turn all values into strings for aggregation.
+		stringMetric := make(map[string]string, len(metric))
+		for field, value := range metric {
+			stringMetric[field] = fmt.Sprintf("%v", value)
+		}
+
+		var groupFieldValues []string
+		for _, field := range query.GroupByFields {
+			groupFieldValues = append(groupFieldValues, fmt.Sprintf("%v", metric[field]))
```
eriq-augustine, line 165: I am not a fan of this string conversion, since then you have to convert it back to a number of numeric aggregation.
String conversions can be costly.
Can this just be kept as an any?

internal/stats/aggregate.go
```diff
@@ -0,0 +1,279 @@
...
+
+	aggregatedResults[GROUP_BY] = groupByValues
+	aggregatedResults[OVERVIEW] = query.OverviewField
+
+	return aggregatedResults, nil
+}
+
+func canAggregateAsNumeric(metric map[string]string, overviewField string) bool {
+	aggregateValue := metric[overviewField]
+	_, err := util.StrToFloat(aggregateValue)
```
eriq-augustine, line 229: This is a bit dangerous, since somethings can be converted to numbers but are supposed to be treated as strings (e.g. locators).
I made a comment earlier about keeping the value as its true type instead of converting to a string.
That would solve this.

internal/stats/course.go
```diff
@@ -33,16 +34,6 @@ type CourseMetric struct {
 	Value uint64 `json:"duration"`
 }
 
-type CourseMetricQuery struct {
```
eriq-augustine, line 36: Since you are making simplification to the query like this,
would it make sense to also simplify the metric to just use `BaseMetric.Attributes` instead of fields.
Then you wouldn't have to deal with matching JSON fields.

Note that this is not an order or strong stance.
This is something for you to really think about and decide if it would be better for your code.

PR #156, opened 2025-03-07
Added support for hard fail and skipped questions during grading.

Lucas-Ellenberger:
This PR adds support for the `hard_fail` and `skipped` fields in a `GradedQuestion`.

These fields allows graders to better communicate results to users by allowing them to indicate either major failures during grading or a question's grading was skipped.

For example, this is already implemented for the [python grader](https://github.com/edulinq/autograder-py/commit/b343d24e5ba9cb56b0d00384999c081ef39f8ac3) provided by  `autograder-py`.


eriq-augustine (approved):
lgtm!

PR #159, opened 2025-03-25
Added the fields of API requests and responses when describing API endpoints.

Lucas-Ellenberger:
This PR adds functionality to display the expected input and output fields when describing the API endpoints on the server.
This update allows interfaces to the `autograder-server` to pre-validate API requests.

As complex custom types cannot be easily displayed in the `endpoints` portion of `resources/api.json`, this PR adds a new section `types` to better display complex types.

While there are many possibilities to expand the current work (show required fields, default values, etc.), this PR serves as a solid base.


internal/util/connection.go
```diff
@@ -28,13 +30,23 @@ func ReadFromNetworkConnection(connection net.Conn) ([]byte, error) {
...
 
-	jsonBuffer := make([]byte, size)
-	_, err = connection.Read(jsonBuffer)
-	if err != nil {
-		return nil, err
+	log.Trace("Reading the following bytes from a connection...", log.NewAttr("expected bytes", size))
+
+	buffer := make([]byte, size)
+	var numBytesRead uint64
+	for numBytesRead = 0; numBytesRead < size; {
```
eriq-augustine, line 37: Comment on what you are doing here and why it is necessary.
(Doesn't need to be anything fancy, just 1-2 sentences.)

internal/util/connection.go
```diff
@@ -28,13 +30,23 @@ func ReadFromNetworkConnection(connection net.Conn) ([]byte, error) {
...
+		currBuffer := buffer[numBytesRead:]
+		currBytesRead, err := connection.Read(currBuffer)
+		if err != nil {
+			return nil, err
```
eriq-augustine, line 41: Wrap this error and include how many bytes have been read so far.

internal/util/connection.go
```diff
@@ -28,13 +30,23 @@ func ReadFromNetworkConnection(connection net.Conn) ([]byte, error) {
...
+		}
+
+		numBytesRead += uint64(currBytesRead)
 	}
```
eriq-augustine, line 46: What if numBytesRead > size?

internal/util/connection.go
```diff
@@ -28,13 +30,23 @@ func ReadFromNetworkConnection(connection net.Conn) ([]byte, error) {
 		return nil, fmt.Errorf("Message content is too large to read.")
 	}
 
-	jsonBuffer := make([]byte, size)
-	_, err = connection.Read(jsonBuffer)
-	if err != nil {
-		return nil, err
+	log.Trace("Reading the following bytes from a connection...", log.NewAttr("expected bytes", size))
+
+	buffer := make([]byte, size)
```
eriq-augustine, line 35: Consider s/size/expectedSize/

internal/util/connection.go
```diff
@@ -56,7 +68,10 @@ func WriteToNetworkConnection(connection net.Conn, data []byte) error {
 
 	responseBuffer.Write(data)
 
-	_, err = connection.Write(responseBuffer.Bytes())
+	log.Trace("Writing the following bytes to a connection...", log.NewAttr("expected bytes", size))
+
+	numBytesWritten, err := connection.Write(responseBuffer.Bytes())
+	log.Trace("Wrote the following bytes to a connection.", log.NewAttr("actual bytes", numBytesWritten))
```
eriq-augustine, line 74: In non-debugging, we generally don't want to get between a statement and it's error check.

internal/util/connection.go
```diff
@@ -56,7 +68,10 @@ func WriteToNetworkConnection(connection net.Conn, data []byte) error {
 
 	responseBuffer.Write(data)
 
-	_, err = connection.Write(responseBuffer.Bytes())
+	log.Trace("Writing the following bytes to a connection...", log.NewAttr("expected bytes", size))
+
+	numBytesWritten, err := connection.Write(responseBuffer.Bytes())
```
eriq-augustine, line 73: Comment on the non-intuitive behavior we see.

eriq-augustine (changes requested):
The output is looking real good.

I just have a few small comments (that just surpassed the level of a nit).

eriq-augustine (approved):
lgtm!

PR #160, opened 2025-03-26
Generlized course and API metrics.

OliverLok:
This PR ties into #155

### DIFF
--- a/internal/analysis/common.go
+++ b/internal/analysis/common.go
@@ -154,18 +154,16 @@ func collectAnalysisStats(fullSubmissionIDs []string, totalRunTime int64, initia
 	now := timestamp.Now()
 
 	for courseID, assignmentID := range seenIdentifiers {
-		metric := stats.CourseMetric{
-			BaseMetric: stats.BaseMetric{
-				Timestamp: now,
-				Attributes: map[string]any{
-					stats.ATTRIBUTE_KEY_ANALYSIS: analysisType,
-				},
+		metric := stats.BaseMetric{
+			Timestamp: now,
+			Attributes: map[string]any{
+				stats.ATTRIBUTE_KEY_ANALYSIS: analysisType,
+				stats.TYPE:                   stats.CourseMetricTypeCodeAnalysisTime,
+				stats.COURSE_ID:              courseID,
+				stats.ASSIGNMENT_ID:          assignmentID,
+				stats.USER_EMAIL:             initiatorEmail,
+				stats.VALUE:                  uint64(totalRunTime),
 			},
-			Type:         stats.CourseMetricTypeCodeAnalysisTime,
-			CourseID:     courseID,
-			AssignmentID: assignmentID,
-			UserEmail:    initiatorEmail,
-			Value:        uint64(totalRunTime),
 		}
 
 		stats.AsyncStoreCourseMetric(&metric)
--- a/internal/api/courses/stats/query.go
+++ b/internal/api/courses/stats/query.go
@@ -10,26 +10,30 @@ type QueryRequest struct {
 	core.APIRequestCourseUserContext
 	core.MinCourseRoleAdmin
 
-	stats.CourseMetricQuery
+	stats.MetricQuery
 }
 
 type QueryResponse struct {
-	Records []*stats.CourseMetric `json:"results"`
+	Records []*stats.BaseMetric `json:"results"`
 }
 
 // Query metrics for a specific course.
 // Only the context course can be queried for, the target-course field will be ignored for this endpoint.
 func HandleQuery(request *QueryRequest) (*QueryResponse, *core.APIError) {
-	// The request must be for the given course.
-	request.CourseMetricQuery.CourseID = request.Course.ID
+	if request.Where == nil {
+		request.Where = make(map[string]any)
+	}
 
-	records, err := db.GetCourseMetrics(request.CourseMetricQuery)
+	// The request must be for the given course.
+	request.Where[stats.COURSE_ID] = request.Course.ID
+
+	records, err := db.GetCourseMetrics(request.MetricQuery)
 	if err != nil {
 		return nil, core.NewInternalError("-618", &request.APIRequestCourseUserContext, "Failed to query course stats.").Err(err)
 	}
 
 	response := QueryResponse{
-		Records: stats.ApplyBaseQuery(records, request.CourseMetricQuery.BaseQuery),
+		Records: stats.ApplyBaseQuery(records, request.MetricQuery.BaseQuery),
 	}
 
 	return &response, nil
--- a/internal/api/courses/stats/query_test.go
+++ b/internal/api/courses/stats/query_test.go
@@ -21,19 +21,95 @@ func TestQuery(test *testing.T) {
 		expectedValues   []int
 	}{
 		// Base
-		{"server-admin", "", stats.MetricQuery{}, []int{100, 200, 300}},
-		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{Sort: 1}}, []int{300, 200, 100}},
-		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{After: timestamp.FromMSecs(150)}}, []int{200, 300}},
+		{
+			"server-admin",
+			"",
+			stats.MetricQuery{},
+			[]int{100, 200, 300},
+		},
+		{
+			"server-admin",
+			"",
+			stats.MetricQuery{
+				BaseQuery: stats.BaseQuery{
+					Sort: 1,
+				},
+			},
+			[]int{300, 200, 100}},
+		{
+			"server-admin",
+			"",
+			stats.MetricQuery{
+				BaseQuery: stats.BaseQuery{
+					After: timestamp.FromMSecs(150),
+				},
+			},
+			[]int{200, 300},
+		},
 
 		// Course Specific
-		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{Where: map[string]any{stats.ASSIGNMENT_ID: "A2"}}}, []int{200}},
-		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{Where: map[string]any{stats.ASSIGNMENT_ID: "ZZZ"}}}, nil},
-		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{Where: map[string]any{stats.USER_EMAIL: "U1"}}}, []int{100, 200}},
-		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{Where: map[string]any{stats.TYPE: stats.CourseMetricTypeGradingTime}}}, []int{100, 300}},
+		{
+			"server-admin",
+			"",
+			stats.MetricQuery{
+				BaseQuery: stats.BaseQuery{
+					Where: map[string]any{
+						stats.ASSIGNMENT_ID: "A2",
+					},
+				},
+			},
+			[]int{200},
+		},
+		{
+			"server-admin",
+			"",
+			stats.MetricQuery{
+				BaseQuery: stats.BaseQuery{
+					Where: map[string]any{
+						stats.ASSIGNMENT_ID: "ZZZ",
+					},
+				},
+			},
+			nil,
+		},
+		{
+			"server-admin",
+			"",
+			stats.MetricQuery{
+				BaseQuery: stats.BaseQuery{
+					Where: map[string]any{
+						stats.USER_EMAIL: "U1",
+					},
+				},
+			},
+			[]int{100, 200},
+		},
+		{
+			"server-admin",
+			"",
+			stats.MetricQuery{
+				BaseQuery: stats.BaseQuery{
+					Where: map[string]any{
+						stats.TYPE: stats.CourseMetricTypeGradingTime,
+					},
+				},
+			},
+			[]int{100, 300},
+		},
 
 		// Error
-		{"server-user", "-040", stats.MetricQuery{}, nil},
-		{"course-student", "-020", stats.MetricQuery{}, nil},
+		{
+			"server-user",
+			"-040",
+			stats.MetricQuery{},
+			nil,
+		},
+		{
+			"course-student",
+			"-020",
+			stats.MetricQuery{},
+			nil,
+		},
 	}
 
 	for _, record := range testRecords {
--- a/internal/api/courses/stats/query_test.go
+++ b/internal/api/courses/stats/query_test.go
@@ -17,23 +17,23 @@ func TestQuery(test *testing.T) {
 	testCases := []struct {
 		email            string
 		permErrorLocator string
-		query            stats.CourseMetricQuery
+		query            stats.MetricQuery
 		expectedValues   []int
 	}{
 		// Base
-		{"server-admin", "", stats.CourseMetricQuery{}, []int{100, 200, 300}},
-		{"server-admin", "", stats.CourseMetricQuery{BaseQuery: stats.BaseQuery{Sort: 1}}, []int{300, 200, 100}},
-		{"server-admin", "", stats.CourseMetricQuery{BaseQuery: stats.BaseQuery{After: timestamp.FromMSecs(150)}}, []int{200, 300}},
+		{"server-admin", "", stats.MetricQuery{}, []int{100, 200, 300}},
+		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{Sort: 1}}, []int{300, 200, 100}},
+		{"server-admin", "", stats.MetricQuery{BaseQuery: stats.BaseQuery{After: timestamp.FromMSecs(150)}}, []int{200, 300}},
 
 		// Course Specific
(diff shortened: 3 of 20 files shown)


internal/stats/query.go
```diff
@@ -31,6 +33,41 @@ type BaseQuery struct {
...
+type MetricQuery struct {
+	BaseQuery
+}
+
+func (this MetricQuery) Match(metric *BaseMetric) bool {
+	metricJSONMap, err := util.ToJSONMap(metric)
+	if err != nil {
+		log.Error("Failed to convert metric to a JSON map.", err)
+		return false
+	}
```
eriq-augustine, line 52: Why convert the entire thing to a general map?
The attributes are the only thing with unsure structure.

internal/analysis/common.go
```diff
@@ -154,18 +154,16 @@ func collectAnalysisStats(fullSubmissionIDs []string, totalRunTime int64, initia
...
-				},
+		metric := stats.BaseMetric{
+			Timestamp: now,
+			Attributes: map[string]any{
+				stats.ATTRIBUTE_KEY_ANALYSIS: analysisType,
+				stats.TYPE_KEY:               stats.CourseMetricTypeCodeAnalysisTime,
+				stats.COURSE_ID_KEY:          courseID,
+				stats.ASSIGNMENT_ID_KEY:      assignmentID,
+				stats.USER_EMAIL_KEY:         initiatorEmail,
+				stats.VALUE_KEY:              uint64(totalRunTime),
```
eriq-augustine, line 165: You BaseMetric needs work.

Why would `stats.TYPE_KEY` and `stats.VALUE_KEY` be entries in the map?
Wouldn't these be in every matric, and therefore be first-class fields?

eriq-augustine, line 165: Should the values be unit64 or float64?

internal/stats/course.go
```diff
@@ -19,30 +19,14 @@ const (
 const (
```
eriq-augustine, line 19: Is CourseMetricType needed anymore?
Its original point is that it is an enum and can be used as a type.
If the values are only going to be used in a general map, then there is not point in having a specific type for it.
If so, then they should just be constants instead of enums.

internal/stats/course.go
```diff
@@ -143,31 +79,53 @@ func AsyncStoreCourseMetric(metric *CourseMetric) {
...
+		attributes[USER_EMAIL_KEY] = userEmail
+	}
+
+	if courseID != "" {
+		attributes[COURSE_ID_KEY] = courseID
+	}
+
+	if assignmentID != "" {
+		attributes[ASSIGNMENT_ID_KEY] = assignmentID
+	}
```
eriq-augustine, line 97: I've seen this almost exact code elsewhere.
Maybe you can abstract it out to a constructor for BaseMetric.

internal/stats/query.go
```diff
@@ -31,6 +33,41 @@ type BaseQuery struct {
...
+	// Keys are field names (e.g., "course") and values are what to include (e.g., course101).
+	// This filter is applied after all other BaseQuery conditions are applied.
+	Where map[string]any `json:"where,omitempty"`
+}
+
+type MetricQuery struct {
+	BaseQuery
+}
+
+func (this MetricQuery) Match(metric *BaseMetric) bool {
```
eriq-augustine, line 47: You should probably match the BaseQuery first.

internal/stats/query.go
```diff
@@ -31,6 +33,41 @@ type BaseQuery struct {
...
 	Sort int `json:"sort"`
+
+	// Filter results to only include metrics that match BaseMetric attribute field values.
+	// Keys are field names (e.g., "course") and values are what to include (e.g., course101).
+	// This filter is applied after all other BaseQuery conditions are applied.
+	Where map[string]any `json:"where,omitempty"`
+}
+
+type MetricQuery struct {
+	BaseQuery
```
eriq-augustine, line 44: Do we even need a base query anymore?
Now that the metrics are all the same, can we just have query?

eriq-augustine (changes requested):
A pretty good start.

I think a bit more design could be put into the base metric,
and I think there are more simplification opportunities with the base query.

internal/stats/metrics.go
```diff
@@ -1,29 +1,126 @@
...
+	USER_EMAIL_KEY    MetricAttribute = "user"
+)
+
+// Values for the type field inside of Metric and Query.
+const (
+	API_REQUEST_STATS_TYPE        MetricType = "api-request-stats"
+	CODE_ANALYSIS_TIME_STATS_TYPE MetricType = "code-analysis-time-stats"
+	GRADING_TIME_STATS_TYPE       MetricType = "grading-time-stats"
+	TASK_TIME_STATS_TYPE          MetricType = "task-time-stats"
+)
```
eriq-augustine, line 30: Since you have known values for both of these, you could use a Validate() function for both Metric and Query.
There you can do things like check for correct keys and empty values.

For a metric, it can be called before storing it.
And for a query, it can be called when received from the user.

internal/api/courses/stats/query.go
```diff
@@ -10,26 +10,30 @@ type QueryRequest struct {
...
-	Records []*stats.CourseMetric `json:"results"`
+	Records []*stats.Metric `json:"results"`
 }
 
 // Query metrics for a specific course.
 // Only the context course can be queried for, the target-course field will be ignored for this endpoint.
 func HandleQuery(request *QueryRequest) (*QueryResponse, *core.APIError) {
+	if request.Where == nil {
+		request.Where = make(map[stats.MetricAttribute]any)
+	}
```
eriq-augustine, line 25: This is something that should probably be done in a Validate() method.

internal/api/core/routing.go
```diff
@@ -152,7 +152,22 @@ func sendAPIResponse(apiRequest ValidAPIRequest, response http.ResponseWriter,
...
+		Attributes: map[stats.MetricAttribute]any{
+			stats.SENDER_KEY:   sender,
+			stats.ENDPOINT_KEY: endpoint,
+		},
+	}
+
+	stats.InsertIntoMapIfPresent(metric.Attributes, stats.USER_EMAIL_KEY, userEmail)
+	stats.InsertIntoMapIfPresent(metric.Attributes, stats.COURSE_ID_KEY, courseID)
+	stats.InsertIntoMapIfPresent(metric.Attributes, stats.ASSIGNMENT_ID_KEY, assignmentID)
+	stats.InsertIntoMapIfPresent(metric.Attributes, stats.LOCATOR_KEY, locator)
```
eriq-augustine, line 168: Consider having methods to set each of these values so you don't have to supply the key each time.
E.g., `metric.SetCourse(courseID)`.

Note that this is a soft suggestion ("consider").
You have a better mental image of the code, so think about it.

internal/db/disk/stats.go
```diff
@@ -8,86 +8,64 @@ import (
...
+	if err != nil {
+		return nil, err
 	}
 
-	path := this.getCourseStatsPath(query.CourseID)
-
-	this.contextReadLock(path)
-	defer this.contextReadUnlock(path)
+	this.statsLock.Lock()
+	defer this.statsLock.Unlock()
```
eriq-augustine, line 40: I think you are misunderstanding the purpose of this lock,
it is being used incorrectly in two different ways.

1) This should be a read locking.
The idea behind this lock is that it signals to the backend that an operation is in progress and the backend can't close until we are done (hence the write Lock() in Close()).

2) You need to acquire a path-specific lock.
Since this lock is not about protecting the resources (it is about just telling the backend an operation is happening), you need another lock to protect the file you are working with.
More specifically, you need a lock for this file specifically (i.e., a contextual lock from the lockmanager).

internal/db/disk/stats.go
```diff
@@ -8,86 +8,64 @@ import (
...
-
-	this.apiRequestLock.Lock()
-	defer this.apiRequestLock.Unlock()
+func (this *backend) getStatsPath(metricType stats.MetricType) (string, error) {
+	if metricType == "" {
+		return "", fmt.Errorf("No metric type was given.")
+	}
 
-	return util.AppendJSONLFile(path, record)
+	return filepath.Join(this.baseDir, string(metricType)+".jsonl"), nil
```
eriq-augustine, line 66: Ensure this path is normalized (you will need it for locking).

internal/stats/metrics.go
```diff
@@ -1,29 +1,126 @@
...
+
+	assignmentID, ok := this.Attributes[ASSIGNMENT_ID_KEY].(string)
+	if ok {
+		attrs = append(attrs, log.NewAssignmentAttr(assignmentID))
+	}
+
+	userEmail, ok := this.Attributes[USER_EMAIL_KEY].(string)
+	if ok {
+		attrs = append(attrs, log.NewUserAttr(userEmail))
+	}
```
eriq-augustine, line 93: What about just sending all the attributes?

eriq-augustine (changes requested):
Looking better.

There are still some semantic around metrics and queries that need to be worked out.

eriq-augustine (changes requested):
Then generalized metrics and queries are looking pretty good.

internal/stats/metrics.go
```diff
@@ -1,29 +1,169 @@
...
+	User_Email_Key                               = "user"
+)
+
+// Values for the type field inside of Metric and Query.
+const (
+	Unknown_Metric_Attribute_Type MetricType = ""
+	API_Request_Stats_Type                   = "api-request"
+	Code_Analysis_Time_Stats_Type            = "code-analysis-time"
+	Grading_Time_Stats_Type                  = "grading-time"
+	Task_Time_Stats_Type                     = "task-time"
```
eriq-augustine, line 33: We don't use Title_Case for pretty much anything.
Use PascalCase like the user roles:
https://github.com/edulinq/autograder-server/blob/main/internal/model/role_server.go#L17

internal/stats/metrics.go
```diff
@@ -1,29 +1,169 @@
...
+	this.addAttr(&attrs, Analysis_Type_Key)
+	this.addAttr(&attrs, Course_ID_Key)
+	this.addAttr(&attrs, Endpoint_Key)
+	this.addAttr(&attrs, Locator_Key)
+	this.addAttr(&attrs, Sender_Key)
+	this.addAttr(&attrs, Task_Type_Key)
+	this.addAttr(&attrs, User_Email_Key)
+
+	return attrs
+}
```
eriq-augustine, line 79: This seems strange to me.
Why not just iterate over `this.Attributes`?

internal/stats/metrics.go
```diff
@@ -1,29 +1,169 @@
...
+
+func (this *Metric) Validate() error {
+	if this == nil {
+		return fmt.Errorf("No metric was given.")
+	}
+
+	if this.Type == Unknown_Metric_Attribute_Type {
+		return fmt.Errorf("Metric attribute was not set.")
+	}
```
eriq-augustine, line 88: You should also verify that it is a known type.

With enums, it is not unusual to have a map around that can quickly be used to see if the value is known.
For example:
https://github.com/edulinq/autograder-server/blob/main/internal/model/role_server.go#L25
If you don't have another purpose for having the map (like the roles do),
then you can just have the value be `true`.

internal/stats/metrics.go
```diff
@@ -1,29 +1,169 @@
...
+
+	if this.Attributes == nil {
+		this.Attributes = make(map[MetricAttribute]any)
+	}
+
+	for key, value := range this.Attributes {
+		if key == Unknown_Metric_Attribute_Key {
+			return fmt.Errorf("Metric attribute key was empty.")
+		}
+
```
eriq-augustine, line 102: Same with the type, you should probably verify that you know what the attribute is.

internal/stats/metrics_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func TestMetricValidationBase(test *testing.T) {
+	defer clearBackend()
+
+	testCases := []struct {
+		metric         *Metric
+		errorSubstring string
+	}{
+		{
+			errorSubstring: "No metric was given.",
```
eriq-augustine, line 20: Explicitly set the metric to nil so everyone can easily see it.

internal/stats/metrics_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+func runStoreStatsTests(test *testing.T, metric Metric) {
+	clearBackend()
+	defer clearBackend()
+
+	if backend != nil {
+		test.Fatalf("Stats backend should not be set during testing.")
+	}
+
+	typedBackend := makeTestBackend()
```
eriq-augustine, line 138: This isn't really testing anything but the test backend.
You should probably use the real Db for this test
(which means you would have to do it in the db package).

internal/stats/query.go
```diff
@@ -31,14 +27,72 @@ type BaseQuery struct {
...
+
+	for field, value := range this.Where {
+		if field == Unknown_Metric_Attribute_Key {
+			return fmt.Errorf("Query attribute field was empty.")
+		}
+
+		if value == nil || value == "" {
+			return fmt.Errorf("Query attribute value was empty.")
+		}
+	}
```
eriq-augustine, line 62: The Type and Where checks look the same between metrics and queries.
I would make a common validation function for these.

internal/stats/query.go
```diff
@@ -31,14 +27,72 @@ type BaseQuery struct {
...
+
+	if this.Type != metric.Type {
+		return false
+	}
+
+	for key, value := range this.Where {
+		if value == "" {
+			return false
+		}
```
eriq-augustine, line 79: This case is not possible on a validated query.

internal/stats/query_test.go
```diff
@@ -3,143 +3,213 @@ package stats
...
-		baseQuery   BaseQuery
-		expected    []BaseMetric
+		query          *Query
+		errorSubstring string
 	}{
 		{
-			simpleBaseMetrics,
-			BaseQuery{},
-			simpleBaseMetrics,
+			errorSubstring: "No query was given.",
```
eriq-augustine, line 26: Same as for metric, make this an explicit nil.

internal/api/courses/stats/query_test.go
```diff
@@ -10,34 +10,113 @@ import (
...
+			},
+		},
+
+		// Code Analysis Time Stats, Error
+		{
+			email:           "course-user",
+			expectedLocator: "-017",
+			query: &stats.Query{
+				Type: stats.Task_Time_Stats_Type,
+			},
```
eriq-augustine, line 114: These cases look strange.
Why -017 (no password)?
Are you trying for a perm error?
If so, why bother with the different query types?

The "Error" in the comment is not useful.
What type of error?

internal/db/stats_test.go
```diff
@@ -45,70 +46,210 @@ func (this *DBTests) DBTestStoreSystemStats(test *testing.T) {
...
 
-	err := StoreAPIRequestMetric(&testRecord)
+	query := stats.Query{
+		Type: stats.Code_Analysis_Time_Stats_Type,
+	}
+
+	runStoreStatsTests(test, &metric, query)
+}
+
+func (this *DBTests) DBTestStoreMetricFailures(test *testing.T) {
```
eriq-augustine, line 140: These tests are strange.

The first one is not a "failure" as the test name implies.
It's just a non-matching query.

Then since you are focused on the storing, I don't think you should specify the query in the test.
You can just make a query that matches the type of the metric you are storing.

eriq-augustine (changes requested):
The core is looking good.

Some tests and little things need some work.

internal/stats/metrics.go
```diff
@@ -1,29 +1,200 @@
...
+	return nil
+}
+
+func validateType(metricType MetricType) error {
+	if metricType == UnknownMetricAttributeType {
+		return fmt.Errorf("Type was not set.")
+	}
+
+	if !knownMetricTypes[metricType] {
+		return fmt.Errorf("Invalid metric type: '%v'.", metricType)
```
eriq-augustine, line 136: s/Invalid/Unknown/

This is just a little more specific.

internal/stats/metrics.go
```diff
@@ -1,29 +1,200 @@
...
+	}
+
+	if config.UNIT_TESTING_MODE.Get() {
+		storeFunc()
+	} else {
+		go storeFunc()
+	}
+}
+
+func (this *Metric) setAttrIfNotEmpty(key MetricAttribute, value string) {
```
eriq-augustine, line 164: s/Attr/Attribute/

Since the field uses the fill name, the method should as well.

internal/grader/grade.go
```diff
@@ -112,8 +112,21 @@ func Grade(ctx context.Context, assignment *model.Assignment, submissionPath str
...
+		Type:      stats.GradingTimeStatsType,
+		Value:     float64((endTimestamp - startTimestamp).ToMSecs()),
+		Attributes: map[stats.MetricAttribute]any{
+			stats.CourseIDKey: gradingInfo.CourseID,
+		},
+	}
+
+	// Add optional fields if non-empty.
+	metric.SetUserEmail(gradingInfo.User)
+	metric.SetAssignmentID(gradingInfo.AssignmentID)
```
eriq-augustine, line 126: How would these be empty?

eriq-augustine (changes requested):
Mainly good, but there are still some issues to deal with.

PR #163, opened 2025-03-29
Added an API endpoint to make proxy submissions.

Lucas-Ellenberger:
Added a new API endpoint for course admins (or above) to make proxy submissions on behalf of other course users.

This feature allows for course admin to make a proxy submission at a "fudged" submission time. If a proxy time is not specified, one will be automatically assigned at the assignment's due date. If there is not a due date, the proxy time will default to the Unix Epoch.

By doing so, this endpoint gives course admin to (effectively) remove the late penalty for a student's submission. As these submissions are made by course admin, they are not subject to rejection.


internal/api/courses/assignments/submissions/proxy/submit.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		stdout := ""
+		stderr := ""
+
+		if (result != nil) && (result.HasTextOutput()) {
+			stdout = result.Stdout
+			stderr = result.Stderr
+		}
+
+		log.LogToSplitLevels(log.LevelDebug, log.LevelInfo, "Submission failed internally.", err, request.Assignment, log.NewAttr("stdout", stdout), log.NewAttr("stderr", stderr), request.User)
```
eriq-augustine, line 66: I wouldn't bother splitting the logs in this file.
(I don't think the normal submit should split its logs either.)

internal/grader/grade.go
```diff
@@ -99,8 +103,17 @@ func Grade(ctx context.Context, assignment *model.Assignment, submissionPath str
...
-	gradingInfo.GradingStartTime = startTimestamp
-	gradingInfo.GradingEndTime = endTimestamp
+	gradingInfo.ProxyUser = options.ProxyUser
+
+	if options.ProxyTime == nil {
+		gradingInfo.GradingStartTime = startTimestamp
+		gradingInfo.GradingEndTime = endTimestamp
+	} else {
+		gradingInfo.GradingStartTime = *options.ProxyTime
+		gradingInfo.GradingEndTime = *options.ProxyTime
```
eriq-augustine, line 113: Should we make this ProxyTime + (endTimestamp - startTimestamp)?

internal/api/courses/assignments/submissions/proxy/submit.go
```diff
@@ -0,0 +1,91 @@
...
+
+		return &response, nil
+	}
+
+	// A proxy submission should never be rejected.
```
eriq-augustine, line 71: You say this, but I don't see any special handling for rejection.

internal/api/courses/assignments/submissions/proxy/submit.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/grader"
+	"github.com/edulinq/autograder/internal/log"
+	"github.com/edulinq/autograder/internal/model"
+	"github.com/edulinq/autograder/internal/timestamp"
+)
+
+type SubmitRequest struct {
+	core.APIRequestAssignmentContext
+	core.MinCourseRoleAdmin
```
eriq-augustine, line 13: Graders can edit grades, so they should probably be allowed to proxy.

Lucas-Ellenberger:
As the DB semantics around GetNextSubmissionID() are based around the current time (and double check that they are the largest), proxy submissions can be made with any fudged time and will still be considered "most recent". These semantics allow a proxy submission to have a "proxy time" that is earlier than other submission times.

I decided to default the proxy time to a day before the assignment due date. While this may lead to multiple assignment's having the same fudged start and end time, the submission IDs will still be unique and favor the most recent API call. I want to include a default time that is prior to the assignment due date, as I imagine a common use case for this endpoint is to forgive a student that submitted late (so the user will want to easily make a submission before the due date).

internal/api/courses/assignments/submissions/proxy/submit.go
```diff
@@ -0,0 +1,85 @@
...
+// Proxy submit an assignment submission to the autograder.
+func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
+	response := SubmitResponse{}
+
+	if !request.ProxyUser.Found {
+		return &response, nil
+	}
+
+	// Proxy submissions are not subject to submission restrictions.
+	gradeOptions := grader.GradeOptions{
```
eriq-augustine, line 41: Instead of constructing yourself,
you should probably get the default options and add your new options to it.

internal/api/courses/assignments/submissions/proxy/submit.go
```diff
@@ -0,0 +1,85 @@
...
+	if reject != nil {
+		log.Error("Proxy submission rejected.", request.Assignment, log.NewAttr("reason", reject.String()), log.NewAttr("request", request), request.User)
+
+		response.Rejected = true
+		response.Message = reject.String()
+		return &response, nil
+	}
+
+	if failureMessage != "" {
+		log.LogToSplitLevels(log.LevelTrace, log.LevelDebug, "Submission got a soft error.", request.Assignment, log.NewAttr("message", failureMessage), log.NewAttr("request", request), request.User)
```
eriq-augustine, line 75: I wouldn't split this.
Just send it to debug.

internal/api/courses/assignments/submissions/proxy/submit_test.go
```diff
@@ -0,0 +1,250 @@
...
+	defer db.ResetForTesting()
+
+	testSubmissions, err := grader.GetTestSubmissions(config.GetTestdataDir(), !config.DOCKER_DISABLE.Get())
+	if err != nil {
+		test.Fatalf("Failed to get test submissions in '%s': '%v'.", config.GetTestdataDir(), err)
+	}
+
+	testSubmission := testSubmissions[0]
+
+	dueDate := timestamp.Timestamp(123456)
```
eriq-augustine, line 38: Make the time a little more obvious, like 1 or -1.

internal/api/courses/assignments/submissions/proxy/submit_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+	testSubmissions, err := grader.GetTestSubmissions(config.GetTestdataDir(), !config.DOCKER_DISABLE.Get())
+	if err != nil {
+		test.Fatalf("Failed to get test submissions in '%s': '%v'.", config.GetTestdataDir(), err)
+	}
+
+	verifySuccessfulTestSubmissions(test, testSubmissions, nil)
+}
+
+func TestProxySubmitAtDueDate(test *testing.T) {
```
eriq-augustine, line 27: Add a simple comment (1-2 sentences) on these tests saying what you are doing and what should happen.

(Especially since the bulk of the test is handled by another function, it can be unclear the case you are testing.)

internal/api/courses/assignments/submissions/submit.go
```diff
@@ -41,21 +41,21 @@ func HandleSubmit(request *SubmitRequest) (*SubmitResponse, *core.APIError) {
 			stderr = result.Stderr
 		}
 
-		log.LogToSplitLevels(log.LevelDebug, log.LevelInfo, "Submission failed internally.", err, request.Assignment, log.NewAttr("stdout", stdout), log.NewAttr("stderr", stderr), request.User)
+		log.Info("Submission failed internally.", err, request.Assignment, log.NewAttr("stdout", stdout), log.NewAttr("stderr", stderr), request.User)
```
eriq-augustine, line 44: Bring up to warn.

internal/db/disk/submission.go
```diff
@@ -69,8 +70,27 @@ func (this *backend) SaveSubmissions(course *model.Course, submissions []*model.
 	return this.saveSubmissionsLock(course, submissions, true)
 }
 
+// Compute the next submission ID based on the current time.
```
eriq-augustine, line 73: Mention what happens on a tie.

internal/db/disk/submission.go
```diff
@@ -69,8 +70,27 @@ func (this *backend) SaveSubmissions(course *model.Course, submissions []*model.
...
+// The next submission ID will be the largest ID for the user.
```
eriq-augustine, line 74: Not just a user, but user/assignment pair.

internal/db/disk/submission.go
```diff
@@ -69,8 +70,27 @@ func (this *backend) SaveSubmissions(course *model.Course, submissions []*model.
...
+		previousSubmissionID, err := strconv.Atoi(previousStringID)
+		if err != nil {
+			return "", err
+		}
+
+		// Ensure the next submission ID is the largest.
+		if int64(previousSubmissionID) >= submissionID {
+			submissionID = int64(previousSubmissionID) + 1
+		}
+	}
```
eriq-augustine, line 92: Doesn't the user context lock during grading and a monotomic clock ensure that the current ID is at least tied with the largest?
https://github.com/edulinq/autograder-server/blob/main/internal/grader/grade.go#L64

Is this necessary?

internal/db/disk/submission.go
```diff
@@ -316,7 +336,8 @@ func (this *backend) getUserSubmissionDir(courseID string, assignmentID string,
 	return filepath.Join(this.getCourseDirFromID(courseID), model.SUBMISSIONS_DIRNAME, assignmentID, user)
 }
 
-// Get the short id of the most recent submission (or empty string if there are no submissions).
+// Get the short ID of the most recent submission (or empty string if there are no submissions).
+// The most recent submission has the largest ID for the user.
```
eriq-augustine, line 340: The semantics for IDs needs to be laid out at the database level (right now you are at the database backend level).
For example, if we have another database backend it needs to follow the same rules.

Note that a well-placed comment can be sufficient for this
(if it is not something that can be easily enforced with code).

internal/grader/grade.go
```diff
@@ -118,6 +133,25 @@ func Grade(ctx context.Context, assignment *model.Assignment, submissionPath str
 	return &gradingResult, nil, "", nil
 }
 
+// Resolve a proxy time for a given assignment.
+// If the proxy time is not specified, fall back to a day before the assignment due date.
+// If the assignment does not have a due date, the current time will be used for the proxy time.
+func ResolveProxyTime(proxyTime *timestamp.Timestamp, assignment *model.Assignment) *timestamp.Timestamp {
+	if proxyTime == nil {
```
eriq-augustine, line 140: Invert this conditional.

internal/grader/grade.go
```diff
@@ -118,6 +133,25 @@ func Grade(ctx context.Context, assignment *model.Assignment, submissionPath str
...
+		if assignment.DueDate == nil {
+			now := timestamp.Now()
+			proxyTime = &now
+		} else {
+			// Convert time (one day) to hours -> minutes -> seconds -> msecs.
+			oneDay := int64(1 * 24 * 60 * 60 * 1000)
```
eriq-augustine, line 146: I think using the time package here would be more clear:
`time.Duration(24 * time.Hour).Milliseconds()`.

internal/grader/grade_test.go
```diff
@@ -50,7 +50,8 @@ func runSubmissionTests(test *testing.T, parallel bool, useDocker bool) {
 	}
 
 	gradeOptions := GradeOptions{
-		NoDocker: !useDocker,
+		NoDocker:       !useDocker,
+		CheckRejection: false,
```
eriq-augustine, line 54: I would just leave it out (since it is default and it was left out originally).

eriq-augustine (changes requested):
Looking pretty good.

> I decided to default the proxy time to a day before the assignment due date. While this may lead to multiple assignment's having the same fudged start and end time, the submission IDs will still be unique and favor the most recent API call. I want to include a default time that is prior to the assignment due date, as I imagine a common use case for this endpoint is to forgive a student that submitted late (so the user will want to easily make a submission before the due date).

I'm not a fan of this default.
What if the submission period is less than a day (like for a quiz)?
What if the assignment is not yet late (like if a student just doesn't have access to submissions for some reason (firewall?)?

Instead, I suggest using the current time if it is not late, and 1 minute before the due date if it is late.
And since multiple submissions can be at the same time (as you explained about the DB semantics),
then we will still be fine.

eriq-augustine (approved):
lgtm!

I'll hold until the Python side.

PR #169, opened 2025-04-05
Added an API endpoint to make proxy resubmissions.

Lucas-Ellenberger:
Added a new API endpoint for course graders (or above) to make proxy resubmissions on behalf of other course users. Proxy resubmissions target a user's previous submission ID.

This feature allows for course graders to make a proxy resubmission at a "fudged" submission time. If a proxy time is not specified, one will be automatically at the earlier between now and one minute before the assignment's due date.

By doing so, this endpoint gives course graders an effective way to resubmit a user's submission without the late penalty. As these submissions are made by course graders, they are not subject to rejection.

This PR should be reviewed with the corresponding [python interface PR](https://github.com/edulinq/autograder-py/pull/48).

Edit: Linked python interface PR.

### DIFF
--- a/internal/api/core/submission.go
+++ b/internal/api/core/submission.go
@@ -1,0 +1,73 @@
+package core
+
+import (
+	"context"
+
+	"github.com/edulinq/autograder/internal/grader"
+	"github.com/edulinq/autograder/internal/log"
+	"github.com/edulinq/autograder/internal/model"
+)
+
+type GradeableRequest interface {
+	GetContext() context.Context
+	GetAssignment() *model.Assignment
+	GetUserEmail() string
+	GetMessage() string
+	ToLogAttrs() []any
+}
+type BaseSubmitResponse struct {
+	Rejected bool   `json:"rejected"`
+	Message  string `json:"message"`
+
+	GradingSuccess bool               `json:"grading-success"`
+	GradingInfo    *model.GradingInfo `json:"result"`
+}
+
+func GradeToSubmissionResponse(request GradeableRequest, submissionPath string, options grader.GradeOptions) BaseSubmitResponse {
+	response := BaseSubmitResponse{}
+
+	result, reject, failureMessage, err := grader.Grade(
+		request.GetContext(),
+		request.GetAssignment(),
+		submissionPath,
+		request.GetUserEmail(),
+		request.GetMessage(),
+		options,
+	)
+	if err != nil {
+		stdout := ""
+		stderr := ""
+
+		if (result != nil) && (result.HasTextOutput()) {
+			stdout = result.Stdout
+			stderr = result.Stderr
+		}
+
+		logAttributes := append([]any{err, log.NewAttr("stdout", stdout), log.NewAttr("stderr", stderr)}, request.ToLogAttrs()...)
+		log.Warn("Submission failed internally.", logAttributes)
+
+		return response
+	}
+
+	if reject != nil {
+		logAttributes := append([]any{log.NewAttr("reason", reject.String())}, request.ToLogAttrs()...)
+		log.Debug("Submission rejected.", logAttributes)
+
+		response.Rejected = true
+		response.Message = reject.String()
+		return response
+	}
+
+	if failureMessage != "" {
+		logAttributes := append([]any{log.NewAttr("message", failureMessage)}, request.ToLogAttrs()...)
+		log.Debug("Submission got a soft error.", logAttributes)
+
+		response.Message = failureMessage
+		return response
+	}
+
+	response.GradingSuccess = true
+	response.GradingInfo = result.Info
+
+	return response
+}
--- a/internal/api/core/submission.go
+++ b/internal/api/core/submission.go
@@ -14,7 +14,7 @@ type BaseSubmitResponse struct {
 	GradingInfo    *model.GradingInfo `json:"result"`
 }
 
-func GradeToSubmissionResponse(request APIRequestAssignmentContext, submissionPath string, email string, message string, options grader.GradeOptions) BaseSubmitResponse {
+func GradeRequestSubmission(request APIRequestAssignmentContext, submissionPath string, email string, message string, options grader.GradeOptions) BaseSubmitResponse {
 	response := BaseSubmitResponse{}
 
 	result, reject, failureMessage, err := grader.Grade(
--- a/internal/api/core/submission.go
+++ b/internal/api/core/submission.go
@@ -1,20 +1,11 @@
 package core
 
 import (
-	"context"
-
 	"github.com/edulinq/autograder/internal/grader"
 	"github.com/edulinq/autograder/internal/log"
 	"github.com/edulinq/autograder/internal/model"
 )
 
-type GradeableRequest interface {
-	GetContext() context.Context
-	GetAssignment() *model.Assignment
-	GetUserEmail() string
-	GetMessage() string
-	ToLogAttrs() []any
-}
 type BaseSubmitResponse struct {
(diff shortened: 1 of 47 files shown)


internal/api/courses/assignments/submissions/proxy/resubmit.go
```diff
@@ -0,0 +1,113 @@
...
+	tempDir, err := util.MkDirTemp("resumbit-request-files-")
+	if err != nil {
+		return nil, core.NewInternalError("-633", &request.APIRequestCourseUserContext, "Failed to create temp resubmit files directory.").
+			Err(err).Assignment(request.Assignment.GetID()).
+			Add("target-user", request.ProxyUser.Email).Add("submission", request.TargetSubmission)
+	}
+
+	err = util.GzipBytesToDirectory(tempDir, gradingResult.InputFilesGZip)
+	if err != nil {
+		util.RemoveDirent(tempDir)
```
eriq-augustine, line 66: Instead of removing here, defer the remove on 63 (right after the error check for its creation).

(Of course, temp dirs get automatically removed, but we like to keep it clean (since the automatic removal is not until shutdown).)

internal/api/courses/assignments/submissions/proxy/resubmit.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if failureMessage != "" {
+		log.Debug("Resubmission got a soft error.", request.Assignment, log.NewAttr("message", failureMessage), log.NewAttr("request", request), request.User)
+
+		response.Message = failureMessage
+		return &response, nil
+	}
+
+	response.GradingSuccess = true
+	response.GradingInfo = result.Info
```
eriq-augustine, line 110: This is the third place we have code similar to this.
See if it can be easily combined.
(Note that I am not looking into the other endpoints right now, so there could be some issue that makes them hard to combine.)

It may also be possible to combine most of the outputs (response fields) into a single struct that can be embedded and then used for this common function.

It would also not be a problem if all the log messages look similar (because the request will get logged with a unique ID).

internal/api/courses/assignments/submissions/proxy/resubmit.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		stdout := ""
+		stderr := ""
+
+		if (result != nil) && (result.HasTextOutput()) {
+			stdout = result.Stdout
+			stderr = result.Stderr
+		}
+
+		log.Warn("Resubmission failed internally.", err, request.Assignment, log.NewAttr("stdout", stdout), log.NewAttr("stderr", stderr), request.User)
```
eriq-augustine, line 88: Instead of pulling out request attributes, you should be able to use getLogAttributesFromAPIRequest():
 - https://github.com/edulinq/autograder-server/blob/main/internal/api/core/request.go#L222
 - https://github.com/edulinq/autograder-server/blob/main/internal/api/core/routing.go#L113

eriq-augustine (changes requested):
Looking pretty good.

internal/api/core/submission.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/edulinq/autograder/internal/model"
+)
+
+type GradeableRequest interface {
+	GetContext() context.Context
+	GetAssignment() *model.Assignment
+	GetUserEmail() string
+	GetMessage() string
+	ToLogAttrs() []any
+}
```
eriq-augustine, line 17: Missing newline after.

eriq-augustine, line 17: This seems like a lot for what is just an `core.APIRequestAssignmentContext` and message.
Why can't GradeToSubmissionResponse() just take those two instead of the interface?

internal/api/courses/assignments/submissions/proxy/resubmit.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (request ResubmitRequest) ToLogAttrs() []any {
+	return core.GetLogAttributesFromAPIRequest(&request)
+}
+
+type ResubmitResponse struct {
+	FoundUser       bool `json:"found-user"`
+	FoundSubmission bool `json:"found-submission"`
+
+	core.BaseSubmitResponse
```
eriq-augustine, line 49: We usually put embedded types first.
(Not always, but it's usually nice in smaller structs like this.)

eriq-augustine (changes requested):
Looking pretty good, but I think the abstraction can be a bit simpler.

internal/api/core/submission.go
```diff
@@ -0,0 +1,64 @@
...
+
+type BaseSubmitResponse struct {
+	Rejected bool   `json:"rejected"`
+	Message  string `json:"message"`
+
+	GradingSuccess bool               `json:"grading-success"`
+	GradingInfo    *model.GradingInfo `json:"result"`
+}
+
+func GradeToSubmissionResponse(request APIRequestAssignmentContext, submissionPath string, email string, message string, options grader.GradeOptions) BaseSubmitResponse {
```
eriq-augustine, line 17: The name feel a bit off.
I feel like names like this usually imply a conversion, but we are doing a full grading (which is conceptually heavier).
Maybe something like "GradeRequestSubmission"?

internal/api/courses/assignments/submissions/proxy/resubmit.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/grader"
+	"github.com/edulinq/autograder/internal/timestamp"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type ResubmitRequest struct {
+	core.APIRequestAssignmentContext
+	core.MinCourseRoleGrader
```
eriq-augustine, line 13: Put a newline here.
I like to separate the admin fiends from the real data fields.

internal/api/courses/assignments/submissions/proxy/resubmit.go
```diff
@@ -0,0 +1,76 @@
...
+	TargetSubmission string `json:"target-submission"`
+
+	ProxyUser core.TargetCourseUser `json:"proxy-email"`
+	ProxyTime *timestamp.Timestamp  `json:"proxy-time"`
+
+	Message string `json:"message"`
```
eriq-augustine, line 19: I think we should probably just keep the message as is.
You could try to work out the semantics of dealing with this new message (i.e., the four possible cases) if you think it is important,
but I think it's easiest just to leave it off.

internal/api/courses/assignments/submissions/proxy/resubmit.go
```diff
@@ -0,0 +1,76 @@
...
+	if !request.ProxyUser.Found {
+		return &response, nil
+	}
+
+	response.FoundUser = true
+
+	gradingResult, err := db.GetSubmissionContents(request.Assignment, request.ProxyUser.Email, request.TargetSubmission)
+	if err != nil {
+		return nil, core.NewInternalError("-632", &request.APIRequestCourseUserContext, "Failed to get submission contents.").
+			Err(err).Assignment(request.Assignment.GetID()).
```
eriq-augustine, line 42: Just an FYI, there should be (but is not) an error function for core.APIRequestAssignmentContext.
I'm going to make one, so you will probably have to deal with a merge
(depending on when it gets in).

eriq-augustine (changes requested):
Looking pretty good.
There is one design decision to make, but everything else is pretty minor.

eriq-augustine (approved):
lgtm!

PR #170, opened 2025-04-09
Added an API endpoint to make proxy regrades.

Lucas-Ellenberger:
This PR adds a new API endpoint `courses/assignment/submissions/proxy/regrade` that allows course graders (or above) to make mass proxy regrades based on a given role.

Regrades take the most recent submission for every user with the filtered role and resubmit with a proxy time that is the same time as the original submission. This features allows course staff to easily regrade every students (or another role's) submissions.

As regrades may take a long time (as there could be many students), the endpoint gives the option to wait for completion. If the requesting user does not wait for completion, the results will be available by looking at the submission history.

### DIFF
--- a/internal/api/courses/assignments/submissions/proxy/regrade.go
+++ b/internal/api/courses/assignments/submissions/proxy/regrade.go
@@ -16,13 +16,13 @@ type RegradeRequest struct {
 }
 
 type RegradeResponse struct {
-	SubmissionInfo map[string]*model.SubmissionHistoryItem `json:"submission-infos"`
+	Complete bool                                    `json:"complete"`
+	Options  grader.RegradeOptions                   `json:"options"`
+	Results  map[string]*model.SubmissionHistoryItem `json:"results"`
 }
 
-// Regrade the most recent submissions for the course.
+// Regrade the most recent submissions for users with the filtered role in the course.
 func HandleRegrade(request *RegradeRequest) (*RegradeResponse, *core.APIError) {
-	response := RegradeResponse{}
-
 	role := model.GetCourseUserRoleString(request.FilterRole)
 	users, err := db.ResolveCourseUsers(request.Course, []string{role})
 	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/api/courses/assignments/submissions/proxy/regrade.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/edulinq/autograder/internal/grader"
+	"github.com/edulinq/autograder/internal/model"
+)
+
+type RegradeRequest struct {
+	core.APIRequestAssignmentContext
+	core.MinCourseRoleGrader
+
+	// Filter results to only users with this role.
+	FilterRole model.CourseUserRole `json:"filter-role"`
```
eriq-augustine, line 15: I think we should target emails rather than roles.
If no emails, do all students (so, default to `"student"`).

I personally think that we should think about phasing out `filter-role` and go for how we handle email resolution instead.

internal/api/courses/assignments/submissions/proxy/regrade.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Wait for the entire regrade to complete and return all results.
+	WaitForCompletion bool `json:"wait-for-completion"`
+}
+
+type RegradeResponse struct {
+	Complete bool                                    `json:"complete"`
```
eriq-augustine, line 22: I don't think this value really makes sense without caching.
For analysis, it can make sense because you may not be waiting, but get all your results from the cache.

internal/api/courses/assignments/submissions/proxy/regrade.go
```diff
@@ -0,0 +1,60 @@
...
+	Users    []string                                `json:"users"`
```
eriq-augustine, line 23: This doesn't seem necessary when you have all the emails in Results.

eriq-augustine (changes requested):
Stopped early, see slack.

Lucas-Ellenberger:
This PR is outdated. Refer to the new PR: #195.

PR #171, opened 2025-04-10
Added descriptions of structs to the API type descriptions.

Lucas-Ellenberger:
This PR adds descriptions of structs to the API type descriptions.

When given a `reflect.Type`, describe can parse the comment off of the type declaration in the source code. This way, we can automatically add custom descriptions to types in the API description's type map.

As not all types that appear in the type map have comments, many still do not have a description.

This PR also improves the structure of the description of struct fields in preparation to include field descriptions, required fields, default field values, etc.

### DIFF
--- a/internal/util/parser.go
+++ b/internal/util/parser.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 )
 
+var knownPackagePaths = map[string]map[string]string{}
+
 // This function takes a file path and a pattern that matches the name of the target function.
 // Returns the comment attached to the first occurrence of the target function.
 // Errors occur when the target function cannot be found.
@@ -45,23 +47,34 @@ func GetDescriptionFromFunction(path string, functionNamePattern *regexp.Regexp)
 	return "", fmt.Errorf("Unable to find a description using the pattern '%s' for a function located in '%s'.", functionNamePattern.String(), path)
 }
 
-func GetDescriptionFromType(packagePath string, typeName string) (string, error) {
-	dirPath := getDirPathFromPackagePath(packagePath)
+// This function takes a non-empty package path and returns a map of custom type names to their description.
+// The types of package paths accepted can be seen in getDirPathFromCustomPackagePath().
+func GetAllTypeDescriptionsFromPackage(packagePath string) (map[string]string, error) {
+	descriptions, ok := knownPackagePaths[packagePath]
+	if ok {
+		return descriptions, nil
+	}
+
+	dirPath := getDirPathFromCustomPackagePath(packagePath)
 
 	filePaths, err := FindFiles("", dirPath)
 	if err != nil {
-		return "", fmt.Errorf("Unable to find file paths for the package path '%s': '%v'.", packagePath, err)
+		return map[string]string{}, fmt.Errorf("Unable to find file paths for the package path '%s': '%v'.", packagePath, err)
 	}
 
-	description, err := getDescriptionFromType(filePaths, typeName)
+	descriptions, err = getDescriptionFromType(filePaths)
 	if err != nil {
-		return "", fmt.Errorf("Unable to get description for the type '%s' in '%s': '%v'.", typeName, packagePath, err)
+		return map[string]string{}, fmt.Errorf("Unable to get descriptions for the package path '%s': '%v'.", packagePath, err)
 	}
 
-	return description, nil
+	knownPackagePaths[packagePath] = descriptions
+
+	return descriptions, nil
 }
 
-func getDescriptionFromType(filePaths []string, typeName string) (string, error) {
+func getDescriptionFromType(filePaths []string) (map[string]string, error) {
+	descriptions := make(map[string]string, 0)
+
 	for _, path := range filePaths {
 		if !IsFile(path) {
 			continue
@@ -74,7 +87,7 @@ func getDescriptionFromType(filePaths []string, typeName string) (string, error)
 		fileSet := token.NewFileSet()
 		node, err := parser.ParseFile(fileSet, path, nil, parser.ParseComments)
 		if err != nil {
-			return "", fmt.Errorf("Error while parsing file to get function description: '%v'.", err)
+			return map[string]string{}, fmt.Errorf("Error while parsing file to get function description: '%v'.", err)
 		}
 
 		for _, decl := range node.Decls {
@@ -97,16 +110,15 @@ func getDescriptionFromType(filePaths []string, typeName string) (string, error)
 					continue
 				}
 
-				if typeSpec.Name.Name == typeName {
-					if genDecl.Doc == nil {
-						return "", nil
-					}
-
-					return strings.TrimSpace(genDecl.Doc.Text()), nil
+				if genDecl.Doc == nil {
+					descriptions[typeSpec.Name.Name] = ""
+					continue
 				}
+
+				descriptions[typeSpec.Name.Name] = strings.TrimSpace(genDecl.Doc.Text())
 			}
 		}
 	}
 
-	return "", fmt.Errorf("Unable to find a description using the name '%s' for a type located in '%v'.", typeName, filePaths)
+	return descriptions, nil
 }
--- a/internal/util/parser.go
+++ b/internal/util/parser.go
@@ -44,3 +44,69 @@ func GetDescriptionFromFunction(path string, functionNamePattern *regexp.Regexp)
 
 	return "", fmt.Errorf("Unable to find a description using the pattern '%s' for a function located in '%s'.", functionNamePattern.String(), path)
 }
+
+func GetDescriptionFromType(packagePath string, typeName string) (string, error) {
+	dirPath := getDirPathFromPackagePath(packagePath)
+
+	filePaths, err := FindFiles("", dirPath)
+	if err != nil {
+		return "", fmt.Errorf("Unable to find file paths for the package path '%s': '%v'.", packagePath, err)
+	}
+
+	description, err := getDescriptionFromType(filePaths, typeName)
+	if err != nil {
+		return "", fmt.Errorf("Unable to get description for the type '%s' in '%s': '%v'.", typeName, packagePath, err)
+	}
+
+	return description, nil
+}
+
+func getDescriptionFromType(filePaths []string, typeName string) (string, error) {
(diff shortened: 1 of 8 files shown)


internal/util/path.go
```diff
@@ -297,3 +297,13 @@ func ShouldNormalizePath(path string) string {
...
 }
+
+// Best effort attempt to determine the absolute path to the directory that holds a custom package path.
+// The package path is expected to be an internal path of the form `github.com/edulinq/autograder/`.
+func getDirPathFromCustomPackagePath(packagePath string) string {
+	if strings.HasPrefix(packagePath, "github.com/edulinq/autograder/") {
+		packagePath = strings.TrimPrefix(packagePath, "github.com/edulinq/autograder/")
+	}
+
+	return ShouldAbs(filepath.Join(ShouldGetThisDir(), "..", "..", packagePath))
```
eriq-augustine, line 308: There is a bit of a mix of path types here.
I think the package path in Go has to have forward slashes, so that's fine.
But, then it could still have those forward slashes when you join.
I know we are POSIX only, but we try to be flexible as we can.

eriq-augustine (changes requested):
In general, the code is looking good.
But, I was getting a lot of flaky fails in the CI for this.

You may want to implement the caching we talk about now and see if it helps.

cmd/call-endpoint/main.go
```diff
@@ -45,7 +45,7 @@ func main() {
 		return
 	}
 
-	apiDescription, err := api.Describe(*api.GetRoutes())
+	apiDescription, err := core.Describe(*api.GetRoutes())
```
eriq-augustine, line 48: Now that this function is moved, is needs a more descriptive name.
DescribeRoutes()?

internal/api/core/routes.go
```diff
@@ -9,6 +9,10 @@ import (
 
 var CURRENT_PREFIX string = fmt.Sprintf("/api/v%02d", util.MustGetAPIVersion())
 
+// API routes will be empty until RunServer() is called.
+// This is a cached version of api.GetRoutes() for packages that cannot access `internal/api` due to import cycles.
+var apiRoutes []Route
```
eriq-augustine, line 14: Explicitly initialize this as nil.
It is not strictly necessary, but it's nice to do.

internal/api/core/routes.go
```diff
@@ -9,6 +9,10 @@ import (
 
 var CURRENT_PREFIX string = fmt.Sprintf("/api/v%02d", util.MustGetAPIVersion())
 
+// API routes will be empty until RunServer() is called.
+// This is a cached version of api.GetRoutes() for packages that cannot access `internal/api` due to import cycles.
```
eriq-augustine, line 13: Swap the order of these sentences.
You always want your first sentence to be the general description/overview.

internal/api/metadata/describe.go
```diff
@@ -14,7 +16,24 @@ type DescribeResponse struct {
...
+			return nil, core.NewInternalError("-501", request, "Unable to describe API endpoints when the cached routes are empty.")
+		}
+
+		apiDescription, err = core.Describe(routes)
+		if err != nil {
+			return nil, core.NewInternalError("-502", request, "Failed to describe API endpoints.").Err(err)
+		}
+
+		core.SetAPIDescription(apiDescription)
+	}
```
eriq-augustine, line 34: I don't think the caching should be done like this.
You are still always forcing a high-cost operation on the first call.
Instead, I think you should just pull the description from the resources dir unless someone explicitly asked for a full computation (or the resources is missing).

internal/config/options.go
```diff
@@ -66,5 +66,8 @@ var (
 	ANALYSIS_INDIVIDUAL_COURSE_POOL_SIZE = MustNewIntOption("analysis.individual.poolsize", 1, "The number of parallel workers per course when computing individual analysis.")
 	ANALYSIS_PAIRWISE_COURSE_POOL_SIZE   = MustNewIntOption("analysis.pairwise.poolsize", 1, "The number of parallel workers per course when computing pairwise analysis.")
 
+	// Metadata
+	UPDATE_API_DESCRIPTIONS = MustNewBoolOption("update.api.descriptions", false, "Update API descriptions when starting the server.")
```
eriq-augustine, line 70: I don't think this should exist.
See my comments on the endpoint caching.

internal/util/parser.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 )
 
+var knownPackagePaths = map[string]map[string]string{}
```
eriq-augustine, line 12: Given the caching you are already doing (and the version I recommended in these comments),
this package variable may not need to exist anymore.

internal/util/path.go
```diff
@@ -297,3 +297,20 @@ func ShouldNormalizePath(path string) string {
...
+// Best effort attempt to determine the absolute path to the directory that holds a custom package path.
+// The package path is expected to be an internal path of the form `github.com/edulinq/autograder/`.
+func getDirPathFromCustomPackagePath(packagePath string) string {
+	if strings.HasPrefix(packagePath, "github.com/edulinq/autograder/") {
+		packagePath = strings.TrimPrefix(packagePath, "github.com/edulinq/autograder/")
+	}
+
+	parts := strings.Split(packagePath, "/")
+
+	path := ShouldAbs(filepath.Join(ShouldGetThisDir(), "..", ".."))
```
eriq-augustine, line 310: Comment this line.

eriq-augustine (changes requested):
The type exploration is looking good.
I think the caching needs some work.

internal/util/parser.go
Lucas-Ellenberger: I still want to cache the package paths because it is faster to describe all types in a package path (directory), then searching for a single type at a time. For example, there are many types in model that will need to be described, so it's easiest to cache the entire package.

If we are worried about the space of the cache, then it could be removed since it will only be used on `force-update` and they are already willing to wait for the most up to date info. But, I don't see the need to slow down the operation more.

eriq-augustine: You should probably cache outside the function to have caching and still manage the lifetime of the cache.
Since the cache is keyed by a path, it should be no problem to just check the cache first before calling.

internal/api/metadata/describe.go
```diff
@@ -14,7 +17,35 @@ type DescribeResponse struct {
...
+	if routes == nil || len(*routes) == 0 {
+		return nil, core.NewInternalError("-501", request, "Unable to describe API endpoints when the cached routes are empty.")
+	}
+
+	apiDescription, err := core.DescribeRoutes(*routes)
+	if err != nil {
+		return nil, core.NewInternalError("-502", request, "Failed to describe API endpoints.").Err(err)
+	}
+
+	core.SetAPIDescription(apiDescription)
```
eriq-augustine, line 44: I'm not a fan of this call being here.
It seems a bit off that an endpoint is setting a cache inside of code
(it feels like core should be the one setting its own cache).

What if GetAPIDescription() accepted a boolean for forcing the recompute?
Then, most of this function would fall inside GetAPIDescription().

It's also fair to think about if you need this cache at all.
Aside from testing, it would just save one disk read.

internal/api/metadata/describe.go
```diff
@@ -14,7 +17,35 @@ type DescribeResponse struct {
...
+
+	if apiDescription != nil {
+		response.APIDescription = *apiDescription
```
eriq-augustine, line 47: Is it possible to embed a pointer instead so you don't have to make a copy?
(IDK if it is possible.)

internal/api/metadata/describe_test.go
```diff
@@ -5,32 +5,189 @@ import (
 	"testing"
 
 	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/db"
 	"github.com/edulinq/autograder/internal/util"
 )
 
 func TestMetadataDescribe(test *testing.T) {
+	defer db.ResetForTesting()
```
eriq-augustine, line 13: What do you need the db for in this test?

eriq-augustine (changes requested):
Looking pretty good.

internal/api/metadata/describe_test.go
```diff
@@ -9,28 +9,199 @@ import (
...
+		expected := DescribeResponse{testCase.Description}
+		if !reflect.DeepEqual(expected, responseContent) {
+			if !softEqual(test, i, testCase.Description, responseContent.APIDescription) {
+				continue
+			}
+		}
+	}
+}
+
+func softEqual(test *testing.T, testNum int, expected *core.APIDescription, actual *core.APIDescription) bool {
```
eriq-augustine, line 177: I'm not sure why this function is necessary over just reflect.DeepEqual().

If you want to get better failure messages, then that makes sense.
But, this function is currently incomplete.
It ranges over `expected` for the two main maps, but does not ensure that `actual` has the same size.

If you are not going to better failure messages, then I'm not sure why reflect.DeepEqual() is not sufficient.
The usual culprits are missing a sort or having elements that need to be zeroed out (like a timestamp).
If it is the non-JSON fields (`json:"-"`) that are giving you trouble,
another option is to convert to JSON and then compare the strings (just with `==`).

eriq-augustine (changes requested):
Just one testing thing.

internal/api/metadata/describe_test.go
Lucas-Ellenberger: I believe I need the weaker test. The better messages are nice, but I have lost the ability to set the API description. Previously, I would set it to something simple and then use DeepEqual to ensure the response was correct. I removed the in memory cache and I do not want to have the ability to set the on disk cache, so I don't have an easy way to have a simple check.

With "force-compute", I can limit the output to just "metadata/describe" which is easy to maintain (and it is DeepEqual).

However, with force-compute false, I am going to receive the entirety of the API description. As you could imagine, this would be cumbersome to maintain as the script wouldn't be able to update the test data. Also, this functionality is tested in "api/describe_test.go".

My concession for testing was to ensure that some (but definitely not all) of the endpoints and types were described correctly.

I'm open to better testing set-ups, but this was my best idea.

eriq-augustine (approved):
lgtm!

PR #174, opened 2025-04-18
Created jobmanager package to abstract some infrastructure from analysis.

Lucas-Ellenberger:
This PR creates a new package, jobmanager. The jobmanager package includes common infrastructure for large jobs that are too big too be run full serial or full parallel.

The jobmanager package provides `RunJob()` that has the utility of parallel pools while abstracting some of the implementation details of synchronization, caching, over-writing caches, and contexts.

`RunJob()` is thoroughly tested and is used by `internal/analysis` for both individual and pairwise. These examples give confidence in the flexibility of the `RunJob()` framework.

### DIFF
--- a/internal/jobmanager/jobmanager.go
+++ b/internal/jobmanager/jobmanager.go
@@ -15,24 +15,21 @@ import (
 // JobOptions contains user level options for a Job.
 // These options allow for optional context cancellation of the Job.
 type JobOptions struct {
+	// Don't save anything.
+	DryRun bool `json:"dry-run"`
+
 	// Remove any existing records before running the job.
 	OverwriteRecords bool `json:"overwrite-records"`
 
 	// Wait for the entire job to complete and return all results.
 	WaitForCompletion bool `json:"wait-for-completion"`
 
-	// If true, do not swap the context to the background context when running.
-	// By default (when this is false), the context will be swapped to the background context when !WaitForCompletion.
-	// The swap is so that jobs do not get canceled when an HTTP request is complete.
-	// Setting this true is useful for testing (as one round of job tests can be wrapped up).
-	RetainOriginalContext bool `json:"-"`
-
 	// A context that can be used to cancel the job.
 	Context context.Context `json:"-"`
 }
 
 // Job provides system level customization of the job's execution.
-// It supports optional record retrieval, record removal from storage, synchronization, and context cancellation.
+// It supports optional record retrieval, record storage, record removal from storage, synchronization, and context cancellation.
 // Given job options, input items, an optional record lookup/removal function, and a work function,
 // Job will process a list of input items through the work func in a parallel pool to produce a JobOutput.
 // If a lock key is provided, Job will block on that key before running preventing overuse of resources or conflicts between the same Job.
@@ -45,6 +42,7 @@ type Job[InputType any, OutputType any] struct {
 	// Only access the state of the output when JobOutput.Done signals.
 	// If the context is cancelled during execution,
 	// JobOutput's state is unknown.
+	// TODO: Remove and modify the shared address.
 	JobOutput[InputType, OutputType]
 
 	// The number of workers in the parallel pool.
@@ -57,8 +55,11 @@ type Job[InputType any, OutputType any] struct {
 	WorkItems []InputType
 
 	// An optional function to retrieve existing records that should not be processed.
-	// Returns a list of processed records, processed items, remaining items, and an error.
-	RetrieveFunc func([]InputType) ([]OutputType, []InputType, []InputType, error)
+	// Returns a list of processed records, remaining items, and an error.
+	RetrieveFunc func([]InputType) ([]OutputType, []InputType, error)
+
+	// An optional function to store the result.
+	StoreFunc func([]OutputType) error
 
 	// An optional function to remove existing records from storage.
 	// TODO: Find a way to not remove when on DryRun (may need to modify retrieve func).
@@ -82,6 +83,7 @@ type Job[InputType any, OutputType any] struct {
 // If the context is cacelled while running a job,
 // the returned JobOutput will be empty.
 type JobOutput[InputType any, OutputType any] struct {
+	// TODO: Can lock shared JobOutput and provide interface methods to get data (that handles the locking).
 	// An error for Job.Run().
 	Error error
 
@@ -148,10 +150,6 @@ func (this *JobOptions) Validate() error {
 		this.Context = context.Background()
 	}
 
-	if !this.WaitForCompletion && !this.RetainOriginalContext {
-		this.Context = context.Background()
-	}
-
 	return nil
 }
 
@@ -171,33 +169,20 @@ func (this *Job[InputType, OutputType]) Run() *JobOutput[InputType, OutputType] 
 	this.RunTime = 0
 	this.WorkErrors = make(map[int]error, 0)
 
-	processedItems := []InputType{}
-
-	// Query for stored records.
-	if this.RetrieveFunc != nil {
-		this.ResultItems, processedItems, this.RemainingItems, this.Error = this.RetrieveFunc(this.WorkItems)
+	// If we are overwriting records, remove all the old records.
+	if this.OverwriteRecords && !this.DryRun && this.RemoveStorageFunc != nil {
+		this.Error = this.RemoveStorageFunc(this.WorkItems)
 		if this.Error != nil {
 			return &this.JobOutput
 		}
 	}
 
-	// If we are overwriting records, remove all the old records.
-	if this.OverwriteRecords {
-		// Clear output when overwriting records.
-		this.ResultItems = []OutputType{}
-		this.RemainingItems = append(processedItems, this.RemainingItems...)
-
-		if this.RemoveStorageFunc != nil {
-			this.Error = this.RemoveStorageFunc(processedItems)
-			if this.Error != nil {
-				return &this.JobOutput
-			}
+	// Query for stored records.
+	if !this.OverwriteRecords && this.RetrieveFunc != nil {
+		this.ResultItems, this.RemainingItems, this.Error = this.RetrieveFunc(this.WorkItems)
+		if this.Error != nil {
+			return &this.JobOutput
 		}
-	}
-
-	if this.done == nil {
-		this.done = make(chan any)
-		this.Done = this.done
 	}
 
 	if this.WaitForCompletion {
@@ -211,7 +196,12 @@ func (this *Job[InputType, OutputType]) Run() *JobOutput[InputType, OutputType] 
 			return nil
 		}
 	} else {
+		// TODO: this.(non embedded)JobOutput.Copy()
+		// return copy
 		go func() {
+			// TEST
+			// Make a new job output here.
+			// Only takes remaining items.
 			this.run()
 			if this.Error != nil {
 				log.Error("Failure while running asynchronous job: '%v'.", this.Error)
@@ -225,6 +215,8 @@ func (this *Job[InputType, OutputType]) Run() *JobOutput[InputType, OutputType] 
 	return &this.JobOutput
 }
 
+// TODO: Can pass in a JobOutput pointer to modify.
+// If it's nil, don't save to results. (only errors)
 func (this *Job[InputType, OutputType]) run() {
 	if len(this.RemainingItems) == 0 {
 		return
@@ -241,13 +233,12 @@ func (this *Job[InputType, OutputType]) run() {
 		return
 	}
 
-	processedItems := []InputType{}
 	var err error = nil
 
 	// If we had to wait for the lock, then check again for stored records.
-	if !noLockWait && this.RetrieveFunc != nil {
-		var partialResults []OutputType = nil
-		partialResults, processedItems, this.RemainingItems, err = this.RetrieveFunc(this.RemainingItems)
+	if !noLockWait && !this.OverwriteRecords && this.RetrieveFunc != nil {
+		partialResults := []OutputType{}
+		partialResults, this.RemainingItems, err = this.RetrieveFunc(this.RemainingItems)
 		if err != nil {
 			this.Error = fmt.Errorf("Failed to re-check record storage before run: '%w'.", err)
 			return
@@ -259,21 +250,6 @@ func (this *Job[InputType, OutputType]) run() {
 
 	if len(this.RemainingItems) == 0 {
 		return
-	}
-
-	// If we are overwriting records, remove all the old records.
-	if this.OverwriteRecords {
-		// Clear output when overwriting records.
-		this.ResultItems = []OutputType{}
-		this.RemainingItems = append(processedItems, this.RemainingItems...)
-
-		if this.RemoveStorageFunc != nil {
-			err = this.RemoveStorageFunc(processedItems)
-			if err != nil {
-				this.Error = fmt.Errorf("Failed to remove old job cache entries: '%w'.", err)
-				return
-			}
-		}
 	}
 
 	type PoolResult struct {
@@ -301,9 +277,11 @@ func (this *Job[InputType, OutputType]) run() {
 
 		startTime := timestamp.Now()
 
+		var errs error = err
+
 		result, err := this.WorkFunc(workItem)
 		if err != nil {
-			err = fmt.Errorf("Failed to perform individual work on item %v: '%w'.", workItem, err)
+			errs = errors.Join(errs, fmt.Errorf("Failed to perform individual work on item %v: '%w'.", workItem, err))
 		}
 
 		runTime := (timestamp.Now() - startTime).ToMSecs()
@@ -312,7 +290,16 @@ func (this *Job[InputType, OutputType]) run() {
 			runTime = 1
 		}
 
-		return PoolResult{workItem, result, runTime, err}, nil
+		// We can separate the semantic options from the save func and retrieve func.
+		// Store the result.
+		if !this.DryRun && this.Context.Err() == nil && this.StoreFunc != nil {
+			err = this.StoreFunc([]OutputType{result})
+			if err != nil {
+				errs = errors.Join(errs, fmt.Errorf("Failed to store result for item '%v': '%w'.", workItem, err))
+			}
+		}
+
+		return PoolResult{workItem, result, runTime, errs}, nil
 	})
 
 	if err != nil {
--- a/internal/jobmanager/jobmanager.go
+++ b/internal/jobmanager/jobmanager.go
@@ -38,13 +38,6 @@ type Job[InputType any, OutputType any] struct {
 	// The user level options for a Job.
 	*JobOptions
 
-	// The current state of the output.
-	// Only access the state of the output when JobOutput.Done signals.
-	// If the context is cancelled during execution,
-	// JobOutput's state is unknown.
-	// TODO: Remove and modify the shared address.
-	JobOutput[InputType, OutputType]
-
 	// The number of workers in the parallel pool.
 	PoolSize int
 
@@ -71,10 +64,6 @@ type Job[InputType any, OutputType any] struct {
 
 	// An optional function to get the locking key for an individual work item.
 	WorkItemKeyFunc func(InputType) string
-
-	// An internal channel to signal the job is complete.
-	// Callers are notified the job is complete via JobOptions.Done.
-	done chan any
 }
 
 // JobOutput is the result of a call to Job.Run().
@@ -106,10 +95,6 @@ type JobOutput[InputType any, OutputType any] struct {
 }
 
 func (this *Job[InputType, OutputType]) Validate() error {
-	return this.validateFull(false)
-}
-
-func (this *Job[InputType, OutputType]) validateFull(setChannel bool) error {
 	if this == nil {
 		return fmt.Errorf("Job is nil.")
 	}
@@ -118,27 +103,11 @@ func (this *Job[InputType, OutputType]) validateFull(setChannel bool) error {
 		return fmt.Errorf("Job cannot have a nil work function.")
 	}
 
-	err := this.JobOptions.Validate()
-	if err != nil {
-		return err
-	}
-
-	if setChannel {
-		if this.done != nil {
-			return fmt.Errorf("Job is actively running and cannot be run again.")
-		}
-
-		this.JobOutput = JobOutput[InputType, OutputType]{}
-
-		this.done = make(chan any)
-		this.Done = this.done
-	}
-
 	if this.PoolSize <= 0 {
 		return fmt.Errorf("Pool size must be positive, got %d.", this.PoolSize)
 	}
 
-	return nil
+	return this.JobOptions.Validate()
 }
 
 func (this *JobOptions) Validate() error {
@@ -158,38 +127,43 @@ func (this *JobOptions) Validate() error {
 // If the context is canceled during execution, returns nil.
 // When not waiting for completion, Job.JobOutput will be populated with the results when the JobOutput.Done channel is closed.
 func (this *Job[InputType, OutputType]) Run() *JobOutput[InputType, OutputType] {
-	err := this.validateFull(true)
+	err := this.Validate()
 	if err != nil {
-		this.Error = fmt.Errorf("Failed to validate job: '%v'.", err)
-		return &this.JobOutput
+		return &JobOutput[InputType, OutputType]{
+			Error: fmt.Errorf("Failed to validate job: '%v'.", err),
+		}
 	}
 
-	this.ResultItems = make([]OutputType, 0, len(this.WorkItems))
-	this.RemainingItems = this.WorkItems
-	this.RunTime = 0
-	this.WorkErrors = make(map[int]error, 0)
+	done := make(chan any)
+
+	output := JobOutput[InputType, OutputType]{
+		Done:           done,
+		ResultItems:    make([]OutputType, 0, len(this.WorkItems)),
+		RemainingItems: this.WorkItems,
+		RunTime:        0,
+		WorkErrors:     make(map[int]error, 0),
+	}
 
 	// If we are overwriting records, remove all the old records.
 	if this.OverwriteRecords && !this.DryRun && this.RemoveStorageFunc != nil {
-		this.Error = this.RemoveStorageFunc(this.WorkItems)
-		if this.Error != nil {
-			return &this.JobOutput
+		output.Error = this.RemoveStorageFunc(this.WorkItems)
+		if output.Error != nil {
+			return &output
 		}
 	}
 
 	// Query for stored records.
 	if !this.OverwriteRecords && this.RetrieveFunc != nil {
-		this.ResultItems, this.RemainingItems, this.Error = this.RetrieveFunc(this.WorkItems)
-		if this.Error != nil {
-			return &this.JobOutput
+		output.ResultItems, output.RemainingItems, output.Error = this.RetrieveFunc(this.WorkItems)
+		if output.Error != nil {
+			return &output
 		}
 	}
 
 	if this.WaitForCompletion {
-		this.run()
+		this.run(&output)
 
-		close(this.done)
-		this.done = nil
+		close(done)
 
 		// If the context was canceled during execution, return immediately.
 		if this.Context.Err() != nil {
@@ -202,23 +176,22 @@ func (this *Job[InputType, OutputType]) Run() *JobOutput[InputType, OutputType] 
 			// TEST
 			// Make a new job output here.
 			// Only takes remaining items.
-			this.run()
-			if this.Error != nil {
-				log.Error("Failure while running asynchronous job: '%v'.", this.Error)
+			this.run(&output)
+			if output.Error != nil {
+				log.Error("Failure while running asynchronous job: '%v'.", output.Error)
 			}
 
-			close(this.done)
-			this.done = nil
+			close(done)
 		}()
 	}
 
-	return &this.JobOutput
+	return &output
 }
 
 // TODO: Can pass in a JobOutput pointer to modify.
 // If it's nil, don't save to results. (only errors)
-func (this *Job[InputType, OutputType]) run() {
-	if len(this.RemainingItems) == 0 {
+func (this *Job[InputType, OutputType]) run(output *JobOutput[InputType, OutputType]) {
+	if len(output.RemainingItems) == 0 {
 		return
 	}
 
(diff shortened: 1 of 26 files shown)


internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type JobOptions struct {
+	// Don't save anything.
+	DryRun bool `json:"dry-run"`
+
+	// Replace any entries currently in the cache,
+	// and do not return any cached entries (when not waiting for completion).
+	OverwriteCache bool `json:"overwrite-cache"`
```
eriq-augustine, line 19: Given the more general nature of jobs, we may want to move away from the "cache" terminology,
and instead talk about "storing" and "retrieving" "records".

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+
+	LockKey string `json:"-"`
+}
+
+func (this *JobOptions) Validate() error {
+	if this == nil {
+		return fmt.Errorf("Job options are nil.")
+	}
+
+	if this.LockKey == "" {
```
eriq-augustine, line 43: Could it be possible to have a job that doesn't use a lock?

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+	return nil
+}
+
+// RunJob() executes a potentially long-running job over a list of input items.
+// It supports optional caching, cache removal, synchronization, and context cancellation.
+// Given job options, input items, an optional cache lookup/removal function, and a work function,
+// RunJob() processes each item in a parallel pool.
+// Returns the result list, number of remaining items, total run time, and an error.
+// If the context is canceled, returns nil, 0, 0, nil.
+func RunJob[InputType any, OutputType any](options *JobOptions, workItems []InputType, cacheFunc func([]InputType) ([]OutputType, []InputType, error), removeCacheFunc func([]InputType) error, workFunc func(InputType) (OutputType, int64, error)) ([]OutputType, int, int64, error) {
```
eriq-augustine, line 68: Given these large signatures, I think you should consider having a "Job" struct.
This can house the options and inputs.
I could see the outputs (maybe including errors) being either in this same struct or in and output (JobOutput) struct.

Maybe then we can also have this as a method (Run()) instead of a function.
This may be nice, since we could carry around state more easily.

This may also make is possible to spit up the JobOptions to it only includes user-centric options (like ones we want to embed in the API),
and leave the system options for the Job struct.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	return nil
+}
+
+// RunJob() executes a potentially long-running job over a list of input items.
+// It supports optional caching, cache removal, synchronization, and context cancellation.
+// Given job options, input items, an optional cache lookup/removal function, and a work function,
+// RunJob() processes each item in a parallel pool.
+// Returns the result list, number of remaining items, total run time, and an error.
```
eriq-augustine, line 66: Expand upon the locking semantics.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+// If the context is canceled, returns nil, 0, 0, nil.
+func RunJob[InputType any, OutputType any](options *JobOptions, workItems []InputType, cacheFunc func([]InputType) ([]OutputType, []InputType, error), removeCacheFunc func([]InputType) error, workFunc func(InputType) (OutputType, int64, error)) ([]OutputType, int, int64, error) {
+	if workFunc == nil {
```
eriq-augustine, line 69: Putting everything in a Job struct (mentioned above) would also let you have a central Validate() method that does things like this and calls options.Validate().

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		completeItems = append(completeItems, results...)
+		remainingItems = nil
+	} else {
+		go func() {
+			_, _, err := runJob(options, remainingItems, cacheFunc, removeCacheFunc, workFunc)
+			if err != nil {
+				log.Error("Failure while running asynchronous job.")
+			}
```
eriq-augustine, line 110: If we have an output struct (whether or not that is apart of the input struct),
then we could do something pretty cool here.
We can have a receive-only chan in the struct that tells the caller when the job is actually done (the chan will be closed on done, like a context).
On WaitForCompletion, that chan will be closed before we return, but for async we can close in this gofunc after we actually finish.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+	// The context has been canceled while waiting for a lock, abandon this job.
+	if options.Context.Err() != nil {
+		return nil, 0, nil
+	}
+
+	results := make([]OutputType, 0, len(workItems))
+
+	// If we had to wait for the lock, then check again for cached results.
+	// If there are multiple requests queued up,
+	// it will be faster to do a bulk check for cached results instead of checking each one individually.
```
eriq-augustine, line 139: This sentence no longer makes sense in this context.
For analysis, we would check caches in bulk and on an individual level,
but here we only check in bulk.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,200 @@
...
+		// Collect the partial results from the cache.
+		results = append(results, partialResults...)
+	}
+
+	if len(workItems) == 0 {
+		return results, 0, nil
+	}
+
+	// If we are overwriting the cache, then remove all the old entries.
+	if options.OverwriteCache && removeCacheFunc != nil && !options.DryRun {
```
eriq-augustine, line 156: I think the semantics here may be a bit off (and they are questionable in analysis).
We just checked the cache (and we checked even before the lock),
but now we are removing `workItems`, which is only the non-cached results.
So I think we may have not overwritten the cache.

Note that analysis may still be wrong, but it is a little different since `options.OverwriteCache` is checked in their caching function:
https://github.com/edulinq/autograder-server/blob/main/internal/analysis/individual.go#L62

A warning here:
Do not worry about multiple calls (e.g. to analysis) overwriting each other in the cache.
Because of the lock they will run serially, and they should always give the same result with the same input.

internal/jobmanager/jobmanager_test.go
```diff
@@ -0,0 +1,674 @@
...
+			},
+			&JobOptions{
+				RetainOriginalContext: true,
+				Context:               context.Background(),
+				LockKey:               testLockKey,
+				PoolSize:              testPoolSize,
+			},
+			"",
+		},
+		// Nil context provided
```
eriq-augustine, line 67: Use Title Case when not using complete sentences for comments.

eriq-augustine (changes requested):
Looking pretty good.
I think we should shift how we think about the cache and the input/output of running a job, but the core is all here.

I mainly paid attention to the core job description and procedure for now.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	// If we are overwriting records, then remove all the old records.
+	if this.OverwriteRecords && this.RemoveStorageFunc != nil {
+		processedItems := getProcessedItems(this.WorkItems, this.RemainingItems)
+		err = this.RemoveStorageFunc(processedItems)
+		if err != nil {
+			return fmt.Errorf("Failed to remove old job cache entries: '%w'.", err)
+		}
+	}
```
eriq-augustine, line 220: !?

I just mentioned the strange semantics here and you still have the same code.

Old comment:
> I think the semantics here may be a bit off (and they are questionable in analysis).
> We just checked the cache (and we checked even before the lock),
> but now we are removing workItems, which is only the non-cached results.
> So I think we may have not overwritten the cache.
> 
> Note that analysis may still be wrong, but it is a little different since options.OverwriteCache is checked in their caching function:
> https://github.com/edulinq/autograder-server/blob/main/internal/analysis/individual.go#L62
> 
> A warning here:
> Do not worry about multiple calls (e.g. to analysis) overwriting each other in the cache.
> Because of the lock they will run serially, and they should always give the same result with the same input.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,280 @@
...
+
+type Job[InputType comparable, OutputType any] struct {
+	JobOptions
+
+	JobOutput[InputType, OutputType]
+
+	// A sorted list of items that need work.
+	WorkItems []InputType
+
+	// A function to retrieve existing records.
```
eriq-augustine, line 21: You should talk about the semantics here a bit more.
Specifically, an record that can be retrieved is considered "done" and should not be processed.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,280 @@
...
+		Error   error
+	}
+
+	poolResults, _, err := util.RunParallelPoolMap(this.PoolSize, this.RemainingItems, this.Context, func(workItem InputType) (PoolResult, error) {
+		result, runTime, err := this.WorkFunc(workItem)
+		if err != nil {
+			err = fmt.Errorf("Failed to perform individual work on item %v: '%w'.", workItem, err)
+		}
+
+		return PoolResult{workItem, result, runTime, err}, nil
```
eriq-augustine, line 235: I'm wondering if we can do more here.

We already take care of caching and locking at the bulk level,
but do you think we can do it here as well?
Like for analysis, we lock for the course, but then further lock for the specific record/analysis.
We could maybe do that here so the caller doesn't have to worry about it.
The caller would need to provide an optional function to produce a lock key from an input type (we already have the retrieval function).
This would also allow us to get pretty accurate timings so the caller wouldn't have to keep track of run time in their work function (mentioned in an earlier comment).

eriq-augustine (changes requested):
Getting better.
I still just focused on the core code.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,280 @@
...
+
+// Job.Run() executes a potentially long-running job over a list of input items.
+// It supports optional record retrieval, record removal from storage, synchronization, and context cancellation.
+// Given job options, input items, an optional record lookup/removal function, and a work function,
+// Job.Run() processes each item in a parallel pool.
+// If a lock key is provided, Job.Run() will block on that key for the duration of the method.
+// This prevents multiple Job.Run() invocations of the same type from overusing resources or conflicting with each other.
+// Returns the result list, number of remaining items, total run time, and an error.
+// If the context is canceled, returns nil, 0, 0, nil.
+func (this *Job[InputType, OutputType]) Run() (JobOutput[InputType, OutputType], error) {
```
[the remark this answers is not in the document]
Lucas-Ellenberger: I wasn't able to get this to work, but if you have suggestions that would make the code look cleaner.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+	// Remove any existing records before running the job.
+	OverwriteRecords bool `json:"overwrite-records"`
+
+	// Wait for the entire job to complete and return all results.
+	WaitForCompletion bool `json:"wait-for-completion"`
+
+	// Return a copy of an intermediate result.
+	// Will not be able to access final output.
+	// If WaitForCompletion is true, this option is overwritten.
+	ReturnCopy bool `json:"return-copy"`
```
eriq-augustine, line 45: I'm not a fan of this name, since it can be misconstrued.
We could return a copy of the final, complete results.

Make the name more specific, something more like "ReturnIncompleteResults".

eriq-augustine, line 45: This doesn't feel like it should be a user-level option.
We have talked before about separating the structs with these user vs system level options.
Regardless of your decision here, this should probably at least not be exported in JSON.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+
+	// A sorted list of items that need work.
+	WorkItems []InputType
+
+	// An optional function to retrieve existing records that should not be processed.
+	// Returns a list of processed records, remaining items, and an error.
+	RetrieveFunc func([]InputType) ([]OutputType, []InputType, error)
+
+	// An optional function to retrieve a record based on a single work item.
+	SingleRetreiveFunc func(InputType) (OutputType, bool, error)
```
eriq-augustine, line 83: I'm pretty skeptical on the need for these single function, when the batch ones work in the single case.
Do you have any specific performance case for this?

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+
+	// A function to transform work items into results.
+	// Returns a result, the time of computation, and an error.
+	WorkFunc func(InputType) (OutputType, error)
+
+	// An optional function to get the locking key for an individual work item.
+	WorkItemKeyFunc func(InputType) string
+
+	// An optional function to store stats on how long the job took.
+	StatFunc func(int64)
```
eriq-augustine, line 99: I don't see a strong reason to make this specific to stats.
This could just be some function that gets called with the full JobOutput object when the job is done.
Maybe "OnComplete()"?

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+	RetrieveFunc func([]InputType) ([]OutputType, []InputType, error)
+
+	// An optional function to retrieve a record based on a single work item.
+	SingleRetreiveFunc func(InputType) (OutputType, bool, error)
+
+	// An optional function to store the result.
+	StoreFunc func([]OutputType) error
+
+	// An optional function to remove existing records from storage.
+	RemoveStorageFunc func([]InputType) error
```
eriq-augustine, line 89: With your current naming scheme, "RemoveFunc" may be more consistent.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+	if this.Context == nil {
+		this.Context = context.Background()
+	}
+
+	return nil
+}
+
+func (this JobOutput[InputType, OutputType]) GetError() error {
+	lockmanager.ReadLock(this.JobID)
+	defer lockmanager.ReadUnlock(this.JobID)
```
eriq-augustine, line 162: Since you have `ReturnCopy` (which may be renamed),
I don't think you need any of this locking behavior.

Let me know if I am missing something, but either:
 - `ReturnCopy` is true and the caller has their own copy they can use.
 - `ReturnCopy` is false, and the caller should wait on the done chan.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+		ResultItems:    make([]OutputType, 0, len(this.WorkItems)),
+		RemainingItems: this.WorkItems,
+		RunTime:        0,
+		WorkErrors:     make(map[int]error, 0),
+		JobID:          util.UUID(),
+	}
+
+	err := this.Validate()
+	if err != nil {
+		close(done)
```
eriq-augustine, line 289: I would put your closes right up against the return,
to make it clear that you are doing this on every return except the final one.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+
+		output.Error = fmt.Errorf("Failed to validate job: '%v'.", err)
+		return &output
+	}
+
+	// If we are overwriting records, remove all the old records.
+	if this.OverwriteRecords && !this.DryRun && this.RemoveStorageFunc != nil {
+		output.Error = this.RemoveStorageFunc(this.WorkItems)
+		if output.Error != nil {
```
eriq-augustine, line 298: I would just save in another error variable here so you can wrap it with `fmt.Errorf()`.

It's not as big a deal for removal,
but for things like retrieval it can help let the caller know where the error happened.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+
+	if this.WaitForCompletion {
+		this.run(&output, true)
+
+		close(done)
+
+		// If the context was canceled during execution, return immediately.
+		if this.Context.Err() != nil {
+			return nil
+		}
```
eriq-augustine, line 323: You can also do this outside of the conditional right before the final return.
Then you check at the last possible chance.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+	} else if this.ReturnCopy {
+		go func() {
+			backgroundOutput := &JobOutput[InputType, OutputType]{
+				RemainingItems: output.GetRemainingItems(),
+			}
+
+			this.run(backgroundOutput, false)
+
```
eriq-augustine, line 331: Remove this newline.
Careful not to separate a function call from its error check.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+	// that failed in RemainingItems.
+	WorkErrors map[int]error
+
+	// The list of results.
+	ResultItems []OutputType
+
+	// The list of tems that still need work.
+	RemainingItems []InputType
+
+	// The total computation time.
```
eriq-augustine, line 122: Expand this comment a bit to emphasize that this is only the total time spent in WorkFunc().

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+		close(done)
+
+		// If the context was canceled during execution, return immediately.
+		if this.Context.Err() != nil {
+			return nil
+		}
+	} else if this.ReturnCopy {
+		go func() {
+			backgroundOutput := &JobOutput[InputType, OutputType]{
+				RemainingItems: output.GetRemainingItems(),
```
eriq-augustine, line 327: If I'm not mistaken, receiving on a nil chan will block forever.
If this is true, then I think you should probably fill out this background object a little more.
I know we don't expect anyone to do anything with this background results,
but if it is not too much work we should make it so people can't really mess up with it.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+
+			close(done)
+		}()
+	}
+
+	return &output
+}
+
+// Given a partial job output, run job processes the remaining items in a thread safe manner.
+// When not updating the output, only the run time and errors will be available.
```
eriq-augustine, line 354: Wording here is off (just grammar).

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+
+	// The context has been canceled while waiting for a lock, abandon this job.
+	if this.Context.Err() != nil {
+		return
+	}
+
+	var err error = nil
+
+	// If we had to wait for the lock, then check again for stored records.
+	if !noLockWait && !this.OverwriteRecords && this.RetrieveFunc != nil {
```
eriq-augustine, line 374: I don't think it is necessary to check for `OverwriteRecords` here.
I think the only check you need is to delete records in the very beginning
(and then you have it on the very first retrieval, but that is just for performance).

The records got dropped, and therefore they were overwritten.
If another job happened be the one ti place the records after they were deleted,
that is just fine.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+			close(done)
+		}()
+	}
+
+	return &output
+}
+
+// Given a partial job output, run job processes the remaining items in a thread safe manner.
+// When not updating the output, only the run time and errors will be available.
+func (this *Job[InputType, OutputType]) run(output *JobOutput[InputType, OutputType], updateOutput bool) {
```
eriq-augustine, line 355: Comment on why `updateOutput` exists (just a sentence is fine).

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+
+			if found {
+				return PoolResult{workItem, result, 0, nil}, nil
+			}
+		}
+
+		// Nothing cached, compute the job.
+		startTime := timestamp.Now()
+
+		var errs error = err
```
eriq-augustine, line 436: Just a few lines above, you have an error that you didn't join.
I don't see why you need to start joining errors here.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+		if config.UNIT_TESTING_MODE.Get() {
+			runTime = 1
+		}
+
+		// Store the result.
+		if !this.DryRun && this.Context.Err() == nil && this.StoreFunc != nil {
+			err = this.StoreFunc([]OutputType{result})
+			if err != nil {
+				errs = errors.Join(errs, fmt.Errorf("Failed to store result for item '%v': '%w'.", workItem, err))
+			}
```
eriq-augustine, line 454: You could be trying to store an error result.

I have a comment above about this, but I think just not joining the errors and handling them upfront could be the answer to this.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+		result, err := this.WorkFunc(workItem)
+		if err != nil {
+			errs = errors.Join(errs, fmt.Errorf("Failed to perform individual work on item %v: '%w'.", workItem, err))
+		}
+
+		runTime := (timestamp.Now() - startTime).ToMSecs()
+		// Standardize the run time so tests can check for exact matches.
+		if config.UNIT_TESTING_MODE.Get() {
+			runTime = 1
+		}
```
eriq-augustine, line 447: I'm not a fan of this.
I know our test engines return a constant runtime (in the current main branch),
but do we really need to do this here?
The tests can probably just zero out/ignore the runtime.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,497 @@
...
+		output.Error = fmt.Errorf("Failed to run job in a parallel pool: '%w'.", err)
+		return
+	}
+
+	// If the job was canceled, exit right away.
+	if this.Context.Err() != nil {
+		return
+	}
+
+	lockmanager.Lock(output.JobID)
```
eriq-augustine, line 470: !

You're playing a dangerous game here.
If possible, always defer the unlock right after the lock.

I know this case is fine, but it would be so easy for a future dev to come along and be like "oh, there is a special case in this loop that we need to return" then BAM... deadlock.

eriq-augustine (changes requested):
Getting closer.

I still only made comments on the core job manager.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,368 @@
...
+
+	// If the job was canceled, exit right away.
+	if this.Context.Err() != nil {
+		return
+	}
+
+	output.RemainingItems = []InputType{}
+
+	for _, poolResult := range poolResults {
+		if poolResult.Error != nil {
```
eriq-augustine, line 347: These error semantics seem off to me.
 - Why are work errors merged with system errors? I could instead see like a single error that is like "there were work errors", then people can check for any errors, but not have system errors polluted.
 - Putting error items in remaining items seems a bit strange to me (I could see either leaving them out or having another field for them).

eriq-augustine (changes requested):
Looking pretty good.

I only reviewed the core code.
We should be able to go over it all in meeting tomorrow.

eriq-augustine (changes requested):
The code is looking pretty much complete.
There are some more semantics to go over, but nothing conceptual.
(Note that it's super common to keep running into discussions on semantics for large changes like this, since improvements in one area show discussion topics in other other areas.)

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	return nil
+}
+
+// Given a customized Job, Job.Run() processes input items in a parallel pool of workers.
+// Returns the collected results in a JobOutput.
+// When not waiting for completion, the JobOutput may still be modified until the Done channel is closed.
+// Returning a copy of results gives immediate access to stable results but the final results will not be accessible later.
+// Returns nil if the context is canceled during execution.
```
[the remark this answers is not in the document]
Lucas-Ellenberger: To get any of the partial results from a context cancellation, we would have to modify `util.RunParallelPoolMap()`. It currently returns (nil, nil, nil), so Job.run() cannot determine the amount of work done: [https://github.com/edulinq/autograder-server/blob/1dac81e39da8fab86280f014e61b210227c51fa9/internal/util/pool.go#L102-L104](https://github.com/edulinq/autograder-server/blob/1dac81e39da8fab86280f014e61b210227c51fa9/internal/util/pool.go#L102-L104)

Are we fine with updating these semantics?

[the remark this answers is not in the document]
eriq-augustine: That's up to you.

internal/util/pool_test.go
```diff
@@ -119,19 +135,44 @@ func TestRunParallelPoolMapCancel(test *testing.T) {
 			cancelFunc()
 		}()
 
-		actual, workErrors, err := RunParallelPoolMap(testCase.numThreads, input, ctx, workFunc)
+		output, err := RunParallelPoolMap(testCase.numThreads, input, ctx, workFunc)
```
eriq-augustine, line 138: In not convinced that this test actually tests cancellation cleanly.
All the other tests may finish before the first one actually releases the wait group.

I think a more specific test (no test cases) may be better that uses wait groups to ensure that one piece of work is done, one is never started, and one is in progress when the cancel comes.
Maybe only use one thread.

eriq-augustine (changes requested):
I'm generally liking the changes to the parallel pool.
The maps and waiting for incomplete results is nice.
There are some things to polish there.

I only looked at parallel pool in this review.

internal/util/pool.go
```diff
@@ -6,105 +6,171 @@ import (
...
+	// Signals that the parallel pool was canceled during execution.
+	Canceled bool
+
+	// An internal done channel to signal the results can be accessed.
+	done chan any
+}
+
+// Use this method to block until the results can be accessed.
+func (this PoolResult[InputType, OutputType]) IsDone() {
+	if this.done != nil {
```
eriq-augustine, line 27: How can the done chan be nil?

internal/util/pool.go
```diff
@@ -6,105 +6,171 @@ import (
...
+		<-this.done
+	}
+}
+
 // Do a map function (one result for one input) with a parallel pool of workers.
 // Unless there is a critical error (the final return value) or cancellation,
-// the output will have the same length as and index-match the input, but will have an empty value if there is an error.
-// A cancellation will return (nil, nil, nil).
-// Consult the returned map of errors using the item's index to check for item-level errors.
+// every input will be in PoolResult.Results.
```
eriq-augustine, line 34: This seems strange.
Why are we putting errors in the results?
It seems like it should be either Results or WorkErrors.

internal/util/pool.go
```diff
@@ -6,105 +6,171 @@ import (
...
-func RunParallelPoolMap[InputType any, OutputType any](poolSize int, workItems []InputType, ctx context.Context, workFunc func(InputType) (OutputType, error)) ([]OutputType, map[int]error, error) {
+func RunParallelPoolMap[InputType comparable, OutputType any](poolSize int, workItems []InputType, ctx context.Context, workFunc func(InputType) (OutputType, error)) (PoolResult[InputType, OutputType], error) {
 	if poolSize <= 0 {
-		return nil, nil, fmt.Errorf("Pool size must be positive, got %d.", poolSize)
-	}
-
-	type WorkItem struct {
-		Index int
-		Item  InputType
+		return PoolResult[InputType, OutputType]{}, fmt.Errorf("Pool size must be positive, got %d.", poolSize)
```
eriq-augustine, line 42: Why not nil?
(And have the function return a pointer?)
I'm generally not a fan of default non-trivial objects.

internal/util/pool.go
```diff
@@ -6,105 +6,171 @@ import (
...
-
 	resultQueue := make(chan ResultItem, poolSize)
-	workQueue := make(chan WorkItem, poolSize)
-	doneChan := make(chan bool, poolSize)
-	exitWaitGroup := sync.WaitGroup{}
+	workQueue := make(chan InputType, poolSize)
+	doneChan := make(chan any)
+	doneWorkingChan := make(chan any)
+
+	workerExitWaitGroup := sync.WaitGroup{}
```
eriq-augustine, line 56: Pair the creation of each wait group with populating it (Add()).

internal/util/pool.go
```diff
@@ -6,105 +6,171 @@ import (
...
 	}
 
-	results := make([]OutputType, len(workItems))
-	workErrors := make(map[int]error)
-
 	resultQueue := make(chan ResultItem, poolSize)
-	workQueue := make(chan WorkItem, poolSize)
-	doneChan := make(chan bool, poolSize)
-	exitWaitGroup := sync.WaitGroup{}
+	workQueue := make(chan InputType, poolSize)
```
eriq-augustine, line 52: With the new chans, you may want to put a comment on each chat/waitgroup saying what it is used for.
Technically the name should be sufficient, but you are on the border here
(especially with "doneChan" vs "doneWorkingChan").

internal/util/pool.go
```diff
@@ -6,105 +6,171 @@ import (
...
 			case <-ctx.Done():
 				return
-			case workQueue <- WorkItem{i, item}:
+			case workQueue <- item:
 				// Item was already sent on the chan, do nothing here.
 			}
 		}
+
+		// Signal to the workers that there is no more work.
+		close(workQueue)
```
eriq-augustine, line 86: This is not hit in the cancellation case.
What about putting it in a defer with `threadExitWaitGroup.Done()`?

internal/util/pool.go
```diff
@@ -6,105 +6,171 @@ import (
...
 				select {
 				case <-ctx.Done():
 					return
-				case workItem := <-workQueue:
-					result, err := workFunc(workItem.Item)
-					resultQueue <- ResultItem{workItem.Index, result, err}
-				case <-doneChan:
-					return
+				case workItem, ok := <-workQueue:
+					if !ok {
```
eriq-augustine, line 133: Note that this is because the work queue was closed.

internal/util/pool_test.go
```diff
@@ -57,82 +63,114 @@ func TestRunParallelPoolMapBase(test *testing.T) {
...
-			return len(input), nil
+	// Block until the cancellation is called.
+	cancelWaitGroup := sync.WaitGroup{}
+	cancelWaitGroup.Add(1)
+
+	// Work will not start on input 'CCC'.
+	workFunc := func(input string) (int, error) {
+		// Signal on the second piece of work so that work on input 'A' will complete normally.
+		if input == "BB" {
+			workWaitGroup.Done()
```
eriq-augustine, line 123: Instead of this, can we just cancel ourselves?
(Like just make the cancel context earlier and call `cancelFunc()`?

eriq-augustine (changes requested):
Parallel pool is looking pretty good,
just some smaller things.

But the bigger thing is that the PR is too big now.
We'll need to break it into three PRs:
 - Pool
 - Job Manager
 - Analysis

Lucas-Ellenberger:
This PR grew too large and was split into three PRs that have been merged:
 - https://github.com/edulinq/autograder-server/commit/84ef1c55e7a13dd23b37cd7e71d2b02ff40c531b refactoring the parallel pool utility function
 - https://github.com/edulinq/autograder-server/commit/523f5270eb69dbcdf72d9ae3f596d92dd72756c2 creating the job manager package
 - https://github.com/edulinq/autograder-server/commit/b370ce694a75894ee34ce6e37a3cff9fdbb05195 refactoring analysis to utilize the job manager
 
 So, this reference PR is no longer needed.

PR #175, opened 2025-05-07
Added API endpoint for server heartbeat.

maozblan:
This PR adds an API endpoint `metadata/heartbeat` that returns the current server version when called in a GET request.

### DIFF
--- a/internal/api/metadata/heartbeat_test.go
+++ b/internal/api/metadata/heartbeat_test.go
@@ -18,10 +18,13 @@ func TestHeartbeat(test *testing.T) {
 
 	expected, err := util.GetFullCachedVersion()
 	if err != nil {
-		test.Fatalf("Failed to get expected server version: %v", err)
-		return
+		test.Fatalf("Failed to get expected server version: '%v'.", err)
 	}
+
 	if heartbeatResponse.ServerVersion != expected {
-		test.Fatalf("Server version mismatch. Expected %s, got %s", expected, heartbeatResponse.ServerVersion)
+		test.Fatalf("Server version mismatch. Expected '%s', Actual '%s'.",
+			util.MustToJSONIndent(expected),
+			util.MustToJSONIndent(heartbeatResponse.ServerVersion),
+		)
 	}
 }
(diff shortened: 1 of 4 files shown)


eriq-augustine (changes requested):
The code itself is looking good,
but there is a core issue here.
You should be failing the tests (`./scripts/run_tests.sh`) locally,
and you are definitely [failing in CI](
https://github.com/edulinq/autograder-server/actions/runs/14872477069/job/41763983881#step:8:82).

Generally, tests should not be failing for code you want to get in.
(Of course, there are always exceptions for things like code still under development.)
The specific test you are failing (TestDescribeRoutesFull) is because the API [description on disk](https://github.com/edulinq/autograder-server/blob/main/resources/api.json) does not match the one generated by the API.
Since you added an endpoint, this is not strange.
We have a script to regenerate the API description:
https://github.com/edulinq/autograder-server/blob/main/scripts/update_api_description.sh
As the endpoint developer, it is your job to verify that the description matches your expectations.
(We purposefully put this test in so that people have an extra step when making (possibly breaking) changes to the API.)

internal/api/metadata/heartbeat_test.go
```diff
@@ -0,0 +1,30 @@
...
+	var heartbeatResponse HeartbeatResponse
+	util.MustJSONFromString(util.MustToJSON(response.Content), &heartbeatResponse)
+
+	expected, err := util.GetFullCachedVersion()
+	if err != nil {
+		test.Fatalf("Failed to get expected server version: '%v'.", err)
+	}
+
+	if heartbeatResponse.ServerVersion != expected {
+		test.Fatalf("Server version mismatch. Expected '%s', Actual '%s'.",
```
eriq-augustine, line 25: s/Expected/Expected:/
s/Actual/Actual:/

eriq-augustine (changes requested):
Just one small nit.

eriq-augustine (approved):
lgtm

I'll hold on to this until the Python side get approved.

PR #176, opened 2025-05-09
Refactored the parallel pool utility to better handle context cancell…

Lucas-Ellenberger:
…ations.

This PR improves the parallel pool utility function to better handle context cancellations. The new implementation still returns partial results quickly upon cancellation. But, the new implementation provides the method `IsDone()` which lets the calling function optionally wait to collect all of the partial results before the cancellation. So, calling functions can wait and clean up and potential artifacts from a partial job, if desired.

The analysis functions are updated to fit into the new parallel pool model, which affects analysis from the model layer to the api layer.

These changes are a stepping stone for the larger [job manager PR](https://github.com/edulinq/autograder-server/pull/174), which has grown too large.

### DIFF
--- a/internal/model/analysis.go
+++ b/internal/model/analysis.go
@@ -1,6 +1,7 @@
 package model
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"regexp"
@@ -123,6 +124,9 @@ type PairwiseAnalysisSummary struct {
 	AggregateTotalMeanSimilarities util.AggregateValues            `json:"aggregate-total-mean-similarity"`
 }
 
+// A named map type allows for custom marshalling and unmarshalling.
+type PairwiseAnalysisMap map[PairwiseKey]*PairwiseAnalysis
+
 func NewPairwiseKey(fullSubmissionID1 string, fullSubmissionID2 string) PairwiseKey {
 	return PairwiseKey([2]string{
 		min(fullSubmissionID1, fullSubmissionID2),
@@ -213,6 +217,41 @@ func (this *PairwiseKey) Course() string {
 	return courseID
 }
 
+func (this PairwiseAnalysisMap) MarshalJSON() ([]byte, error) {
+	temp := make(map[string]*PairwiseAnalysis, len(this))
+
+	for key, analysis := range this {
+		keyString := key.String()
+
+		temp[keyString] = analysis
+	}
+
+	return json.Marshal(temp)
+}
+
+func (this *PairwiseAnalysisMap) UnmarshalJSON(data []byte) error {
+	temp := make(map[string]*PairwiseAnalysis)
+	err := json.Unmarshal(data, &temp)
+	if err != nil {
+		return err
+	}
+
+	results := make(PairwiseAnalysisMap, len(temp))
+	for text, analysis := range temp {
+		keyParts := strings.Split(text, PAIRWISE_KEY_DELIM)
+		if len(keyParts) != 2 {
+			return fmt.Errorf("Invalid PairwiseKey: '%s'.", text)
+		}
+
+		pairwiseKey := NewPairwiseKey(keyParts[0], keyParts[1])
+		results[pairwiseKey] = analysis
+	}
+
+	*this = results
+
+	return nil
+}
+
 func NewPairwiseAnalysis(pairwiseKey PairwiseKey, assignment *Assignment, similarities map[string][]*FileSimilarity, unmatches [][2]string, skipped []string) *PairwiseAnalysis {
 	meanSimilarities := make(map[string]float64, len(similarities))
 	totalMeanSimilarity := 0.0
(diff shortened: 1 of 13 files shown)


internal/util/pool.go
```diff
@@ -6,105 +6,175 @@ import (
...
-	exitWaitGroup := sync.WaitGroup{}
+	workQueue := make(chan InputType, poolSize)
+
+	// The done working channel is used by the pool workers to signal work is complete to the result collectors.
+	doneWorkingChan := make(chan any)
+	// The done channel is used to signal that the all results are collected and accessible.
+	// It also waits to signal until all worker threads join the main thread.
+	doneChan := make(chan any)
+
+	// A wait group used to determine when the workers have completed there work.
```
eriq-augustine, line 58: s/there/their/

internal/util/pool.go
```diff
@@ -6,105 +6,175 @@ import (
...
+
+			// All workers completed, drain remaining results.
+			// If the context was canceled, the number of results may be less than the number of work items.
+			for {
+				select {
+				case resultItem := <-resultQueue:
+					output.Results[resultItem.Input] = resultItem.Result
+					if resultItem.Error != nil {
+						output.WorkErrors[resultItem.Input] = resultItem.Error
+					}
```
eriq-augustine, line 105: This is inconsistent with the function-level comment:
> // every input will either be in PoolResult.Results or PoolResult.WorkErrors.

Just use if/else here and in the primary collection to make sure only one gets populated for each item.

internal/util/pool.go
```diff
@@ -6,105 +6,175 @@ import (
...
-
 	resultQueue := make(chan ResultItem, poolSize)
-	workQueue := make(chan WorkItem, poolSize)
-	doneChan := make(chan bool, poolSize)
-	exitWaitGroup := sync.WaitGroup{}
+	workQueue := make(chan InputType, poolSize)
+
+	// The done working channel is used by the pool workers to signal work is complete to the result collectors.
+	doneWorkingChan := make(chan any)
+	// The done channel is used to signal that the all results are collected and accessible.
```
eriq-augustine, line 54: I would put a newline here to separate the two.
Although it can happen, it is less common to have comment within a visual block.

(Note that this is purely a nit, since there are other changes.)

internal/util/pool.go
```diff
@@ -6,105 +6,175 @@ import (
...
 		}()
 	}
 
 	// Wait on the wait group in the background so we can select between it and the context.
 	// Technically we could wait on the done channel, but this will ensure that all workers have exited.
-	exitWaitGroupChan := make(chan bool, 1)
 	go func() {
-		exitWaitGroup.Wait()
-		exitWaitGroupChan <- true
+		// Once all pool workers have completed there work, signal to collect the remaining results.
```
eriq-augustine, line 157: s/there/their/

internal/model/analysis.go
```diff
@@ -213,6 +217,41 @@ func (this *PairwiseKey) Course() string {
...
+		}
+
+		pairwiseKey := NewPairwiseKey(keyParts[0], keyParts[1])
+		results[pairwiseKey] = analysis
+	}
+
+	*this = results
+
+	return nil
+}
```
eriq-augustine, line 253: I think you can instead get away with implementing `encoding.TextMarshaler` since this is for map keys,
and that should be easier to implement than raw JSON marshaling.
See https://pkg.go.dev/encoding/json#Marshal
(search for TextMarshaler in this section).

internal/model/analysis.go
```diff
@@ -294,19 +333,20 @@ func NewIndividualAnalysisSummary(results []*IndividualAnalysis, pendingCount in
...
 
-	for i, result := range results {
+	for _, result := range results {
 		if result.Failure {
 			failureCount++
 			continue
 		}
 
-		if (i == 0) || (result.AnalysisTimestamp < firstTimestamp) {
+		if firstItem || (result.AnalysisTimestamp < firstTimestamp) {
```
eriq-augustine, line 345: You can also use `firstTimestamp.IsZero()` instead of `firstItem`.

internal/model/analysis.go
```diff
@@ -123,6 +124,9 @@ type PairwiseAnalysisSummary struct {
 	AggregateTotalMeanSimilarities util.AggregateValues            `json:"aggregate-total-mean-similarity"`
 }
 
+// A named map type allows for custom marshalling and unmarshalling.
+type PairwiseAnalysisMap map[PairwiseKey]*PairwiseAnalysis
```
eriq-augustine, line 128: Since we need the marshaling for this now (even with the below comment),
make a real quick test for it (just marshal and back).

eriq-augustine (changes requested):
Looking REALLY good.
The core concepts here are all clear and robust.

Since there will be several chained changes coming in, I don't think you need to worry about updating the Python side until the analysis changes are in.
Since they Python side is no pegged to a specific (docker image) version and not the latest, this should be safe.

internal/model/analysis_test.go
```diff
@@ -10,6 +10,32 @@ import (
...
+			NewPairwiseKey("failed", "analysis"): NewFailedPairwiseAnalysis(NewPairwiseKey("failed", "analysis"), nil, ""),
+		},
+	}
+
+	for _, testCase := range testCases {
+		pairwiseAnalysisJSON := util.MustToJSON(testCase)
+
+		var pairwiseAnalysisResult map[PairwiseKey]*PairwiseAnalysis
+		util.MustJSONFromString(pairwiseAnalysisJSON, &pairwiseAnalysisResult)
+	}
```
eriq-augustine, line 36: Compare `testCase` and `pairwiseAnalysisResult` for equality.

eriq-augustine (changes requested):
Looking real good.
Just one testing addition.

eriq-augustine (approved):
lgtm

PR #177, opened 2025-05-09
Added the jobmanager package to help structure running large jobs.

Lucas-Ellenberger:
This PR adds the `jobmanager` package which provides common infrastructure for running large jobs.

The `jobmanager` allows for flexible configuration through the `Job` struct. It allows for synchronization (at the job and work item level), context cancellation, record retrieval, record storage, record removal, and asynchronous output processing. By adding a layer of abstraction, large operations that require some or all of these features can be more easily defined and run.

This PR builds off of the recent [changes to the parallel pool](https://github.com/edulinq/autograder-server/commit/84ef1c55e7a13dd23b37cd7e71d2b02ff40c531b). This PR is the next step in completing the job manager PR that grew too large, which can be found [here](https://github.com/edulinq/autograder-server/pull/174).

Next, individual and pairwise analysis will be routed through the job manager to streamline the analysis code.

### DIFF
--- a/internal/jobmanager/jobmanager.go
+++ b/internal/jobmanager/jobmanager.go
@@ -10,8 +10,7 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
-// JobOptions contains user level options for a Job.
-// These options allow for optional context cancellation of the Job.
+// The user level options for a job, including an optional context.
 type JobOptions struct {
 	// Don't save anything.
 	DryRun bool `json:"dry-run"`
@@ -26,26 +25,15 @@ type JobOptions struct {
 	Context context.Context `json:"-"`
 }
 
-// Job provides system level customization of the job's execution.
-// It supports the following optional functionality:
-//   - synchronization
-//   - context cancellation
-//   - record retrieval
-//   - record storage
-//   - record removal from storage
-//   - asynchronous output processing
-//
+// Provides system level customization of the job's execution.
 // Given job options, input items, and a work function,
-// Job will invoke the work func on each input item in a parallel pool to produce a JobOutput.
-// If a lock key is provided, Job will block on that key before running.
-// The locking behavior prevents overuse of resources and conflicts between the same Job.
-// Provide an input key generation function to synchronize at the input item level in addition to the job level.
+// a job will invoke the work func on each input item in a parallel pool.
 type Job[InputType comparable, OutputType any] struct {
 	// The user level options for a Job.
 	*JobOptions
 
 	// Instead of adding to the results as they are processed,
-	// return a copy of the potentially incomplete results fetched before the job starts.
+	// return a copy of the potentially incomplete results fetched before the job starts, e.g. from a cache.
 	// If WaitForCompletion is true, this option is ignored.
 	ReturnIncompleteResults bool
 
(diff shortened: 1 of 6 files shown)


internal/jobmanager/jobmanager_test.go
```diff
@@ -0,0 +1,1031 @@
...
+	"A",
+	"BB",
+	"CCC",
+}
+
+var workFunc = func(input string) (int, error) {
+	return len(input), nil
+}
+
+type printableJob[InputType any, OutputType any] struct {
```
eriq-augustine, line 28: Can you just get away with this by putting the JSON tags on a normal job (and omitting `"-"`) the funcs?

eriq-augustine (changes requested):
Code looking real good.

I made a lot of comments on comments.
I stopped after a few, so take a pass over the comments with my comments in mind.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,393 @@
...
+	OverwriteRecords bool `json:"overwrite-records"`
+
+	// Wait for the entire job to complete and return all results.
+	WaitForCompletion bool `json:"wait-for-completion"`
+
+	// A context that can be used to cancel the job.
+	Context context.Context `json:"-"`
+}
+
+// Provides system level customization of the job's execution.
```
eriq-augustine, line 28: s/system level/system-level/

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,393 @@
...
+
+	// An optional function to process the final JobOutput upon successful completion.
+	OnSuccess func(JobOutput[InputType, OutputType]) `json:"-"`
+}
+
+// The output from running a job.
+// A critical error (JobOutput.Error) may not generate work errors, so both must be checked independently.
+// Consult the map of errors using the item to check for item-level errors.
+// Always wait for the Done channel to be closed before handling the output,
+// which can be achieved by calling JobOutput.IsDone().
```
eriq-augustine, line 78: s/can be/can also be/

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,393 @@
...
+
+	return nil
+}
+
+// Process input items in a parallel pool of workers and return the collected results.
+// When not waiting for completion, the output may still be modified until the Done channel is closed.
+// Returning a copy of results gives immediate access to stable results but the final results will be inaccessible.
+// If the context is canceled during execution, the output will signal accordingly.
+func (this *Job[InputType, OutputType]) Run() *JobOutput[InputType, OutputType] {
+	// Run closes the channel to signal the returned output is safe to handle.
```
eriq-augustine, line 143: s/Run/Run()/

Suffixing function names with parens makes it easier to identify them.

internal/jobmanager/jobmanager.go
```diff
@@ -0,0 +1,393 @@
...
+	return &output
+}
+
+// Process the remaining items and update the partial job output.
+// If the results will be inaccessible, the output will not be updated to reduce memory usage.
+// However, the run time and error will be updated for stats and logging purposes.
+// The parameter updateOutput signals whether to add results to the output or not.
+func (this *Job[InputType, OutputType]) run(output *JobOutput[InputType, OutputType], updateOutput bool) {
+	if len(output.RemainingItems) == 0 {
+		return
```
eriq-augustine, line 232: I think this would be considered a successful run, so this.OnSuccess should probably be called.
I think this is also the case with line 266.

You may want to add a method to Job (runComplete()?) where you can check if this.Error is nil (and if this.OnSucccess is not nil), and then call this.OnSuccess.
Then you can just defer it as the first thing in this method.
(And in the future if we have a function to call on error, and one to call on completion (like a finally), then this would be a great place to do it.)

---
On a side note, notice how I used "this.OnSuccess", but I just said in an earlier comment to suffix function names with "()".
But, this.OnSuccess is a variable that contains a function and not an actual function.
(You may have been wondering this, or you may have already figured it out.)

eriq-augustine (changes requested):
Looking pretty good!
Just one semantic thing and a couple nits.

internal/jobmanager/jobmanager.go
```diff
@@ -283,7 +285,20 @@ func (this *Job[InputType, OutputType]) run(output *JobOutput[InputType, OutputT
...
+}
+
+// Perform optional work on the final output.
+func (this *Job[InputType, OutputType]) runComplete(output *JobOutput[InputType, OutputType]) {
+	// Encountered a cancellation, return immediately.
+	if output.Canceled {
+		return
+	}
+
+	if output.Error != nil || len(output.WorkErrors) != 0 {
```
eriq-augustine, line 297: The `len(output.WorkErrors) != 0` semantics are a little difficult.
Is it a success if all but one just run fine?
I think your initial decision here makes sense (and there are multiple correct answers).
But, your answer to the above is "no, something failed".
But, I believe the idea for analysis was to have them log stats in this function
(which we want to do even if some jobs fail).

So... what to do?

I believe the possible states on completion are:
 - Hard/System Error
 - Success with Work Error(s)
 - Cancel
 - Cancel (with Work Error(s))
 - Clean Success

Making functions for the desired subsets of these may be a bit much.

_One of several possibilities_:  
We can take inspiration [from promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#instance_methods) and simplify possible functions to: success, error, and finally (names are just placeholders/concepts).
(We don't have to implement them all now.)
If we do that, then I think it would be best to leave the work errors as "soft errors" and only consider hard errors an error.

Breaking down the previous cases with this mindset would be:
 - Hard/System Error -- Calls: [error, finally]
 - Success with Work Error(s) -- Calls: [success, finally]
 - Cancel -- Calls: [finally]
 - Cancel (with Work Error(s)) -- Calls: [finally]
 - Clean Success -- Calls: [success, finally]

eriq-augustine (changes requested):
Just one semantic thing.

eriq-augustine (approved):
LGTM!

PR #178, opened 2025-05-12
Refactored analysis to utilize the job manager infrastructure.

Lucas-Ellenberger:
This PR refactors the analysis package to leverage the new jobmanager infrastructure. The refactoring greatly reduces the complexity of analysis via the jobmanager abstraction.

These changes further emphasize the efficacy of the jobmanager package's ability to run large jobs in a flexible manner. Since this PR is a refactoring, the results of running analysis remain unchanged.

This PR is the final PR in the trio of PRs to replace the outdated and unmanageable [original jobmanager PR](https://github.com/edulinq/autograder-server/pull/174).

### DIFF
--- a/internal/analysis/individual.go
+++ b/internal/analysis/individual.go
@@ -10,6 +10,7 @@ import (
 	"github.com/edulinq/autograder/internal/config"
 	"github.com/edulinq/autograder/internal/db"
 	"github.com/edulinq/autograder/internal/jobmanager"
+	"github.com/edulinq/autograder/internal/log"
 	"github.com/edulinq/autograder/internal/model"
 	"github.com/edulinq/autograder/internal/timestamp"
 	"github.com/edulinq/autograder/internal/util"
@@ -72,7 +73,11 @@ func IndividualAnalysis(options AnalysisOptions) (map[string]*model.IndividualAn
 	}
 
 	if len(output.WorkErrors) != 0 {
-		return nil, 0, fmt.Errorf("Failed to run individual analysis for '%d' submissions.", len(output.WorkErrors))
+		for submissionID, err := range output.WorkErrors {
+			log.Error("Failed to run individual analysis.", err, log.NewAttr("submission-id", submissionID), log.NewAttr("job-id", output.ID))
+		}
+
+		return nil, 0, fmt.Errorf("Failed to run individual analysis for '%d' submissions during job '%s'.", len(output.WorkErrors), output.ID)
 	}
 
 	return output.ResultItems, len(output.RemainingItems), nil
--- a/internal/analysis/individual.go
+++ b/internal/analysis/individual.go
@@ -216,6 +216,8 @@ func collectIndividualStats(fullSubmissionIDs []string, totalRunTime int64, init
 	collectAnalysisStats(fullSubmissionIDs, totalRunTime, initiatorEmail, "individual")
 }
 
+// Best effort attempt to convert a submission ID into fine grain log attributes.
+// Returns a slice of any instead of a slice of *log.Attr due to type constraints of the logging functions.
 func submissionIDToLogValues(fullSubmissionID string) []any {
 	logAttributes := make([]any, 0)
 
--- a/internal/analysis/individual.go
+++ b/internal/analysis/individual.go
@@ -79,9 +79,8 @@ func IndividualAnalysis(options AnalysisOptions) (map[string]*model.IndividualAn
 			workErrors[fullSubmissionID] = err.Error()
 
 			logAttributes := submissionIDToLogValues(fullSubmissionID)
-
 			logAttributes = append([]any{err}, logAttributes...)
-
+			logAttributes = append(logAttributes, log.NewAttr("job-id", output.ID))
 			log.Error("Failed to run individual analysis.", logAttributes...)
 		}
 	}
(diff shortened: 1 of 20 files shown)


eriq-augustine (changes requested):
Looking pretty good, just one semantics things.

internal/analysis/individual.go
```diff
@@ -72,7 +73,11 @@ func IndividualAnalysis(options AnalysisOptions) (map[string]*model.IndividualAn
 	}
 
 	if len(output.WorkErrors) != 0 {
-		return nil, 0, fmt.Errorf("Failed to run individual analysis for '%d' submissions.", len(output.WorkErrors))
+		for submissionID, err := range output.WorkErrors {
+			log.Error("Failed to run individual analysis.", err, log.NewAttr("submission-id", submissionID), log.NewAttr("job-id", output.ID))
+		}
+
+		return nil, 0, fmt.Errorf("Failed to run individual analysis for '%d' submissions during job '%s'.", len(output.WorkErrors), output.ID)
```
eriq-augustine, line 80: We usually don't quote ints.
(Sometimes we do, just not normally.)

internal/analysis/pairwise.go
```diff
@@ -105,7 +105,11 @@ func PairwiseAnalysis(options AnalysisOptions) (map[model.PairwiseKey]*model.Pai
 	}
 
 	if len(output.WorkErrors) != 0 {
-		return nil, 0, fmt.Errorf("Failed to run pairwise analysis for '%d' submissions.", len(output.WorkErrors))
+		for pairwiseKey, err := range output.WorkErrors {
+			log.Error("Failed to run pairwise analysis.", err, log.NewAttr("pairwise-key", pairwiseKey.String()), log.NewAttr("job-id", output.ID))
```
eriq-augustine, line 109: Logging the components here is a bit more tricky, since there could be multiple elements.
You may need to see if there are multiple keys, sort them, use the first key as the primary thing (course, assignment, user), and then put the second course in some alternate key (e.g., "alt-course"?).

Since this is complex, I would probably stash it away in a LogValue() method:
https://github.com/edulinq/autograder-server/blob/main/internal/log/attribute.go#L20

eriq-augustine (changes requested):
The ids are looking pretty good.
We need some more fine-grained logging support.

internal/analysis/individual.go
```diff
@@ -20,220 +19,76 @@ import (
...
-		if !ok {
-			errs = errors.Join(errs, fmt.Errorf("Unable to find result for submission: '%s'.", id))
-			continue
-		}
+	if len(output.WorkErrors) != 0 {
+		for fullSubmissionID, err := range output.WorkErrors {
+			courseID, assignmentID, email, shortSubmissionID, splitErr := common.SplitFullSubmissionID(fullSubmissionID)
+			if splitErr != nil {
+				log.Error("Failed to split full submission ID.", splitErr, log.NewAttr("submission-id", fullSubmissionID))
+				log.Error("Failed to run individual analysis.", err, log.NewAttr("submission-id", fullSubmissionID), log.NewAttr("job-id", output.ID))
```
eriq-augustine, line 81: This looks a tad messy (not because of you, just the nature of the situation).

Would it be possible to make a function that returns `[]*log.Attr` and then splat/unpack (`...`) them into the log statement?

eriq-augustine (changes requested):
Just a few more things.

internal/analysis/individual.go
```diff
@@ -19,221 +18,75 @@ import (
...
-		return nil, 0, fmt.Errorf("Failed to compute individual analysis for '%s': '%w'.", fullSubmissionID, err)
-	}
+			logAttributes = append([]any{err}, logAttributes...)
 
-	// Store the result.
-	if !options.DryRun && (options.Context.Err() == nil) {
-		err = db.StoreIndividualAnalysis([]*model.IndividualAnalysis{result})
-		if err != nil {
-			return nil, 0, fmt.Errorf("Failed to store individual analysis for '%s' in DB: '%w'.", fullSubmissionID, err)
+			log.Error("Failed to run individual analysis.", logAttributes...)
```
eriq-augustine, line 85: You log attributes are missing the job id.
(Pairwise already has this.)

eriq-augustine, line 85: Just a possible nit.

I would put the two `logAttributes` together (no newline between them).
You can also put the log.Error() in the same visual block.

internal/api/courses/assignments/submissions/analysis/individual.go
```diff
@@ -16,10 +16,12 @@ type IndividualRequest struct {
 }
 
 type IndividualResponse struct {
-	Complete bool                                 `json:"complete"`
-	Options  analysis.AnalysisOptions             `json:"options"`
-	Summary  *model.IndividualAnalysisSummary     `json:"summary"`
-	Results  map[string]*model.IndividualAnalysis `json:"results"`
+	Success    bool                                 `json:"success"`
```
eriq-augustine, line 19: `model.AnalysisSummary` already has a `FailureCount` (which is properly maintained).
I think we should add an `ErrorCount` to summarize work errors.

I think we should probably also drop `Success` here.
Feel free to disagree, but I think its just a bit too awkward in too many cases, e.g.,:
 - 1,000,000 jobs completed successfully, but one had a work error.
 - There were some failures.
 - There were 100% failures.
 - Nothing was run.

`Success` is not "wrong" in these cases, just awkward.

eriq-augustine (changes requested):
Just a couple things.

eriq-augustine (approved):
lgtm

PR #182, opened 2025-05-21
Removed type information for embedded types from the API description.

Lucas-Ellenberger:
This PR removes the type information for types embedded in the API request or response structs.

As these types are embedded, the custom type is not displayed in the API input or output fields. So, leaving the type information is unnecessary as users will never need these types.

By removing them, we reduce the amount of review needed when updating the API description information.


eriq-augustine (approved):
lgtm!

PR #183, opened 2025-05-22
Added course user references to provide a flexible targetting of users.

Lucas-Ellenberger:
This PR adds course user references which provide an easy, flexible way to target course users.

The input strings can be as follows:
 - An email literal
 - A course role (which targets all users in the course with that role)
 - A literal start `*` (which target all users in the course)
 - Any of the above options preceded by a dash `-` (which indicates that user or group will NOT be included in the final results)
 
Input ordering does not affect the outcome as exclusions always supersede inclusions (e.g. `[]{"-admin", "*"}` targets all users in the course that are not course admins).

The efficacy of this powerful course user targetting system is displayed by updating `courses/users/list`.


internal/db/user_reference.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/edulinq/autograder/internal/model"
+)
+
+// Process a list of user inputs in the context of a course.
+// See model.CourseUserReferenceInput for the list of acceptable inputs.
+// Inputs are parsed without checking permissions.
+// APIs using this function should require a core.MinCourseRoleGrader or custom permissions checking.
```
eriq-augustine, line 13: Requiring `MinCourseRoleGrader` is not set in stone.
Keep the comment a little more general, like:
"API endpoints using this function should require a sufficient role or ...".

internal/db/user_reference.go
```diff
@@ -0,0 +1,95 @@
...
+// Returns a reference, user errors, and a system error.
+// User-level errors return (partial reference, user errors, nil).
+// A system-level error returns (nil, nil, error).
+func ParseCourseUserReference(course *model.Course, rawReferences []model.CourseUserReferenceInput) (*model.CourseUserReference, map[string]error, error) {
```
eriq-augustine, line 17: I don't see too much of a purpose for these soft errors.
I know you were just in the swamps of the job manager where we always needed soft and hard errors,
but here all likely errors are already the caller's fault.

I think you should have two versions of this function:
 - One that queries the db for the course users and then call the next function.
 - One that resolves the references given a list of course users.

With the above pattern, `courses/users/list` can now keep `CourseUsers`
(let the API infrastructure fetch those),
and then any errors that happen during parsing references is the users fault
(which means we can give an http 400 instead of 500).

internal/db/user_reference.go
```diff
@@ -0,0 +1,95 @@
...
+		if strings.HasPrefix(reference, "-") {
+			exclude = true
+
+			reference = strings.TrimPrefix(reference, "-")
+			reference = strings.TrimSpace(reference)
+		}
+
+		if reference == "" {
+			continue
+		}
```
eriq-augustine, line 52: Move this before the exclusion check.
"-" should return an error (no need for you to be too nice to these jerks that can't format a user reference).

internal/db/user_reference.go
```diff
@@ -0,0 +1,95 @@
...
+			if exclude {
+				courseUserReference.ExcludeEmails[reference] = nil
+			} else {
+				courseUserReference.Emails[reference] = nil
+			}
+
+			continue
+		}
+
+		if reference == "*" {
```
eriq-augustine, line 70: I don't like these disjoint conditionals (since the cases are logically independent).

I know this pattern is fine with returns, but I feel like continues are weaker.

internal/db/user_reference.go
```diff
@@ -0,0 +1,95 @@
...
+		if reference == "" {
+			continue
+		}
+
+		if strings.Contains(reference, "@") {
+			_, ok := courseUsers[reference]
+			if !ok {
+				workErrors[string(rawReference)] = fmt.Errorf("Unknown course user: '%s'.", reference)
+				continue
+			}
```
eriq-augustine, line 59: You should probably be able to send to people outside the course.
It seems super normal to cc someone related to, but not enrolled in the course.
(I, for example, get course email all the time.)

internal/db/user_reference.go
eriq-augustine: The second function may even be able to live in `model`.
(Code in lower dependency levels is generally preferable, since it doesn't have to "know" about a lot of other things.)

eriq-augustine: Actually, another one of my comments calls into question the need for `courseUsers`.
So, you may not even need them or the first function.

internal/db/user_resolve.go
```diff
@@ -9,14 +9,62 @@ import (
 	"github.com/edulinq/autograder/internal/model"
 )
 
+// Returns a sorted list of users based on the course reference and an error.
+func ResolveCourseUsers(course *model.Course, reference *model.CourseUserReference) ([]*model.CourseUser, error) {
```
eriq-augustine, line 13: I think both the name and semantics here need to be changed slightly,
since one of my previous comments said that outside emails should probably be allowed.

internal/db/user_resolve.go
```diff
@@ -9,14 +9,62 @@ import (
...
+
 // Resolve course email addresses.
 // Take a course and a list of strings (containing emails specs) and returns a sorted slice of lowercase emails without duplicates.
 // An email spec can be:
 // an email address,
 // a course role (which will include all course users with that role),
 // a literal "*" (which includes all users enrolled in the course),
 // or an email address preceded by a dash ("-") (which indicates that this email address should NOT be included in the final results).
-func ResolveCourseUsers(course *model.Course, emails []string) ([]string, error) {
+func ResolveCourseUserEmails(course *model.Course, emails []string) ([]string, error) {
```
eriq-augustine, line 67: This function is probably real close to being removed.
So, you should probably make a function that resolves course user references to emails,
and make this function just wrap that one.

eriq-augustine (changes requested):
Pretty good start.
There are some design things to consider.

Also, make sure to update the documentation:
https://github.com/edulinq/autograder-server/tree/main/docs#course-email-specification-courseemailspec

internal/model/user_reference.go
```diff
@@ -0,0 +1,189 @@
...
+
+// Returns a sorted list of emails based on the course reference.
+// Emails can target users outside of the course.
+func ResolveCourseUserEmails(users map[string]*CourseUser, reference *ParsedCourseUserReference) []string {
+	if reference == nil {
+		return nil
+	}
+
+	emailSet := make(map[string]any, 0)
+	excludeSet := make(map[string]any, 0)
```
eriq-augustine, line 154: The semantics for the exclude set confuse me.

Let's say we have someone observing a course with an `other` role.
Wouldn't ['admin', 'owner', 'some-observer@test.edulinq.org'], exclude the observer?

I also don't see a test case for this situations (but I could have missed it).

docs/types.md
```diff
@@ -72,10 +72,11 @@ Underlying Type: String
 
 An email address.
 
-#### Course Email Specification (CourseEmailSpec)
+#### Course User Reference (CourseUserReference)
 
 When used in the context of a course,
-a `CourseEmailSpec` can be used to generalize email recipients.
+a `CourseUserReference` can be used to generalize targeting users
+(e.g, email recipients).
```
eriq-augustine, line 79: Because of this generalization, I would move this section up a level and put it above email (and then you can reference it in email).

internal/api/courses/admin/email.go
```diff
@@ -29,37 +32,32 @@ func HandleEmail(request *EmailRequest) (*EmailResponse, *core.APIError) {
 		return nil, core.NewBadRequestError("-627", request, "No email subject provided.")
 	}
 
-	var err error
-	var errs error
-
-	request.To, err = db.ResolveCourseUsers(request.Course, request.To)
-	errs = errors.Join(errs, err)
+	recipients, userErrors := request.CourseMessageRecipients.ToMessageRecipients(request.Users)
```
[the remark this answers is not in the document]
eriq-augustine: Unresolving a comment from my last pass.

internal/api/courses/users/list.go
```diff
@@ -2,27 +2,53 @@ package users
...
+
+	reference, userErrors := model.ParseCourseUserReferences(request.TargetUsers)
+
+	errors := make(map[string]string, len(userErrors))
+
+	for reference, err := range userErrors {
+		errors[reference] = err.Error()
+
+		log.Warn("Failed to parse user reference.", err, log.NewAttr("reference", reference))
+	}
```
eriq-augustine, line 37: As with courses/admin/email, I believe that the errors here are the users fault and just get an immediate 400.

internal/model/role_course.go
```diff
@@ -48,12 +57,14 @@ func GetCourseUserRoleString(role CourseUserRole) string {
 	return courseRoleToString[role]
 }
 
-func GetAllCourseUserRoles() map[CourseUserRole]string {
-	return courseRoleToString
-}
+func GetCommonCourseUserRolesCopy() map[string]CourseUserRole {
+	commonCourseUserRoleCopy := make(map[string]CourseUserRole, len(commonStringToCourseUserRole))
+
```
eriq-augustine, line 62: Nit.

I would remove this newline, so the visual block is just "make the copy".

eriq-augustine (changes requested):
Getting close.
I think there are still a few semantic things lurking.

internal/model/user_reference.go
```diff
@@ -0,0 +1,194 @@
...
```
eriq-augustine, line 159: Comment on why this exists, and give an example
(e.g., someone excluded based on role, but explicitly listed by email).
This is tricky because it also relies on the ordering of operations.

Remember you want to make sure that people have an easy time reading your code,
so when you do need to be crafty just comment it.

docs/types.md
```diff
@@ -66,16 +66,13 @@ The non-alphanumeric characters cannot be repeated in a sequence (e.g. you can h
...
-### Email
+### Course User Reference (CourseUserReference)
 
 Underlying Type: String
 
-An email address.
-
-#### Course Email Specification (CourseEmailSpec)
-
 When used in the context of a course,
```
eriq-augustine, line 73: A CourseUserReference should always be in the context of a course.

docs/types.md
```diff
@@ -66,16 +66,13 @@ The non-alphanumeric characters cannot be repeated in a sequence (e.g. you can h
...
-a `CourseEmailSpec` can be used to generalize email recipients.
+a `CourseUserReference` can be used to generalize targeting users
```
eriq-augustine, line 74: s/users/any number of course users/

docs/types.md
```diff
@@ -66,16 +66,13 @@ The non-alphanumeric characters cannot be repeated in a sequence (e.g. you can h
...
+(e.g, email recipients).
 The following values are allowed:
 
  - Email - Normal email addresses may be used.
```
eriq-augustine, line 78: Put a sub-bullet here that mentions how the exact treatment of non-course emails may different depending on the use case (they may be ignored, used normally, result in an error, etc).

docs/types.md
```diff
@@ -84,6 +81,16 @@ The following values are allowed:
  - Negative Email - An email address preceded by a minus sign (e.g., "-alice@test.edulinq.org")
    will remove this address from the email recipients (even if they are not currently there).
    This can be useful when using course roles but you want to exclude someone.
+ - Negative Course Role - A course role preceded by a minus sign (e.g., "-student")
+   will remove all course users with that role. This can be useful when using the "\*"
+   but you want to exclude a role.
```
eriq-augustine, line 86: The text in these bullets needs a smoothing pass.
They were fine when we were talking about emails, but now we are talking about users.
So, some of the phrasing is a little awkward.

For example, the first bullet: "Normal email addresses may be used.".
A reader could be like "wait... I thought we were talking about course users?".
Another way to phrase this bullet could be something like:
"The email of the requested user."

docs/types.md
```diff
@@ -84,6 +81,16 @@ The following values are allowed:
...
+
+### Email
+
+Underlying Type: String
+
+An email address.
+Some places that accept an email address use an underlying type of [CourseUserReference](#course-user-reference-courseuserreference).
```
eriq-augustine, line 93: s/use an underlying type of/use/

Note that "underlying type" would be incorrect here.

docs/types.md
```diff
@@ -84,6 +81,16 @@ The following values are allowed:
...
    This can be useful when using course roles but you want to exclude someone.
+ - Negative Course Role - A course role preceded by a minus sign (e.g., "-student")
+   will remove all course users with that role. This can be useful when using the "\*"
+   but you want to exclude a role.
+
+### Email
+
+Underlying Type: String
+
+An email address.
```
eriq-augustine, line 92: Put in a sentence here about how an email address is usually used to identify a user on the server.

eriq-augustine (changes requested):
Code's looking pretty good.
Most comments on documentation.

PR #184, opened 2025-05-27
Added required fields to API input fields in API descriptions.

Lucas-Ellenberger:
This PR adds required fields to API input fields in the automatically generated API description.

By adding the field `required` to the appropriate input fields, interfaces that call the `autograder-server` can pre-validate API requests.

This information is parsed from the source code by looking for a `required` field tag. The value of the tag is ignored, so it should be left empty. The pre-validation of API endpoints will be strengthened by the addition of the default value field, which will come in a future PR.

### DIFF
--- a/internal/api/core/describe.go
+++ b/internal/api/core/describe.go
@@ -32,16 +32,22 @@ type APIDescription struct {
 }
 
 type EndpointDescription struct {
-	Description  string             `json:"description"`
-	RequestType  string             `json:"-"`
-	ResponseType string             `json:"-"`
-	Input        []FieldDescription `json:"input"`
-	Output       []FieldDescription `json:"output"`
+	Description  string                 `json:"description"`
+	RequestType  string                 `json:"-"`
+	ResponseType string                 `json:"-"`
+	Input        []FieldDescription     `json:"input"`
+	Output       []BaseFieldDescription `json:"output"`
+}
+
+type BaseFieldDescription struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
 }
 
 type FieldDescription struct {
-	Name string `json:"name"`
-	Type string `json:"type"`
+	BaseFieldDescription
+
+	Required bool `json:"required"`
 }
 
 type TypeDescription struct {
@@ -157,12 +163,18 @@ func DescribeRoutes(routes []Route) (*APIDescription, error) {
 			errs = errors.Join(errs, err)
 		}
 
+		outputBaseFields := make([]BaseFieldDescription, 0, len(output.Fields))
+
+		for _, field := range output.Fields {
+			outputBaseFields = append(outputBaseFields, field.BaseFieldDescription)
+		}
+
 		endpointMap[apiRoute.GetBasePath()] = EndpointDescription{
 			Description:  apiRoute.Description,
 			RequestType:  apiRoute.RequestType.String(),
 			ResponseType: apiRoute.ResponseType.String(),
 			Input:        input.Fields,
-			Output:       output.Fields,
+			Output:       outputBaseFields,
 		}
 	}
 
@@ -350,6 +362,8 @@ func describeStructFields(customType reflect.Type, info TypeInfoCache) ([]FieldD
 			continue
 		}
 
+		requiredField := isFieldRequired(field)
+
 		// Handle embedded fields.
 		if field.Anonymous {
 			fieldDescription, fieldTypeID, _, err := DescribeType(field.Type, false, info)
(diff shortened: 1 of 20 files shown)


internal/api/core/describe.go
```diff
@@ -449,6 +475,15 @@ func describeStruct(customType reflect.Type, knownPackages map[string]StructDesc
 	return description, nil
 }
 
+func isFieldRequired(field reflect.StructField) bool {
+	_, ok := field.Tag.Lookup("required")
```
eriq-augustine, line 479: We want to discourage tags like `required:"false"`,
so you should probably ensure that the tag value is empty.

internal/email/message.go
```diff
@@ -10,11 +10,11 @@ import (
 )
 
 type Message struct {
-	To      []string `json:"to"`
+	To      []string `json:"to" required:""`
```
eriq-augustine, line 13: You can send an email without a to, e.g., only bccs.

internal/email/message.go
```diff
@@ -10,11 +10,11 @@ import (
...
 	CC      []string `json:"cc"`
 	BCC     []string `json:"bcc"`
-	Subject string   `json:"subject"`
-	Body    string   `json:"body"`
+	Subject string   `json:"subject" required:""`
+	Body    string   `json:"body" required:""`
```
eriq-augustine, line 17: You can send messages without a body.
Especially before most common chat platforms like Slack,
people would sent emails that were only subjects and end with "EOM" ("End of Message").

internal/api/core/describe.go
```diff
@@ -32,18 +32,24 @@ type APIDescription struct {
...
-type FieldDescription struct {
+type BaseFieldDescription struct {
 	Name string `json:"name"`
 	Type string `json:"type"`
 }
 
+type FieldDescription struct {
+	BaseFieldDescription
+
+	Required bool `json:"required"`
```
eriq-augustine, line 50: I'm thinking you should use `omitzero` with this.
For required is not a big deal,
but we probably don't want to make a pattern of having all the extra fields for each type.

resources/api.json
```diff
@@ -1527,23 +1764,28 @@
             "fields": [
                 {
                     "name": "courses",
-                    "type": "map[string]core.EnrollmentInfo"
+                    "type": "map[string]core.EnrollmentInfo",
+                    "required": false
```
eriq-augustine, line 1768: Having the "required" key here is a bit awkward for output types.
It's mostly false (because that's the default), but what does it mean when it is true
(which could happen since a common type may be used as input and output)?
A user may think that required being false may mean the field may be optional in the output (when that probably isn't true).

You may want to just have the required field on the endpoint fields and leave it off of "types" entries.

eriq-augustine (changes requested):
The concept is looking pretty good.

I think there are some more semantics to think about for what entries should show the required field.
For example, there are input types, output types, and internal types (entries under "types").
I don't think the "required" semantics hold for all of these.

docs/development.md
```diff
@@ -102,6 +102,16 @@ The structure of API endpoint files must follow these rules:
...
 
+### Automatic API Descriptions
+
+The server generates API descriptions automatically from source code.
+
+ - Each API endpoint must have a comment describing the endpoint attached to the handler's function declaration.
+ - User-facing input fields must include a `json` tag to appear in the description.
+ - Required fields must also include a `required` tag with an empty value.
+
+For example, the field tag for the required `UserEmail` input field is `json:"user-email" required:""`.
```
eriq-augustine, line 113: This can be hard for first timers to understand.
Why don't you put in a very simple example endpoint (request, response, and handler),
just in this document.

internal/api/core/describe.go
```diff
@@ -157,18 +193,26 @@ func DescribeRoutes(routes []Route) (*APIDescription, error) {
 			errs = errors.Join(errs, err)
 		}
 
+		outputBaseFields := make([]BaseFieldDescription, 0, len(output.Fields))
+
```
eriq-augustine, line 197: I would remove this newline.
Then creating and filling the slice can be in the same visual block.

internal/api/core/describe.go
```diff
@@ -449,6 +507,19 @@ func describeStruct(customType reflect.Type, knownPackages map[string]StructDesc
...
+	tag, ok := field.Tag.Lookup("required")
+	if !ok {
+		return false
+	}
+
+	if tag == "" {
+		return true
+	}
+
+	return false
```
eriq-augustine, line 520: This silent failure seems a bit dangerous.
Unless I am missing something else,
a user can put in a bad tag (e.g.: `required:"true"`) and not have the field be required (unbeknownst to them).

eriq-augustine (dismissed):
Just some small things.

eriq-augustine (approved):
lgtm

PR #188, opened 2025-06-13
Added server user references to provide flexible targeting of server users.

Lucas-Ellenberger:
This PR adds server user references which provide a flexible way to target server users and user groups.

The input strings can be as follows:
* Server user references can represent the following:
* An email address.
* A literal "*" (which includes all users on the server).
* A server role (which will include all server users with that role).
* <course-id>::<course-role> (which will include all users in the target course with that role).
* *::<course-role> (which will include all users with the course role in any course).
* <course-id>::* (which will include all users in the target course).
* *::* (which includes all users enrolled in at least one course).
* Any of the above options preceded by a dash ("-") (which indicates that the user or group will NOT be included in the final results).

Input order does not affect the outcome as exclusions always supersede inclusions (e.g. []{"-admin", "*"} targets all users in the server that are not server admins).

The efficacy of this powerful course user targetting system is displayed by updating `users/list`.


internal/model/user_reference.go
```diff
@@ -129,6 +396,89 @@ func ParseCourseUserReferences(rawReferences []CourseUserReference) (*ParsedCour
...
+
+		_, ok = excludeSet[email]
+		if ok {
+			continue
+		}
+
+		emailSet[email] = nil
+	}
+
+	for _, courseReference := range reference.CourseUserReferences {
```
eriq-augustine, line 455: Put a simple comment on this loop to maintain symmetry with your other blocks.
Make sure that mention that the Excludes()/ReferseTo() above already dealt with role-based course references.

internal/api/users/list_test.go
```diff
@@ -15,32 +15,51 @@ func TestList(test *testing.T) {
 
 	users := make([]*model.ServerUser, 0, len(usersMap))
 	for _, user := range usersMap {
+		if user.GetName(true) == "root" {
```
eriq-augustine, line 18: Use `user.Email == model.RootUserEmail`.

Using name is dangerous, since user's can set their own name.
I know this is just testing code, but imagine an attack where the user just needs to set their name to "root" to get privileges.
I think I remember an OSX attack like that a few years back.

eriq-augustine (changes requested):
Just some small and sneaky things.

internal/model/course_user_reference.go
```diff
@@ -28,6 +28,65 @@ type ParsedCourseUserReference struct {
...
+		mergedReference.CourseUserRoles[role] = nil
+	}
+
+	for role, _ := range this.ExcludeCourseUserRoles {
+		mergedReference.ExcludeCourseUserRoles[role] = nil
+	}
+
+	for role, _ := range other.ExcludeCourseUserRoles {
+		mergedReference.ExcludeCourseUserRoles[role] = nil
+	}
```
eriq-augustine, line 85: Instead of the pairs for this/other, you can do:
```
range _, reference := range []*ParsedCourseUserReference{this, other} {
    ...
}
```

internal/model/server_user_reference.go
```diff
@@ -0,0 +1,319 @@
...
+		return
+	}
+
+	currentCourseUserReference, ok := this.CourseUserReferences[courseID]
+	if !ok {
+		this.CourseUserReferences[courseID] = courseUserReference
+		return
+	}
+
+	this.CourseUserReferences[courseID] = currentCourseUserReference.Merge(courseUserReference)
```
eriq-augustine, line 53: If you handle a nil `this` in Merge(), then you can do all this in one line:
```
this.CourseUserReferences[courseID] = this.CourseUserReferences[courseID].Merge(courseUserReference)
```

eriq-augustine (changes requested):
Just a couple more things.

eriq-augustine (approved):
lgtm!

PR #189, opened 2025-06-16
Phased filter role out in favor of user references for more expressive user targeting.

Lucas-Ellenberger:
This PR phases out the use of filter role in favor of user references, which allows for more expressive user targeting.

These changes are implemented at the database level, which effects the `courses/assignments/submissions/fetch/course/attempts` and `courses/assignments/submissions/fetch/course/scores` API endpoints. The endpoints remain semantically equivalent, but offer finer grain filtering due to the power of course user references.

These changes are the final stepping stone before the `courses/assignments/submissions/proxy/regrade` API endpoint can be implemented (which would replace [the outdated PR](https://github.com/edulinq/autograder-server/pull/170)).

### DIFF
--- a/internal/model/user_reference_course.go
+++ b/internal/model/user_reference_course.go
@@ -45,7 +45,7 @@ func (this *ParsedCourseUserReference) ToParsedServerUserReference(courseID stri
 }
 
 func (this *ParsedCourseUserReference) Merge(other *ParsedCourseUserReference) *ParsedCourseUserReference {
-	mergedReference := &ParsedCourseUserReference{
+	mergedReference := ParsedCourseUserReference{
 		Emails:                 make(map[string]any, 0),
 		ExcludeEmails:          make(map[string]any, 0),
 		CourseUserRoles:        make(map[CourseUserRole]any, 0),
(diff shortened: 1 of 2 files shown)


eriq-augustine (changes requested):
Nice.
The RefersTo() usage seems much more natural at the DB level than checking the role).

Something we missed last PR is that the names of the user reference files are off.
You can see with symmetric types like roles, we have `role_course.go` and `role_server.go`.
We should probably use the same pattern here (this way they show up next to each other when sorted).

internal/model/user_reference_course.go
eriq-augustine, line 1: Since the DB was updated to use pointers to user references,
you should probably update the methods to use pointer receivers.

You usually want to be consistent with all pointer or all non-pointer receivers.
(Using all pointer receivers tends to be more common, since all non-pointer receivers indicates a small immutable type (which we just don't use as much in Go).)

Also (but not as important), you are making (probably unintended) copies,
since Go has to dereference the pointer to make the call.

Since you can, also consider adding in handling for nil cases so a nil user reference refer to no one.

eriq-augustine (changes requested):
Just one low-level thing.

Issue #141, opened 2025-01-24
Incorporate Assignment Options in Code Analysis


eriq-augustine:
Right now, there is no way to give input/options to the code analysis system.
It would be great if assignments could provide some options to the process.

For example:
 - Template Files*
 - Files to Ignore
 - General Similarity Options (e.g., a general "sensitivity" option that engines can interpret how they want)
 - Engine-Specific Similarity Options

*Template files is the most important option listed here.
The idea is that all code in a "template" could be ignored, so we can just focus on a student's code.
So we can ignore all boiler plate and provided code.

A tricky part here is what to do when submissions from two different assignments (with different options) are compared.
This would not be an unusual situation if there are multiple sections of the same course.
We could just arbitrarily choose the first assignment in the key and use those options.

eriq-augustine:
For include/exclude, we can allow globs and use two different lists.

Default values:
Include: `["*"]`
Exclude: `[]`

eriq-augustine:
Include/Exclude support was added in 8101b39ecd874d89846e3e5f9dd9656b6133e010.

For multiple-assignment option resolution, we just use the options from the first submission.

Documentation: https://github.com/edulinq/autograder-server/blob/main/docs/types.md#analysis-options-analysisoptions

PR #192, opened 2025-06-24
Incorporate Assignment Options in Code Analysis (Resolves Issue #141)

anvichip:
PR for issue: https://github.com/edulinq/autograder-server/issues/141

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -10,6 +10,7 @@ Extra keys will generally be ignored.
  - [Semantic Types](#semantic-types)
    - [Identifier](#identifier)
    - [Course User Reference (CourseUserReference)](#course-user-reference-courseuserreference)
+   - [Server User Reference (ServerUserReference)](#server-user-reference-serveruserreference)
    - [Email](#email)
    - [Timestamp](#timestamp)
    - [Pointer](#pointer)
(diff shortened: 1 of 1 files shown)


internal/util/engine_options.go
```diff
@@ -0,0 +1,45 @@
...
+		log.Warn("Engine-specific options are not map[string]any.", log.NewAttr("engine", engineName))
+		return result, false
+	}
+
+	for _, key := range keys { //loop through the keys
+		val, exists := engineOptions[key]
+		if !exists {
+			continue
+		}
+		if fVal, ok := val.(float64); ok { //extract the float64 value
```
Lucas-Ellenberger, line 37: We are severely limiting the types of options that can appear. If all the values in the map must be a float, it should be a `map[string]float64`.

Since you probably made the map to accept the type `any` for the value, you should be able to support `any` (not just `float64`).

testdata/course101/HW0/assignment.json
```diff
@@ -6,6 +6,14 @@
     ],
     "image": "ghcr.io/edulinq/grader.python:0.1.0.2-alpine",
     "analysis-options": {
+        "engine-options": {
```
Lucas-Ellenberger, line 9: Since you are modifying testdata, the md5 hash of the testdata folder will change. You will need to update the corresponding test.

internal/analysis/jplag/engine.go
```diff
@@ -50,7 +50,20 @@ func (this *JPlagEngine) IsAvailable() bool {
 	return docker.CanAccessDocker()
 }
 
-func (this *JPlagEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context) (*model.FileSimilarity, error) {
+func (this *JPlagEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context, options any) (*model.FileSimilarity, error) {
+
+	effectiveMinTokens := this.MinTokens // Start with the engine's default/configured minTokens
+
+	optMap, ok := util.ExtractEngineOptionMap(options, "jplag", []string{"minTokens"})
```
Lucas-Ellenberger, line 57: I think we need to rethink how we get the option values out of the map.

I believe there needs to be a function for both `dolos` and `jplag`. Each of these functions will know more about the keys and values they expect out of the map. This will help you create more safe code in terms of type safety and more extendible code as we potentially add new options to the engines in the future.

Note, these engine specific helper functions can use a common util function (if that is something that makes sense).

PR #194, opened 2025-07-05
Incorporate Assignment Options in Code Analysis (Resolves Issue #141)

anvichip:
This PR adds support for passing engine options (e.g., minTokens in Jplag) to Engines, allowing them to be configured per assignment.


eriq-augustine:
Note that this is a new version of #192.

internal/analysis/pairwise.go
```diff
@@ -248,22 +256,41 @@ func computeFileSims(options AnalysisOptions, inputDirs [2]string, assignment *m
...
 
 		for i, engine := range engines {
+
+			// Initialize specific options for the engine.
+			var specificEngineOptions map[string]any
+
+			// Get specific options for this engine.
+			specificEngineOptionsAny, found := engineOptions[engine.GetName()]
+
+			if found && specificEngineOptionsAny != nil { // Check if found and specificEngineOptions are not nil.
```
Lucas-Ellenberger, line 266: Please read my comments. I can almost guarantee that Eriq would have stopped the review by now given the fact that there has been 3-4 (and counting) previous review comments that were marked as resolved but remain unaddressed.

In this case, it is the usage of inline comments: https://github.com/edulinq/autograder-server/pull/192#discussion_r2165147329

This is frustrating as a reviewer because I have to repeat the same work over again.

internal/analysis/jplag/engine_test.go
```diff
@@ -15,11 +15,13 @@ func TestJPlagComputeFileSimilarityBase(test *testing.T) {
...
 	}
 
-	// Lower the token minimum for testing.
 	engine := GetEngine()
-	engine.MinTokens = 5
 
-	core.RunEngineTestComputeFileSimilarityBase(test, engine, false, expected)
+	engineOptions := map[string]any{
+		"minTokens": 5,
+	}
```
Lucas-Ellenberger, line 22: Something to think about, would it be better to declare this as an `JPlagEngineOptions` and convert it to a `map[string]any` using JSON?

This could help us avoid issues with typing an incorrect JSON field name. Remember that you had an issue with mistyping the JSON name that gave you trouble debugging. We want to be nice as possible to future developers (within reason).

internal/analysis/pairwise.go
```diff
@@ -248,22 +256,41 @@ func computeFileSims(options AnalysisOptions, inputDirs [2]string, assignment *m
...
+				var ok bool
+				specificEngineOptions, ok = specificEngineOptionsAny.(map[string]any)
+				if !ok {
+					// Handle the case where the assertion fails.
+					specificEngineOptions = nil
+					log.Warn("Engine options for '%s' could not be converted into map[string]any. Engines will use default values.", engine.GetName())
+					continue // Skip to the next engine.
+				}
```
Lucas-Ellenberger, line 274: This entire section is much too complicated for this block of code. Computing the file similarity for every engine in parallel is complicated enough for this loop. We need to reduce the readers cognitive load and abstract out the enging option exctraction.

Perhaps the `SimilarityEngine` option could require all engines to implement a method that extracts engine options from `EngineOptions`. This is a quick suggestion, so take some time to think about a better way to extract options for each engine type. Using the engine name is a good idea, but there are some things to iron out or abstract.

internal/analysis/pairwise.go
```diff
@@ -248,22 +256,41 @@ func computeFileSims(options AnalysisOptions, inputDirs [2]string, assignment *m
...
+			}
+
 			// Compute the file similarity for each engine in parallel.
 			// Note that because we know the index for each engine up-front, we don't need a channel.
 			engineWaitGroup.Add(1)
-			go func(index int, simEngine core.SimilarityEngine) {
+			// Pass options to function.
+			go func(index int, simEngine core.SimilarityEngine, EngineOptions map[string]any) {
```
Lucas-Ellenberger, line 281: Why is the `EngineOptions` parameter exported while the `index` and `simEngine` parameters are not?

internal/analysis/pairwise_test.go
```diff
@@ -802,3 +796,56 @@ func TestPairwiseAnalysisFailureBase(test *testing.T) {
 			expectedMessageSubstring, results[key].FailureMessage)
 	}
 }
+
+func TestExtractJplagOptions(test *testing.T) {
+	testCases := []struct {
+		name     string
```
Lucas-Ellenberger, line 802: This field in the anonymous struct is not used in the loop over `testCases`. So, it seems to be only useful in explaining the purpose of the test. Since it is purely explanatory (not useful in code), it should be a comment. Take a look at other testing functions in this project to get an idea of the comments included for test cases.

Remember, you are responsible for all code submitted in a PR. Unused fields should not make it into a PR. As Eriq recommended in our last meeting, take the time to read through every line of your diff to catch simple mistakes like this one.

internal/analysis/pairwise_test.go
```diff
@@ -802,3 +796,56 @@ func TestPairwiseAnalysisFailureBase(test *testing.T) {
...
+		{
+			name:     "other options present, minTokens valid",
+			input:    map[string]any{"minTokens": 200, "anotherOption": "value"},
+			expected: jplag.JPlagEngineOptions{MinTokens: 200},
+		},
+	}
+
+	var extractedOptions jplag.JPlagEngineOptions
+	for i, testCase := range testCases {
+		extractedOptions = jplag.ExtractJplagOptions(testCase.input)
```
Lucas-Ellenberger, line 845: Do we need to declare the tyoe of `extractedOptions` ahead of time? Can we not let the type be inferred on line 845 (using the `:=` operator)?

internal/model/analysis.go
```diff
@@ -27,6 +27,8 @@ type AssignmentAnalysisOptions struct {
 
 	TemplateFiles   []*util.FileSpec      `json:"template-files,omitempty,omitzero"`
 	TemplateFileOps []*util.FileOperation `json:"template-file-ops,omitempty,omitzero"`
+
+	EngineOptions map[string]any `json:"engine-options,omitempty,omitzero"`
```
Lucas-Ellenberger, line 31: Include a comment on the expected structure of these engine options. What are acceptable map keys and values?

testdata/course101/HW0/assignment.json
```diff
@@ -6,6 +6,11 @@
     ],
     "image": "ghcr.io/edulinq/grader.python:0.1.0.2-alpine",
     "analysis-options": {
+        "engine-options": {
+            "jplag": {
+                "minTokens": 0.9
```
Lucas-Ellenberger, line 11: You updated `minTokens` to be an int for `jplag`'s engine options. Why would our own testdata include a float?

Even if the code properly converts this to the desired int, other repositories and course developers look at our testdata to better understand the expected structure of the data. So, we should set a good example and use the proper type.

internal/analysis/jplag/engine.go
```diff
@@ -32,14 +32,14 @@ var (
 	imageLock sync.Mutex
 )
 
-type JPlagEngine struct {
+type JPlagEngineOptions struct {
 	MinTokens int
```
Lucas-Ellenberger, line 36: Other comments include suggestions to use JSON to convert raw options from the course into `JPlagEngineOptions`. So, include a json field tag. Be sure to follow the JSON field naming convention of the rest of the project.

internal/analysis/pairwise_test.go
```diff
@@ -802,3 +796,56 @@ func TestPairwiseAnalysisFailureBase(test *testing.T) {
...
+
+		if !reflect.DeepEqual(extractedOptions.MinTokens, testCase.expected.MinTokens) {
+			test.Errorf("Case: %d extractJplagOptions() error, got = %v want %v", i, extractedOptions.MinTokens, testCase.expected.MinTokens)
```
Lucas-Ellenberger, line 848: Include a `continue` after an `Errorf()`. I know there isn't any code below this conditional, but it is nice to include in case there additional checks for each test case are included (e.g., checking for the correct error and then checking for the correct value).

Lucas-Ellenberger (changes requested):
I'm glad we got to the second iteration, but this round had some disappointments.

We are getting closer to extracting the options in a better way, but I think we can do better.

Most importantly, there were issues marked as resolved from last time, but were not resolved.

internal/analysis/pairwise.go
anvichip: Added a func in util which can carry out this process.

internal/analysis/pairwise.go
```diff
@@ -248,22 +256,41 @@ func computeFileSims(options AnalysisOptions, inputDirs [2]string, assignment *m
 		var engineWaitGroup sync.WaitGroup
 
 		for i, engine := range engines {
+
```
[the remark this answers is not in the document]
Lucas-Ellenberger: This comment was not addressed.

internal/analysis/test.go
```diff
@@ -21,7 +21,7 @@ func (this *fakeSimiliartyEngine) IsAvailable() bool {
 	return true
 }
 
-func (this *fakeSimiliartyEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context) (*model.FileSimilarity, error) {
+func (this *fakeSimiliartyEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context, options map[string]any) (*model.FileSimilarity, error) {
```
Lucas-Ellenberger, line 24: Options is unused, follow the procedure we did with Dolos and use the `_` to indicate the parameter is unused.

Lucas-Ellenberger (changes requested):
Getting closer.

You really need to be careful and apply feedback throughout your PR. There were comments marked as resolved which were not addressed. Concerningly, you still used the standard library `log` package even after our lengthy discussions last week.

internal/analysis/pairwise_test.go
```diff
@@ -321,10 +318,18 @@ func TestPairwiseAnalysisDefaultEnginesSpecificFiles(test *testing.T) {
...
+	}
+	jsonBytes, err := json.Marshal(engineOptsStruct)
+	if err != nil {
+		log.Fatalf("Error marshaling JPlagEngineOptions to JSON: %v", err)
+	}
+	var engineOptions map[string]any
+	err = json.Unmarshal(jsonBytes, &engineOptions)
+	if err != nil {
+		log.Fatalf("Error unmarshaling JSON to map[string]any: %v", err)
+	}
```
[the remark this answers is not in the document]
anvichip: Currently, we do not have any method in place that is being used to converts JPlagEngineOptions into map[string]any. 
Cannot use previously applied logic here.

internal/analysis/pairwise.go
```diff
@@ -245,13 +248,20 @@ func computeFileSims(options AnalysisOptions, inputDirs [2]string, assignment *m
 		var engineWaitGroup sync.WaitGroup
 
 		for i, engine := range engines {
+			// Extract specific options for the engine.
+			specificEngineOptions, err := core.GetSpecificEngineOptions(engineOptions, engine.GetName())
+			if err != nil {
+				log.Warn("No engine options found for '%s' or engine options are empty. Engine will use default values.", engine.GetName())
```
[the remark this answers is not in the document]
anvichip: Handled the error returning outside the getEngines() function, I felt like making this integration inside of the function would have caused a lot of changes.

internal/analysis/jplag/engine.go
```diff
@@ -50,8 +56,33 @@ func (this *JPlagEngine) IsAvailable() bool {
 	return docker.CanAccessDocker()
 }
 
-func (this *JPlagEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context) (*model.FileSimilarity, error) {
-	err := ensureImage()
+// Sets engine options by marshalling the options map
+// to JSON and then unmarshalling it into the struct.
```
eriq-augustine, line 60: I don't think this comment is necessary here:
 - It seems to dive into unnecessary details.
 - Your function signature is pretty clear and self-documenting.

internal/analysis/jplag/engine.go
```diff
@@ -50,8 +56,33 @@ func (this *JPlagEngine) IsAvailable() bool {
...
+func GetJPlagEngineOptions(rawOptions map[string]any) (*JPlagEngineOptions, error) {
+	var effectiveOptions JPlagEngineOptions
+	if (len(rawOptions) == 0) || (rawOptions == nil) {
+		return nil, nil
+	}
```
eriq-augustine, line 65: Is this case already handled by util.JSONTransformTypes()?

internal/analysis/jplag/engine.go
```diff
@@ -50,8 +56,33 @@ func (this *JPlagEngine) IsAvailable() bool {
...
+
+	effectiveOptions, err := util.JSONTransformTypes(rawOptions, effectiveOptions)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to convert raw options to JPlagEngineOptions: '%v'.", err)
```
eriq-augustine, line 69: Careful.

s/%v/%w/

internal/analysis/jplag/engine.go
```diff
@@ -50,8 +56,33 @@ func (this *JPlagEngine) IsAvailable() bool {
...
+
+func (this *JPlagEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context, options map[string]any) (*model.FileSimilarity, error) {
+	effectiveOptions, err := GetJPlagEngineOptions(options)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to set custom JPlag engine options: '%w'.", err)
+	}
+
+	if effectiveOptions == nil {
+		effectiveOptions = GetDefaultJPlagOptionsCopy()
+	}
```
eriq-augustine, line 83: !
SNEAKY BUG

If I only set a subset (including none) options at the JSON level,
then they will be successfully parsed as far as JSON marshal/unmarshal is concerned.
But, the options I don't set are defaulted to empty values.

Include a test case for this bug.

Consider having your parse function return default options.

internal/analysis/pairwise_test.go
```diff
@@ -802,3 +801,71 @@ func TestPairwiseAnalysisFailureBase(test *testing.T) {
 			expectedMessageSubstring, results[key].FailureMessage)
 	}
 }
+
+func TestGetJplagOptions(test *testing.T) {
```
eriq-augustine, line 805: This test seems like it should be in the jplag package.

(And if this is true, then we don't have to export the parsing function.)

internal/model/analysis.go
```diff
@@ -21,12 +21,20 @@ const (
...
+type OptionsMap map[string]any
+
 type AssignmentAnalysisOptions struct {
 	IncludePatterns []string `json:"include-patterns,omitempty,omitzero"`
 	ExcludePatterns []string `json:"exclude-patterns,omitempty,omitzero"`
 
 	TemplateFiles   []*util.FileSpec      `json:"template-files,omitempty,omitzero"`
 	TemplateFileOps []*util.FileOperation `json:"template-file-ops,omitempty,omitzero"`
+
+	// EngineOptions includes the options for the analysis engines.
```
Lucas-Ellenberger, line 34: s/the analysis engines/analysis engines

This was a part of Eriq's comment in the previous review.

internal/analysis/dolos/engine.go
```diff
@@ -23,13 +23,31 @@ const (
...
+	DEFAULT_KGRAM_LENGTH    = 23
 )
 
 var (
 	hasImage  bool = false
 	imageLock sync.Mutex
 )
 
+type DolosEngineOptions struct {
+	// --kgrams-in-window
```
Lucas-Ellenberger, line 37: This isn't super helpful by itself. If we were working in the terminal, developers could see Dolos's help message. Since we aren't, I would provide more context about the option. I know Eriq commented to mention "--min-tokens", but I'm sure he would want the comment to provide more detail.

Note: this applies to the other comments on *EngineOption's fields.

internal/analysis/core/engine.go
```diff
@@ -16,5 +16,6 @@ type SimilarityEngine interface {
 	// Working on two files (submissions) at a time will typically be less efficient than working on all files at the same time,
 	// but a lot of shorter jobs is more flexible than one large job.
 	// On a timeout, (nil, nil) should be returned.
-	ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context) (*model.FileSimilarity, error)
+	// Options carries engine‑specific parameters, e.g., the minimum number of tokens required to perform similarity analysis by JPlag.
```
Lucas-Ellenberger, line 19: s/tokens/words

Eriq suggested this in a previous review and explained developers may not be familiar with JPlag "tokens".

internal/analysis/jplag/engine.go
```diff
@@ -50,8 +57,24 @@ func (this *JPlagEngine) IsAvailable() bool {
...
+		return nil, fmt.Errorf("Failed to convert raw options to JPlagEngineOptions: '%w'.", err)
+	}
+
+	return effectiveOptions, nil
+}
+
+func (this *JPlagEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context, options model.OptionsMap) (*model.FileSimilarity, error) {
+	effectiveOptions, err := parseEngineOptions(options)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to set custom JPlag engine options: '%w'.", err)
```
Lucas-Ellenberger, line 74: s/set/parse

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,90 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+			}
+		} else {
+			if testCase.extractionError {
+				test.Errorf("Case %d: Did not get an expected error.", i)
+				continue
+			}
+		}
+
+		if !reflect.DeepEqual(effectiveOptions, testCase.expected) {
+			test.Errorf("Case %d: Unexpected result. Expected = '%v', Actual = '%v'.", i, testCase.expected, effectiveOptions)
```
Lucas-Ellenberger, line 125: s/Expected = /Expected: /
s/Actual = /Actual: /

This is our standard way to report differences in expected data during tests. It's nice when we can keep our error messages consitent. Some of our older tests use slight variations, but new code follows this convention.

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,90 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+		if err != nil {
+			if !testCase.extractionError {
+				test.Errorf("Case %d: Got an unexpected error: '%v'.", i, err)
+				continue
+			}
+		} else {
+			if testCase.extractionError {
+				test.Errorf("Case %d: Did not get an expected error.", i)
+				continue
+			}
```
Lucas-Ellenberger, line 121: I don't think this needs to be in an `else` branch. I would put the continue outside of the `if !testCase.extractionError` branch (but still inside the first if block.

This way, if we have an error, we will always continue at the end of the block. But, only the test cases that were not expecting an error will call `test.Errorf()`.

Then, you can remove the else statement and bring this block of code one level of identation back.

Both versions of the code function the same, but I believe this version is more readable. We follow this pattern quite often in tests.
Ex:
https://github.com/edulinq/autograder-server/blob/d4839e39d4dfe4700ffd98e35196e7aa96b4b35b/internal/jobmanager/jobmanager_test.go#L226-L241

testdata/course101/HW0/assignment.json
```diff
@@ -6,6 +6,14 @@
     ],
     "image": "ghcr.io/edulinq/grader.python:0.1.0.2-alpine",
     "analysis-options": {
+        "engine-options": {
+            "jplag": {
+                "min-tokens": 5
+            },
+            "dolos": {
+                "kgram-length": 22
```
Lucas-Ellenberger, line 14: Should we include a `kgrams-in-window`? I'll leave this up to you (not a hard requirement).

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,90 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+		{
+			input:           model.OptionsMap{"min-tokens": 75.5},
+			expected:        nil,
+			extractionError: true,
+		},
+		{
+			input:           model.OptionsMap{"min-tokens": "abc"},
+			expected:        nil,
+			extractionError: true,
+		},
```
Lucas-Ellenberger, line 93: Normally we put the error test cases at the bottom.

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,90 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+		extractionError bool
+	}{
+		// Base Case
+		{
+			input:           model.OptionsMap{"min-tokens": 100},
+			expected:        &JPlagEngineOptions{MinTokens: 100},
+			extractionError: false,
+		},
+
+		// Empty options
```
Lucas-Ellenberger, line 71: What about a `nil` input?

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,90 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+		{
+			input:           map[string]any{},
+			expected:        GetDefaultJPlagOptions(),
+			extractionError: false,
+		},
+		{
+			input:           model.OptionsMap{},
+			expected:        GetDefaultJPlagOptions(),
+			extractionError: false,
+		},
```
Lucas-Ellenberger, line 81: I would put these as the first test cases. Normally we go from the simplest test cases to the more complex ones.

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,90 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+	core.RunEngineTestComputeFileSimilarityBase(test, engine, true, expected, engineOptions)
+}
+
+func TestParseJplagOptions(test *testing.T) {
+	testCases := []struct {
+		input           map[string]any
+		expected        *JPlagEngineOptions
+		extractionError bool
+	}{
+		// Base Case
```
Lucas-Ellenberger, line 64: Why is this the "base case"? I would imagine most people will not pass any options and accept the default.

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,90 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+		// Fallback to Default
+		{
+			input:           model.OptionsMap{"min-tokens": nil},
+			expected:        GetDefaultJPlagOptions(),
+			extractionError: false,
+		},
+
+		// Extra Options
+		{
+			input:           model.OptionsMap{"min-tokens": 200, "another-option": "value"},
```
Lucas-Ellenberger, line 104: When specifying multiple key-value pairs, this project prefers to place them on their own line. To keep things consistent in this file, you should make all test cases put the KV pairs on their own line (even if there is only one).

Ex:
```
model.CircusStaff{
    "clowns": 100,
    "ring-leader": 1,
}
```

docs/types.md
```diff
@@ -379,6 +380,37 @@ When working with these patterns, keep the following in mind:
...
+
+Dolos is a plagiarism detection tool optimized for performance and detailed similarity reporting. 
+It works by parsing code into tokens, generating fingerprints, comparing these fingerprints, and then visualizing the results to help educators identify potential plagiarism
+
+Current Supported Options:
+
+| Name               | Type           | Required | Description |
+|--------------------|----------------|----------|-------------|
+`kgram-length` | number | false | Sets the minimum number of tokens in a k-gram. Common fragments between two files that are shorter than `$$k$$` tokens will not be found during plagiarism detection.
+`window-length`	| number | false | The size of the window used during winnowing algorithm. It select one k-grams from each overlapping window of `w` subsequent k-grams.
```
Lucas-Ellenberger, line 412: !?

This isn't the name of the option. You called it `kgrams-in-window` (which makes sense to match the name of the Dolos option.).

It would be really bad if we included this in our documentation. Users following our documentation would pass `window-length` and be VERY confused and frustrated when the options were not being used (because it's the incorrect JSON key).

internal/analysis/dolos/engine_test.go
```diff
@@ -20,7 +21,10 @@ func TestDolosComputeFileSimilarityBase(test *testing.T) {
 		Score:    0.717949,
 	}
 
-	core.RunEngineTestComputeFileSimilarityBase(test, GetEngine(), false, expected)
+	// Empty engine option map for testing.
+	engineOpts := make(model.OptionsMap)
```
Lucas-Ellenberger, line 25: I am not a fan of `engineOpts`.

Don't optimize for variable name length. It's much better for the name to have three more letters and be called `engineOptions`.

Also, you have been calling it `engineOptions` for a while in this PR now, so why change?

internal/analysis/pairwise.go
```diff
@@ -186,6 +186,17 @@ func computeFileSims(options AnalysisOptions, inputDirs [2]string, assignment *m
...
 
+	allEngineOptions := make(map[string]model.OptionsMap)
+	if (assignment != nil) && (assignment.AssignmentAnalysisOptions != nil) {
+		allEngineOptions = assignment.AssignmentAnalysisOptions.EngineOptions
+	}
+
+	allEngineOptionsFinal := make(map[string]model.OptionsMap)
+	for _, engine := range engines {
+		specificEngineOptions := GetEngineOptions(allEngineOptions, engine.GetName())
+		allEngineOptionsFinal[engine.GetName()] = specificEngineOptions
```
Lucas-Ellenberger, line 197: I'm not a fan of the name `allEngineOptionsFinal`.

Maybe we can do the following:
* s/allEngineOptions/engineOptions (or rawEngineOptions?)
* s/allEngineOptionsFinal/parsedEngineOptions
* Do we need `specificEngineOptions` at all? Can we combine lines 196 and 197 since `GetEngineOptions()` doesn't return an error anymore?

Lucas-Ellenberger (changes requested):
The core is looking better.

Some comments on documentation and code maintainability.

Lucas-Ellenberger (changes requested):
So close!

Just some more notes about documentation and a couple testing changes.

internal/analysis/jplag/engine.go
```diff
@@ -32,16 +32,23 @@ var (
 	imageLock sync.Mutex
 )
 
-type JPlagEngine struct {
-	MinTokens int
+type JPlagEngineOptions struct {
+	// Minimum number of consecutive tokens that must match between two code submissions to be considered similar. (--min-tokens).
```
Lucas-Ellenberger, line 36: s/similar. (/similar (

Extra period

internal/analysis/test.go
```diff
@@ -21,7 +21,7 @@ func (this *fakeSimiliartyEngine) IsAvailable() bool {
 	return true
 }
 
-func (this *fakeSimiliartyEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context) (*model.FileSimilarity, error) {
+func (this *fakeSimiliartyEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context, Options model.OptionsMap) (*model.FileSimilarity, error) {
```
Lucas-Ellenberger, line 24: s/Options/options

I commented on this last time. Carefully read the substitution so we match the interface's pattern. This looks extra strange because all other variable start with a lowercase letter.

Lucas-Ellenberger (changes requested):
Getting real close.

A few documentation things and one more code thing.

internal/analysis/dolos/engine.go
```diff
@@ -44,8 +62,28 @@ func (this *dolosEngine) IsAvailable() bool {
 	return docker.CanAccessDocker()
 }
 
-func (this *dolosEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context) (*model.FileSimilarity, error) {
-	err := ensureImage()
+func parseEngineOptions(rawOptions model.OptionsMap) (*DolosEngineOptions, error) {
```
eriq-augustine, line 65: I think this function can be generalized to include the functionality for JPlag as well.

A possible signature could be:
```
func parseEngineOptions[T any](rawOptions model.OptionsMap, defaultOptions T) (T, error) {
```

It would probably be able to be called like:
```
options, err := parseEngineOptions(rawOptions, GetDefaultDolosOptions())
```

internal/analysis/dolos/engine.go
```diff
@@ -44,8 +62,28 @@ func (this *dolosEngine) IsAvailable() bool {
...
+	effectiveOptions, err := util.JSONTransformTypes(rawOptions, effectiveOptions)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to convert raw options to DolosEngineOptions: '%w'.", err)
+	}
+
+	return effectiveOptions, nil
+}
+
+func (this *dolosEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context, options model.OptionsMap) (*model.FileSimilarity, error) {
+	effectiveOptions, err := parseEngineOptions(options)
```
eriq-augustine, line 81: Let's rename "options" so it's more clear that people shouldn't be using it pass the first line of this function,
and instead they should be using "effectiveOptions".

Maybe:
s/options/rawOptions/
s/effectiveOptions/options/

internal/analysis/dolos/engine_test.go
```diff
@@ -35,5 +36,113 @@ func TestDolosComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+	core.RunEngineTestComputeFileSimilarityBase(test, GetEngine(), true, expected, nil)
+}
+
+func TestParseDolosOptions(test *testing.T) {
+	testCases := []struct {
+		input           map[string]any
+		expected        *DolosEngineOptions
+		extractionError bool
+	}{
+		// Empty options.
```
eriq-augustine, line 48: s/options/Options/

internal/analysis/dolos/engine_test.go
```diff
@@ -35,5 +36,113 @@ func TestDolosComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
+			expected:        GetDefaultDolosOptions(),
+			extractionError: false,
+		},
+		{
+			input:           model.OptionsMap{},
+			expected:        GetDefaultDolosOptions(),
+			extractionError: false,
+		},
+
+		// Custom Options.
```
eriq-augustine, line 65: I would remove your periods from your "Title Case" comments.

internal/analysis/jplag/engine_test.go
```diff
@@ -30,9 +40,111 @@ func TestJPlagComputeFileSimilarityWithIgnoreBase(test *testing.T) {
...
 
-	core.RunEngineTestComputeFileSimilarityBase(test, engine, true, expected)
+	// Lower the token minimum for testing.
+	engineOptionsStruct := JPlagEngineOptions{
+		MinTokens: 5,
+	}
+
+	engineOptions, err := util.ToJSONMap(engineOptionsStruct)
+	if err != nil {
+		test.Errorf("Failed to convert JPlagEngineOption to map[string]any: '%v'.", err)
```
eriq-augustine, line 52: This error is too specific.
Keep it more simple with something like:
"Failed to convert engine options: '%v'."

eriq-augustine, line 52: Since we can't continue the test here, this should probably be an test.Fatalf() instead.

(Found by Lucas.)

internal/analysis/pairwise.go
```diff
@@ -186,6 +186,16 @@ func computeFileSims(options AnalysisOptions, inputDirs [2]string, assignment *m
 		return nil, nil, nil, err
 	}
 
+	rawEngineOptions := make(map[string]model.OptionsMap)
```
eriq-augustine, line 189: Add the initial capacity on make.

internal/model/analysis.go
```diff
@@ -21,12 +21,20 @@ const (
 	DEFAULT_INCLUDE_REGEX string = ".+"
 )
 
+// Holds options specific to a particular engine.
```
eriq-augustine, line 24: s/engine/analysis engine/

eriq-augustine (changes requested):
Looking pretty good.
Very close.

Lucas-Ellenberger (changes requested):
So close.

Stay vigilant even though we are super close. I feel like there is some weirdness around the generic, but let me know if the ideal version isn't possible.

internal/analysis/test.go
```diff
@@ -21,7 +21,7 @@ func (this *fakeSimiliartyEngine) IsAvailable() bool {
 	return true
 }
 
-func (this *fakeSimiliartyEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context) (*model.FileSimilarity, error) {
+func (this *fakeSimiliartyEngine) ComputeFileSimilarity(paths [2]string, templatePath string, ctx context.Context, options model.OptionsMap) (*model.FileSimilarity, error) {
```
Lucas-Ellenberger, line 24: Unresolved from last time.

s/options/rawOptions/

I even added a sentence to remind you to update the fake similarity engine.

Lucas-Ellenberger (approved):
lgtm!

PR #195, opened 2025-07-07
Added an API endpoint to make proxy regrades.

Lucas-Ellenberger:
This PR adds a new API endpoint `courses/assignment/submissions/proxy/regrade` that allows course graders (or above) to make mass proxy regrades based on a course user references.

Regrades take the most recent submission for every user with the raw references and resubmit with a proxy time that is the same time as the original submission. This features allows course staff to easily regrade every students (or another role's) submissions.

Users are given the option to specify a `regrade-after` time, which means all users must have a new submission made after the target time. If a time is not provided, the current time is used. This semantic allows for better regrade caching.

As regrades may take a long time (as there could be many students), the endpoint gives the option to wait for completion. If the requesting user does not wait for completion, the results will be available by looking at the submission history or calling regrade with the same `regrade-after` time.

Note that this PR is a new version #170. The old PR will remain open as a reference until this PR is merged.

### DIFF
--- a/internal/grader/regrade.go
+++ b/internal/grader/regrade.go
@@ -17,7 +17,7 @@ type RegradeOptions struct {
 	jobmanager.JobOptions
 	GradeOptions `json:"-"`
 
-	// The raw references of users to regrade.
+	// The raw course user references to regrade.
 	RawReferences []model.CourseUserReference `json:"target-users" required:""`
 
 	// Ensure every user has made a new submission after this time.
--- a/internal/grader/regrade.go
+++ b/internal/grader/regrade.go
@@ -57,13 +57,12 @@ func Regrade(assignment *model.Assignment, options RegradeOptions) (*RegradeResu
 		options.Context = context.Background()
 	}
 
-	var regradeAfter timestamp.Timestamp = 0
 	if options.RegradeAfter == nil {
-		regradeAfter = timestamp.Now()
-	} else {
-		regradeAfter = *options.RegradeAfter
+		now := timestamp.Now()
+		options.RegradeAfter = &now
 	}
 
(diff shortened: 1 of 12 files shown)


internal/grader/regrade.go
```diff
@@ -0,0 +1,229 @@
...
+	output := job.Run()
+	if output.Error != nil {
+		return nil, 0, fmt.Errorf("Failed to run regrade job '%s': '%w'.", output.ID, output.Error)
+	}
+
+	workErrors := make(map[string]string, len(output.WorkErrors))
+	for email, err := range output.WorkErrors {
+		workErrors[email] = err.Error()
+
+		logAttributes := make([]any, 3)
```
eriq-augustine, line 99: This allocation seems like it could be off to me.
 - Why 3?
 - The length is being set, when it seems like it should be the capacity.

internal/grader/regrade.go
```diff
@@ -0,0 +1,229 @@
...
+	return &regradeResult, len(output.RemainingItems), nil
+}
+
+func performSingleRegrade(courseUsers map[string]*model.CourseUser, assignment *model.Assignment, options RegradeOptions, email string) (*model.SubmissionHistoryItem, error) {
+	_, ok := courseUsers[email]
+	if !ok {
+		return nil, fmt.Errorf("Cannot regrade an unknown user: '%s'.", email)
+	}
+
+	// Get the students most recent submission.
```
eriq-augustine, line 121: s/students/student's/

eriq-augustine (changes requested):
Just some small and testing things.

eriq-augustine (approved):
lgtm!

PR #199, opened 2025-08-15
Added support for overriding the automatic type description.

Lucas-Ellenberger:
This PR adds support for overriding the automatic type description that is generated when describing all API endpoints.

One limitation of this PR is that types that are ignored by JSON (using the `-` JSON tag) cannot have their type description overriden. In the future, we will want to add support for this situation so API endpoints can display that they accept POST files.

Currently, the only types to include a type overrides are `Min*Role*` types.

### DIFF
--- a/internal/api/core/describe.go
+++ b/internal/api/core/describe.go
@@ -332,7 +332,7 @@ func DescribeType(customType reflect.Type, addType bool, info TypeInfoCache) (Fu
 		return FullTypeDescription{}, "", TypeInfoCache{}, err
 	}
 
-	overrideTypeID, overrideTypeDescription, err := getTypeOverride(description)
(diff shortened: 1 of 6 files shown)


internal/api/core/describe.go
```diff
@@ -13,6 +13,8 @@ import (
 	"github.com/edulinq/autograder/internal/util"
 )
 
+const TYPE_OVERRIDE_INDICATOR = "__TYPE_DESCRIPTION_OVERRIDE__"
```
eriq-augustine, line 16: The name is pretty specific, but we would ideally want overrides to apply to different things (like fields as well).

eriq-augustine (changes requested):
Discussed synchronously.
Here are some comments.

PR #205, opened 2026-01-22
Add grace time feature for assignment due dates

### LINKED ISSUE #204: Submissison Grace Time
Andi has brought up the concept of having "grace time" for due dates.
Something like allowing submissions up to 10 minutes past the due date without any penalty.
In an idea world, we would just change the due date to be a little latter.
But, this grace time can be very useful in preventing student complaints about "just being a second late".
This can also help compensate for clocks that may be a bit off.

This generally shouldn't be too hard to implement,
since our time checks are simple int comparisons:
https://github.com/edulinq/autograder-server/blob/main/internal/scoring/late.go#L318

Since we already have a structure for representing late policies,
it would just be adding an int to it representing the grace time:
https://github.com/edulinq/autograder-server/blob/main/internal/model/late.go#L22

aabhinavvvvvvv:
This PR fixes #204

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -832,7 +832,7 @@ All late policies share common fields:
 |------------------------|------------|----------|-------------|
 | `type`                 | String     | true     | The type of late policy being used. Valid values are: `baseline`, `constant-penalty`, `percentage-penalty`, and `late-days`. |
 | `reject-after-days`    | Integer    | false    | After this number of days past the assignment due date, do not accept any more submissions. Submissions past this time are fully ignored, they will not be neither stored nor stored. A zero (or no) value will result in submissions never being rejected for being late. |
-| `grace-mins`           | Integer    | false    | Grace time in minutes to add to the due date before calculating late penalties. Submissions within this grace period are not considered late. Defaults to 0 (no grace time). |
+| `grace-mins`           | Integer    | false    | Time in minutes to add to the due date before calculating late penalties. Submissions within this grace period are not considered late. This helps account for network lag, slow servers, or skewed clocks that might cause submissions to arrive slightly after the due date. Must be non-negative. Defaults to 0 (no grace time). |
 
 ### Baseline Late Policy (baseline)
 


internal/model/late.go
```diff
@@ -23,6 +23,7 @@ type LateGradingPolicy struct {
 	Type            LateGradingPolicyType `json:"type"`
 	Penalty         float64               `json:"penalty,omitempty"`
 	RejectAfterDays int                   `json:"reject-after-days,omitempty"`
+	GraceMinutes    int                   `json:"grace-minutes,omitempty"`
```
eriq-augustine: Use "grace-mins" instead of "grace-minutes".
This is not a strong convention or anything, we just use "-secs" in our config options elsewhere, so "mins" would be a little more consistent.

eriq-augustine (changes requested):
Looking real good!

You will also need to make sure to describe this option in the config:
https://github.com/edulinq/autograder-server/blob/main/docs/types.md#late-policy-latepolicy

internal/model/late.go
eriq-augustine: @aabhinavvvvvvv Instead of replying to comments that you have addressed,
you want to use the resolve button.
Only comment on comments that you have continuing questions or issues with.

docs/types.md
```diff
@@ -832,6 +832,7 @@ All late policies share common fields:
 |------------------------|------------|----------|-------------|
 | `type`                 | String     | true     | The type of late policy being used. Valid values are: `baseline`, `constant-penalty`, `percentage-penalty`, and `late-days`. |
 | `reject-after-days`    | Integer    | false    | After this number of days past the assignment due date, do not accept any more submissions. Submissions past this time are fully ignored, they will not be neither stored nor stored. A zero (or no) value will result in submissions never being rejected for being late. |
+| `grace-mins`           | Integer    | false    | Grace time in minutes to add to the due date before calculating late penalties. Submissions within this grace period are not considered late. Defaults to 0 (no grace time). |
```
eriq-augustine: Add in a clause about how negative values are not accepted.

PR #206, opened 2026-01-24
Add benevolent late day allocation feature

### LINKED ISSUE #139: Late Policy: Benevolent Late Days
Make a late policy (or add option to the LateDays policy) that does a "benevolent" distribution of late days that favors the student the most (recovers the maximum score).

aabhinavvvvvvv:
This PR fixes #139

### DIFF
--- a/docs/types.md
+++ b/docs/types.md
@@ -867,14 +867,31 @@ the number of actual days late, the number of remaining grace days the student h
 If the number of effective grace days is not sufficient to cover the number of actual days the assignment is late,
 then each remaining late day is penalized in the same way as the `percentage-penalty` late policy.
 
-| Name                 | Type    | Required | Description |
-|----------------------|---------|----------|-------------|
-| `penalty`            | Float   | true     | The proportion of the assignment's max points to apply as a penalty for each day of being late (after applying grace days). Must be in larger than 0.0 and less than or equal to 1.0. |
-| `max-late-days`      | Integer | true     | The maximum number of late days students may expend on this assignment. Must be less than or equal to `reject-after-days`. |
-| `late-days-lms-id`   | String  | false*   | The LMS ID for the assignment that will be used to track late days for each student. This assignment is usually not included in the final grade, but serves as a great place for students to view how many late days they have left. The autograder should have permissions to read and write this assignment. Instructors should populate it with the initial number of available late days for each student. |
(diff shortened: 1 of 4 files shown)


eriq-augustine (changes requested):
I did not do a full review because of some larger issues that may change things (noted below),
but the high level documentation is looking pretty good.

I think your procedure here is a bit overly complicated.
As we discussed before, there is a greedy solution to this.
The current late days allocation can also be solved greedily (assign to the earliest assignment with open late days slots).
So (unless I am missing something), I can see a simpler solution where both methods share more of the same code.

I don't think "benevolent allocation" is a good name for user facing configs.
I think it makes sense for when we are discussion how something work, but it is too ambiguous to use in actual code/config.
People would naturally ask "what does it mean to be benevolent?".
Instead, I think you should choose something more direct, like "optimize score" or "optimal allocation".

You should also note that we use standard go formatting (`go fmt`), and have a script for it: `scripts/format_code.sh`.

If any of these things confuse you, we can chat in real time.

eriq-augustine (changes requested):
I think you choose a good name for the option.

I have still only done a high-level review, as I think there are probably some more structural changes coming.

I am still concerned with the complexity.
As written, there are a lot of corner cases.
For example, what about when you are looking at assignment A, but then decide the late days should be reclaimed from assignment B and given to assignment C.
I believe you code will currently just ignore that (according to these lines):
 - https://github.com/aabhinavvvvvvv/autograder-server/blob/fbbef4abe15177d252a784b82f7ffd7f3a7c20ff/internal/scoring/late.go#L459
 - https://github.com/aabhinavvvvvvv/autograder-server/blob/fbbef4abe15177d252a784b82f7ffd7f3a7c20ff/internal/scoring/late.go#L217
(I also think there might be a bug around having an assignment that uses a different late policy (e.g. one assignment does not use late days),
but I have not looked at that deeply.)

You have put in a lot of infrastructure to deal with the collective nature of handling late days (i.e., you have to look at multiple assignments at the same time),
but that has made things complex.
You may want to consider fundamental changes to this,
like computing late days for all assignments at the same time.
Remember what I said before, if you need to refactor to do the right thing, you can.

You should also keep in mind that the number of available late days may change between scorings.
It is not uncommon for a prof to give a student an additional late day if they have some special circumstance.

eriq-augustine:
Based on the inactivity, I am closing this PR.
Feel free to make another PR if you want to try again.

PR #207, opened 2026-02-03
Avoid race when checking submission limits under concurrent requests

Aman-Cool:
# Avoid a race when checking submission limits under concurrent requests

## Summary

While setting up Lynx Grader locally, I was testing submissions from the same user made close together (for example, from multiple browser tabs near a deadline) and noticed a case where the submission limit can be exceeded.

This happens because the submission limit check currently runs before the per-user grading lock is acquired. When two requests arrive very close together, both can observe the same submission history and proceed.

This change moves the existing lock acquisition to occur before the rejection check so that the submission history read and grading decision are serialized for a given user and assignment.

---

## Observed Behavior

- Two concurrent submission requests from the same user can both pass the submission limit check.
- Both requests proceed to grading.
- The total number of submissions can exceed the configured maximum by one.

This was reproducible locally with parallel submission requests.

---

## Expected Behavior

- A user should not be able to exceed the configured maximum number of submissions.
- Concurrent submissions for the same user and assignment should be evaluated against a consistent submission history.

---

## Change

- Acquire the per-user grading lock before running the rejection check.
- Perform the submission limit validation while holding the lock.
- Continue with grading only after the limit has been checked under serialization.

Behavior for normal sequential submissions remains unchanged.

---

## Verification

- Build: Passed
- go vet: Passed
- Existing rejection-related tests passed locally:
  - `TestRejectSubmissionMaxAttempts`
  - `TestRejectLateSubmissionWithoutAllow`
  - `TestRejectWindowMaxMessage`
- Other test failures are due to the Docker daemon not being available on this machine and are unrelated to this change.

---

## Notes

I may be missing some historical or performance-related context around the current ordering, so please feel free to push back if there’s a preferred approach here.  
Happy to revise or re-scope if needed.

Thanks for taking a look.

### DIFF
--- a/internal/grader/grade.go
+++ b/internal/grader/grade.go
@@ -51,6 +51,17 @@ func GradeDefault(assignment *model.Assignment, submissionPath string, user stri
 // Full success is only when ((reject == nil) && (softGradingError == "") && (error == nil)).
 func Grade(ctx context.Context, assignment *model.Assignment, submissionPath string, user string, message string, options GradeOptions) (
 	*model.GradingResult, RejectReason, string, error) {
+	gradingKey := fmt.Sprintf("%s::%s::%s", assignment.GetCourse().GetID(), assignment.GetID(), user)
+
+	// Get the grading start time right before we acquire the user's lock.
+	startTimestamp := timestamp.Now()
+
+	// Ensure the user can only have one submission (of each assignment) running at a time.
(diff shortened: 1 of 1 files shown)


eriq-augustine (approved):
lgtm!

For small and straightforward issues like this, you can cut down on your PR description.
A much simpler description would suffice here.

Issue #203, opened 2026-01-04
Sandboxed File Preparation


eriq-augustine:
There are some file operations that happen outside of Docker (e.g., preparing for a Docker image build).
We do some sandboxing in these cases, but we could do much better.
Technically, these operations are only invoked from code that comes from someone managing a course,
but a bad actor could compromise them to get to us.

Some issues I am concerned with:
 - How long operations could take.
   - Could an attacker infinitely delay the update an of assignment by making an image build take/loop forever?
 - Output Size
   - We are currently vulnerable to zip bombs and even just really large downloads/clones.
 - General path sandboxing.
   - We already try to protect against this, but any syntactic/lexical sandboxing will always make me nervous vs a virtualization sandbox.

My proposed solution (which is just one of many) would be to create a Docker image that understands FileSpecs/FileOps and can do all the preparation for us.
(This image would live in its own repo.)
We can mount the context/base dir read-only on the image, and take just what results in the output dir/mount.
If I recall correctly, limiting the size of a docker container is a bit hard and requires host support (see #120).
But, we may be able to do some real fancy things (could we make an xfs disk on a virtual block device inside a Docker container?).
Hopefully, this should (ungracefully) stop containers from getting too large.
Additionally, we can wrap network/unarchive operations to try and more gracefully prevent using too much disk.
This would be similar to applications like [trickle](https://wiki.archlinux.org/title/Trickle),
but we can just handle it on the Go side and watch how many bytes are written.

Something like (read as pseudocode):
```golang
outFile, err := os.Create(outPath)
defer outFile.Close()

// This could also be a stream reader from some archive (e.g., we are unzipping).
response, err := http.Get(url)
defer response.Body.Close()

totalBytes := 0
for (totalBytes < BYTE_LIMIT):
    count, err := io.CopyN(outFile, response.Body, BUFFER_SIZE)
    totalBytes += count
```

Note that we are ignoring `response.ContentLength`,
because there is a variety of reasons (some malicious) where it can be off.

eriq-augustine:
We are always interested in doing things one step at a time instead of all at ones, we prefer doing things that way.
Just because something is one issue does not means it needs to be solved in one PR.

This effort should be started by running the file operations within their own docker container.

PR #208, opened 2026-02-03
Move submission file copy into Docker container for isolation

Aman-Cool:
## Summary

Move the submission file copy step from the host into a dedicated Docker container.

## Change

The submission directory copy now runs inside a small, dedicated Docker container instead of directly on the host. The source directory is mounted read-only and the destination is mounted read-write, establishing a clear isolation boundary for file preparation.

## Scope

This change is intentionally minimal:
- No resource limits are introduced
- No behavioral changes beyond where the copy runs
- Focused on a single file-prep step only

### DIFF
--- a/internal/docker/fileprep.go
+++ b/internal/docker/fileprep.go
@@ -1,0 +1,62 @@
+package docker
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/edulinq/autograder/internal/config"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+const FILE_PREP_IMAGE = "alpine:3.19"
+const FILE_PREP_TIMEOUT_SECS = 60
+
+// Copy directory contents inside a Docker container for isolation.
+// Source is mounted read-only, dest is mounted read-write.
+// Equivalent to: cp -r /source/* /dest/
+// Falls back to host-based copy when Docker is disabled.
+func CopyDirContentsInContainer(source string, dest string) error {
+	if config.DOCKER_DISABLE.Get() {
+		return util.CopyDirContents(source, dest)
+	}
+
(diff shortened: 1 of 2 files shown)


eriq-augustine (changes requested):
There are several core issues with this PR.
I think you may not understand the key issue here.

 - As per the issue (#203), file preparation should have its own image.
 - What you have here didn't sandbox anything. You just copied already collected files.
 - We are not concerned with sandboxing the submissions (which are already sent limited in their size).
 - Your timeout for prepping files is way to strict.

Based on these issues, you may not be ready for a medium issue.
You seemed to really rush this (based on the time between this PR and when you contacted me).
I suggest that you take a step back and decide if you want to take up another issue or slow down and really understand this one.

If you decide to carry on with this issue, we may need to chat synchronously.

Aman-Cool:
Thanks for taking the time to write this out — I really appreciate the direct feedback.

I see now that I misunderstood the core of #203. I scoped “file preparation” too narrowly and focused on submission copying, rather than the broader and riskier prep operations (downloads, extraction, build context prep, etc.) that happen before submissions are involved. That’s on me.

I’d also like to clarify one point on timing: I’d been reading through the issue and surrounding code for a while before reaching out, and the quick turnaround on the PR was more about trying to be responsive than rushing through the problem. That said, I agree I should have slowed down and aligned more clearly on scope before putting code up.

Given this, I agree this PR doesn’t meaningfully address the concerns raised in the issue. I’m happy to close it and take a step back to properly understand the prep pipeline and intended direction before attempting anything further.

If you think it would be helpful at some point, I’d be open to syncing to make sure I’m aligned before taking another pass — but no rush at all on my end.

PR #209, opened 2026-02-08
Fix docs: Docker graders should write to `result.json`, not `results.json`

Aman-Cool:
Fix documentation to consistently reference `result.json` as the Docker grader output file.

The server expects `result.json`, but parts of the docs incorrectly mention `results.json`, which causes grading to fail if followed.

Docs-only change. No code or behavior changes.

### DIFF
--- a/docs/tutorials/create-a-course.md
+++ b/docs/tutorials/create-a-course.md
@@ -259,7 +259,7 @@ If you put in your invocation manually, it would be:
 ```json
 {
     ... the rest of an assignment object ...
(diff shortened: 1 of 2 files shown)


internal/docker/fileprep.go
eriq-augustine: This file is outside the scope of your PR.

eriq-augustine (changes requested):
Good catch.

As a noted in one of your previous PRs,
you should go for a more concise PR message/description for simple changes like this.

eriq-augustine (approved):
lgtm!

PR #214, opened 2026-02-18
 Better API Response Logging

### LINKED ISSUE #212: Better API Response Logging
We currently don't log API responses.
Ideally, we would want to log them as TRACE.

However, the trouble here is when handling very large responses (like files/images).
We don't want to send MBs to the log.

We should find some way to log large objects (usually (base64) strings) safely.
Some possibilities:
 - Make a custom type (`LongLogString`?) that is an alias for `string`, but with a custom String() method that truncates long strings.
 - Do something with LogAttributes().
 - Handle all long strings specially in log.

I am partial to the custom type, since it is pretty direct and can potentially be used in other places.

dawgdevv:
Implements LongLogString type with custom String() method that truncates strings longer than 1000 characters. This prevents large API responses (like base64-encoded files/images) from flooding logs.

- Format: `content... [truncated X more bytes]`
- Works with `fmt.Sprintf("%s", value)`
- Explicit `.String()` call required in logging context

 Changes
- `internal/log/longlogstring.go`: New type with truncation logic
- `internal/log/longlogstring_test.go`: Unit tests for truncation behavior
- `internal/api/core/routing.go`: Add TRACE log in `sendAPIResponse()`

 Testing
All `LongLogString` tests pass:
- Short strings: no truncation
- Long strings: truncated with marker
- Empty strings: handled correctly
- Exact threshold: no truncation

Let me know if have to do some changes happy to do them

Fixes #212


internal/log/longlogstring.go
```diff
@@ -0,0 +1,17 @@
+package log
+
+import "fmt"
+
+type LongLogString string
+
+const DEFAULT_MAX_LOG_STRING_LENGTH = 1000
```
eriq-augustine: The name is misleading, since there is no way to change this value
(so the "default" prefix doesn't make sense).

internal/log/longlogstring.go
```diff
@@ -0,0 +1,17 @@
...
+
+func (s LongLogString) String() string {
+	str := string(s)
+	if len(str) > DEFAULT_MAX_LOG_STRING_LENGTH {
+		return fmt.Sprintf("%s... [truncated %d more bytes]",
+			str[:DEFAULT_MAX_LOG_STRING_LENGTH],
+			len(str)-DEFAULT_MAX_LOG_STRING_LENGTH)
+	}
+	return str
```
eriq-augustine: Put a newline above this.

internal/log/longlogstring_test.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			lls := LongLogString(tc.input)
+			result := lls.String()
+
+			if result != tc.expected {
+				t.Errorf("Expected length %d, got length %d", len(tc.expected), len(result))
+				t.Errorf("Expected: %q", tc.expected)
+				t.Errorf("Got: %q", result)
+			}
```
eriq-augustine: Errors are not prints, one error should be one statement/call.

eriq-augustine (changes requested):
Good effort.

There are several core design things that need to be addressed
(all in comments).

One thing that you need to be very careful of is that it looks like you didn't look at almost any other code in this project.
Make sure to read through some of the other code to see how we do things.

Note that I generally only make a comment once, even if it appears multiple times.
So, check all your code against every comment.

It would also be great to include a test that is in the context of an API endpoint (since that was the original inspiration for #214).

dawgdevv:
> Good effort.
> 
> There are several core design things that need to be addressed (all in comments).
> 
> One thing that you need to be very careful of is that it looks like you didn't look at almost any other code in this project. Make sure to read through some of the other code to see how we do things.
> 
> Note that I generally only make a comment once, even if it appears multiple times. So, check all your code against every comment.
> 
> It would also be great to include a test that is in the context of an API endpoint (since that was the original inspiration for #214).

will remember that for future contribution thanks for the feedback

internal/api/core/routing.go
```diff
@@ -170,6 +170,14 @@ func sendAPIResponse(apiRequest ValidAPIRequest, response http.ResponseWriter,
 
 	stats.AsyncStoreMetric(&metric)
 
+	// Log API Response at TRACE level (payload truncated via LongLogString).
+	log.Trace("API Response",
+		log.NewAttr("endpoint", endpoint),
+		log.NewAttr("status", apiResponse.HTTPStatus),
+		log.NewAttr("success", apiResponse.Success),
+		log.NewAttr("duration_ms", (apiResponse.EndTimestamp-startTime).ToMSecs()),
+		log.NewAttr("payload", log.LongLogString(payload).String()))
```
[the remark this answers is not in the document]
dawgdevv: func (this *APIResponse) LogValue() []*log.Attr {
	return []*log.Attr{
		log.NewAttr("id", this.ID),
		log.NewAttr("status", this.HTTPStatus),
		log.NewAttr("success", this.Success),
		log.NewAttr("duration-ms", this.EndTimestamp-this.StartTimestamp),
	}
}

i have implemented in response.go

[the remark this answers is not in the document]
dawgdevv: still working will share an update soon

eriq-augustine:
@dawgdevv

No need to comment on comments that require no further discussion.
Keep in mind what things will look like for the reviewer.
When they re-review a PR, they should only see things that need action on their part.
So, all comments should be either resolved "via the resolve button", or commented on if further discussion is required.
Also, reviewers may get an email for each comment you make.

Note that this also applies when a reviewer asks a question in a comment,
and you agree with the premise of the question and make changes accordingly.
For example, if a review says: "Shouldn't this be 1?",
then you can just resolve if you agree that it should be 1 and you make that change.
Only comment if further discussion is required.

internal/types/longstring_test.go
```diff
@@ -0,0 +1,69 @@
...
+			name:     "Short string - no truncation",
+			input:    "hello world",
+			maxLen:   100,
+			expected: "hello world",
+		},
+		{
+			name:     "Long string - truncated with hash",
+			input:    strings.Repeat("a", 2000),
+			maxLen:   100,
+			expected: "truncated",
```
eriq-augustine: No.

This test has a deterministic output, match the exact output.

internal/util/clip.go
```diff
@@ -0,0 +1,101 @@
...
+// Below this, we just return an empty string.
+const minClipLength = 10
+
+// ClipString truncates a string to maxLength and returns it.
+// If the string is longer than maxLength, it returns a truncated version
+// with a suffix indicating how much was truncated.
+// Returns empty string if maxLength is too small to fit any meaningful content.
+func ClipString(text string, maxLength int) string {
+	if maxLength < minClipLength {
+		return ""
```
eriq-augustine: !

This behavior seems pretty dangerous/unexpected.

internal/util/clip_test.go
```diff
@@ -0,0 +1,279 @@
...
+		test.Error("Different strings should produce different clipped results.")
+	}
+
+	// Check lengths
+	if utf8.RuneCountInString(result1) > maxLen || utf8.RuneCountInString(result2) > maxLen {
+		test.Error("Results should not exceed maxLength.")
+	}
+}
+
+func TestTruncateRunes(test *testing.T) {
```
eriq-augustine: This seems like the same thing as TestClipString,
they should probably be the same function.

internal/util/clip_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	for _, testCase := range testCases {
+		test.Run(testCase.name, func(test *testing.T) {
+			result := ClipString(testCase.input, testCase.maxLen)
+
+			if testCase.expected == "truncated" {
+				// Check that result contains truncation marker
+				if !strings.Contains(result, "...") {
+					test.Errorf("Test case '%s': Expected truncation marker, got: %q.", testCase.name, result)
```
eriq-augustine: !!

Look at how our other test files handle looping over test cases.

eriq-augustine (changes requested):
There are still some serious issues here.

The core of the clipping code seems fine,
but there needs to be some work around it.

As I mentioned before, please look at how we do tests in other files.
You are closer, but still missing key points.

At the API-level, things are not good.
You are STILL missing an API-level tests.
How you use the string in the API response is completely off and it's like you didn't look at what I referenced previously.

If you are connected with OSRE/OSPO, contact me in Slack.

eriq-augustine:
@dawgdevv

Additionally, you don't need a huge comment summarizing your changes for each re-review.
I should already know the summary of your changes based off the resolved/responded comments.
The summary makes more work for your reviewer.

eriq-augustine:
@dawgdevv 
Are you still pursuing this?

dawgdevv:
> @dawgdevv Are you still pursuing this?

Yes I will be working on this got caught up in some other work

Issue #198, opened 2025-08-11
API Endpoint: Fetch Gradebook


eriq-augustine:
It would be nice to have a single endpoint that fetches a full "gradebook"
(the current score for each user/assignment).
Note that this API-level functionality could be achieved with multiple calls
(one to fetch assignments and then one for each assignment),
but we should just have one with everything built in.

PR #196, opened 2025-07-23
Course report endpoint

maozblan:
This PR adds an API endpoint to the course scoring report, returning a pointer to the course report created using the functionality in `internal/report/course.go`.

PR #228, opened 2026-07-13
Added an endpoint to fetch a course's gradebook

pshayes06:
Added a courses/gradebook endpoint to address issue #198. Allows clients to directly retrieve the most recent submission for each user on each assignment of a course, without having to make multiple calls on their end.

Follows existing work seen in courses/assignments/submissions/fetch/course/scores which grabs recent submissions for all users on an assignment. By making a call to GetAssignments() we can build a response by making a recent submission call on every assignment in the course. The response maps assignment IDs to a map of user emails to a SubmissionHistoryItem. Assignments with no matching users still appear as keys. The shown users are filtered using the same approach seen in scores, based on a combination of CourseUserReferences (if not specified, will just match all users).

Most of the tests are pulled from scores again, checking for correct behavior based on permissions, role escalation, and filtering.

### DIFF
--- a/internal/api/courses/gradebook.go
+++ b/internal/api/courses/gradebook.go
@@ -37,7 +37,7 @@ func HandleGradebook(request *GradebookRequest) (*GradebookResponse, *core.APIEr
 	}
 
 	if len(request.TargetAssignments) == 0 {
-		for id := range request.Course.GetAssignments() {
+		for id, _ := range request.Course.GetAssignments() {
 			request.TargetAssignments = append(request.TargetAssignments, id)
 		}
 	}
--- a/internal/api/courses/gradebook.go
+++ b/internal/api/courses/gradebook.go
@@ -10,12 +10,14 @@ type GradebookRequest struct {
 	core.APIRequestCourseUserContext
 	core.MinCourseRoleGrader
 
-	// If not empty, filter results to matching users.
-	// Unknown users will not raise an error.
+	// The users to include in the gradebook.
+	// If nothing is specified, all course users will be included.
+	// Unknown users are ignored and will not raise an error.
 	TargetUsers []model.CourseUserReference `json:"target-users"`
 
-	// If not empty, filter results to matching assignments.
-	// Unknown and malformed assignments will not raise an error.
+	// The assignments to include in the gradebook.
+	// If nothing is specified, all course assignments will be included.
+	// Unknown and malformed assignments are ignored and will not raise an error.
 	TargetAssignments []string `json:"target-assignments"`
 }
 
--- a/internal/api/courses/gradebook.go
+++ b/internal/api/courses/gradebook.go
@@ -1,0 +1,42 @@
+package courses
+
+import (
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+)
+
+type GradebookRequest struct {
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleGrader
+
+	TargetUsers []model.CourseUserReference `json:"target-users"`
+}
+
+type GradebookResponse struct {
+	Gradebook map[string]map[string]*model.SubmissionHistoryItem `json:"gradebook"`
+}
+
+// Get gradebook (most recent score for each user on each assignment) for a course.
+func HandleGradebook(request *GradebookRequest) (*GradebookResponse, *core.APIError) {
+
+	if len(request.TargetUsers) == 0 {
+		request.TargetUsers = model.NewAllCourseUserReference()
+	}
+
+	reference, err := model.ParseCourseUserReferences(request.TargetUsers)
+	if err != nil {
+		return nil, core.NewBadRequestError("-644", request, "Failed to parse target users.").Err(err)
+	}
+
+	gradebook := make(map[string]map[string]*model.SubmissionHistoryItem)
+	for id, assignment := range request.Course.GetAssignments() {
+		submissionInfos, err := db.GetRecentSubmissionSurvey(assignment, reference)
+		if err != nil {
+			return nil, core.NewInternalError("-645", request, "Failed to get submission summaries.").Err(err)
+		}
+		gradebook[id] = submissionInfos
+	}
+
+	return &GradebookResponse{gradebook}, nil
+}
--- a/internal/api/courses/gradebook.go
+++ b/internal/api/courses/gradebook.go
@@ -1,8 +1,6 @@
 package courses
 
 import (
-	"fmt"
-
 	"github.com/edulinq/autograder/internal/api/core"
 	"github.com/edulinq/autograder/internal/db"
 	"github.com/edulinq/autograder/internal/model"
@@ -12,15 +10,20 @@ type GradebookRequest struct {
 	core.APIRequestCourseUserContext
 	core.MinCourseRoleGrader
 
-	TargetUsers       []model.CourseUserReference `json:"target-users"`
-	TargetAssignments []string                    `json:"target-assignments"`
+	// If not empty, filter results to matching users.
+	// Unknown users will not raise an error
+	TargetUsers []model.CourseUserReference `json:"target-users"`
+
+	// If not empty, filter results to matching assignments.
+	// Unknown and malformed assignments will not raise an error
+	TargetAssignments []string `json:"target-assignments"`
 }
 
 type GradebookResponse struct {
 	Gradebook map[string]map[string]*model.SubmissionHistoryItem `json:"gradebook"`
 }
 
-// Get gradebook (most recent score for each user on each assignment) for a course.
+// Get a gradebook (most recent score for each user on each assignment) for a course.
 func HandleGradebook(request *GradebookRequest) (*GradebookResponse, *core.APIError) {
 	if len(request.TargetUsers) == 0 {
 		request.TargetUsers = model.NewAllCourseUserReference()
@@ -41,14 +44,12 @@ func HandleGradebook(request *GradebookRequest) (*GradebookResponse, *core.APIEr
 	for _, id := range request.TargetAssignments {
 		assignment := request.Course.GetAssignment(id)
 		if assignment == nil {
-			return nil, core.NewBadRequestError("-645", request,
-				fmt.Sprintf("Could not find assignment: '%s'.", id))
-
+			continue
 		}
 
 		submissionInfos, err := db.GetRecentSubmissionSurvey(assignment, reference)
 		if err != nil {
-			return nil, core.NewInternalError("-646", request, "Failed to get submission summaries.").Err(err)
+			return nil, core.NewInternalError("-645", request, "Failed to get submission summaries.").Err(err)
 		}
 
 		gradebook[assignment.GetID()] = submissionInfos
--- a/internal/api/courses/gradebook.go
+++ b/internal/api/courses/gradebook.go
@@ -11,11 +11,11 @@ type GradebookRequest struct {
 	core.MinCourseRoleGrader
 
 	// If not empty, filter results to matching users.
-	// Unknown users will not raise an error
+	// Unknown users will not raise an error.
 	TargetUsers []model.CourseUserReference `json:"target-users"`
 
 	// If not empty, filter results to matching assignments.
-	// Unknown and malformed assignments will not raise an error
+	// Unknown and malformed assignments will not raise an error.
 	TargetAssignments []string `json:"target-assignments"`
 }
 
--- a/internal/api/courses/gradebook.go
+++ b/internal/api/courses/gradebook.go
@@ -1,6 +1,8 @@
 package courses
 
 import (
+	"fmt"
+
 	"github.com/edulinq/autograder/internal/api/core"
 	"github.com/edulinq/autograder/internal/db"
 	"github.com/edulinq/autograder/internal/model"
@@ -10,7 +12,8 @@ type GradebookRequest struct {
 	core.APIRequestCourseUserContext
 	core.MinCourseRoleGrader
 
-	TargetUsers []model.CourseUserReference `json:"target-users"`
+	TargetUsers       []model.CourseUserReference `json:"target-users"`
+	TargetAssignments []string                    `json:"target-assignments"`
 }
 
 type GradebookResponse struct {
@@ -19,7 +22,6 @@ type GradebookResponse struct {
 
 // Get gradebook (most recent score for each user on each assignment) for a course.
 func HandleGradebook(request *GradebookRequest) (*GradebookResponse, *core.APIError) {
-
 	if len(request.TargetUsers) == 0 {
 		request.TargetUsers = model.NewAllCourseUserReference()
 	}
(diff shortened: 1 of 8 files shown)


internal/api/courses/gradebook.go
```diff
@@ -0,0 +1,42 @@
...
+	TargetUsers []model.CourseUserReference `json:"target-users"`
+}
+
+type GradebookResponse struct {
+	Gradebook map[string]map[string]*model.SubmissionHistoryItem `json:"gradebook"`
+}
+
+// Get gradebook (most recent score for each user on each assignment) for a course.
+func HandleGradebook(request *GradebookRequest) (*GradebookResponse, *core.APIError) {
+
```
eriq-augustine: Unnecessary newline.

internal/api/courses/gradebook.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+)
+
+type GradebookRequest struct {
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleGrader
+
+	TargetUsers []model.CourseUserReference `json:"target-users"`
```
eriq-augustine, line 16: You should also be able to filter assignments as well as users.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"maps"
+	"testing"
+
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/model"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestGradebook(test *testing.T) {
+
```
eriq-augustine, line 26: Unnecessary newline.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,128 @@
...
+	allUsers := map[string]map[string]string{
```
eriq-augustine: I would make this variable name more relevant.
Something like `fullGradebook`?

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,128 @@
...
+			if !ok {
+				test.Errorf("Case %d: Missing assignment '%s'.", i, assignmentID)
+				continue
+			}
+
+			if !maps.Equal(expectedIDs, actualIDs) {
+				test.Errorf("Case %d: Submission IDs do not match for assignment '%s'. Expected: '%+v', actual: '%+v'.",
+					i, assignmentID, expectedIDs, actualIDs)
+			}
+		}
```
eriq-augustine: It's not that the code you have here is incorrect,
but can't you just use reflect.DeepEqual()?
For user-facing code, the extra descriptions can be useful;
but for testing code, I think the simplicity would be better.

internal/api/courses/routes.go
```diff
@@ -15,6 +15,7 @@ import (
 var baseRoutes []core.Route = []core.Route{
 	core.MustNewAPIRoute(`courses/get`, HandleGet),
 	core.MustNewAPIRoute(`courses/list`, HandleList),
+	core.MustNewAPIRoute(`courses/gradebook`, HandleGradebook),
```
eriq-augustine: Scoot this up to maintain alphabetic order.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	testCases := []struct {
+		email       string
+		targetUsers []model.CourseUserReference
+		locator     string
+		expected    map[string]map[string]string
+	}{
+
+		// Valid Permissions
+		{"course-grader", nil, "", allUsers},
```
eriq-augustine: Once you implement the assignment filtering (comment above),
you may want to switch up these test cases to have each value on their own line (or multiple lines).
It's already a little crowded on the line (still bearable),
but another slice will really crowd things.

eriq-augustine (changes requested):
A fantastic start!

internal/api/courses/gradebook.go
pshayes06: For this implementation, would it be better to create a simple filter inside gradebook.go to skip every assignment ID passed through the request (such as ["hw0", "hw2"], and include all if empty)? There is also the option of creating something similar to model/user_reference_course.go, but for assignments. It could include filtering through similar means like '*' and '-hw0'.

eriq-augustine: Keep it simple, especially to start.

internal/api/courses/gradebook.go
```diff
@@ -0,0 +1,58 @@
...
+
+	TargetUsers       []model.CourseUserReference `json:"target-users"`
+	TargetAssignments []string                    `json:"target-assignments"`
+}
+
+type GradebookResponse struct {
+	Gradebook map[string]map[string]*model.SubmissionHistoryItem `json:"gradebook"`
+}
+
+// Get gradebook (most recent score for each user on each assignment) for a course.
```
eriq-augustine: s/Get gradebook/Get a gradebook/

(If you are not familiar with this syntax,
I am using a sed-style find/replace command to suggest an edit.
`s` indicates find/replace, `/` separates the parts of the command, the first piece of text is what to find, and the second piece of text is the replacement.)

internal/api/courses/gradebook.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+
+	gradebook := make(map[string]map[string]*model.SubmissionHistoryItem, len(request.TargetAssignments))
+	for _, id := range request.TargetAssignments {
+		assignment := request.Course.GetAssignment(id)
+		if assignment == nil {
+			return nil, core.NewBadRequestError("-645", request,
+				fmt.Sprintf("Could not find assignment: '%s'.", id))
+
```
eriq-augustine: Stray newline.

Be careful, you had a number of extra newlines in your previous pass.

internal/api/courses/gradebook.go
```diff
@@ -0,0 +1,58 @@
...
+			request.TargetAssignments = append(request.TargetAssignments, id)
+		}
+	}
+
+	gradebook := make(map[string]map[string]*model.SubmissionHistoryItem, len(request.TargetAssignments))
+	for _, id := range request.TargetAssignments {
+		assignment := request.Course.GetAssignment(id)
+		if assignment == nil {
+			return nil, core.NewBadRequestError("-645", request,
+				fmt.Sprintf("Could not find assignment: '%s'.", id))
```
eriq-augustine: The semantics here seem a bit off to me.
If a user is not found, we will just ignore them (not including malformed roles).
But if an assignment is not found, we will raise an error.
Of course, assignments are more structural than users, but it still feels a bit off.

Whatever we end up doing, make sure to document it.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,209 @@
...
+			"course-grader",
+			nil,
+			nil,
+			"",
+			fullGradebook,
+		},
+		{
+			"course-admin",
+			[]model.CourseUserReference{},
+			nil,
```
eriq-augustine: For this test case, use an empty slice here as well.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,209 @@
...
+			"",
+			fullGradebook,
+		},
+		{
+			"server-admin",
+			[]model.CourseUserReference{"student"},
+			nil,
+			"",
+			studentOnlyGradebook,
+		},
```
eriq-augustine: I don't think you need a test for role escalation (and you definitely don't need two).
Since role escalation is a core concept to the API and used in other places,
you don't need to have tests especially for it in every endpoint.

I know it may seem strange to be asking for fewer tests,
but like non-test code, tests also carry a maintenance burden with them.
Tests need to be maintained, updated, and documented.
Of course we want enough tests, but unnecessary tests will increase that burden without a good reason.

eriq-augustine (changes requested):
Looking pretty good.

internal/api/courses/gradebook.go
pshayes06: To clarify, do you see this as more of a consistency issue? I could try working on an error checker for specific targetUsers being passed, though it will touch other areas of code since other endpoints rely on this. Otherwise I could go through with what I have for now and comment what would and wouldn't raise an error for unknown inputs.

eriq-augustine: | To clarify, do you see this as more of a consistency issue?

I don't know what you mean by "consistency issue".
If you mean that the semantics of this endpoint are inconsistent, then yes.
How you handle errors with your two different types of inputs is inconsistent.

Consistency is very important for an API,
so we will want to achieve that here.
I have opinions on which what I think would be best, but I want to give you the opportunity to work it out
(since I know you are a student and we can use this as a learning opportunity).

pshayes06: I'm leaning towards consistently ignoring both for simplicity, and to align with the existing filtering model where matching nothing is a valid outcome, though clients will need to ensure they aren't missing mis-typed assignments.

internal/api/courses/gradebook.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+)
+
+type GradebookRequest struct {
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleGrader
+
+	// If not empty, filter results to matching users.
+	// Unknown users will not raise an error.
```
eriq-augustine: Since these descriptions automatically make it into the API documentation (see you `api.json`),
we will want to expand this description a bit.
Start with a simple overview sentence, and then you can have you specific documentation.

internal/api/courses/gradebook.go
```diff
@@ -0,0 +1,59 @@
...
+
+	reference, err := model.ParseCourseUserReferences(request.TargetUsers)
+	if err != nil {
+		return nil, core.NewBadRequestError("-644", request, "Failed to parse target users.").Err(err)
+	}
+
+	if len(request.TargetAssignments) == 0 {
+		for id := range request.Course.GetAssignments() {
+			request.TargetAssignments = append(request.TargetAssignments, id)
+		}
```
eriq-augustine: Does this work instead of the loop:
```
request.TargetAssignments = append(request.TargetAssignments, request.Course.GetAssignments()...)
```

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	testCases := []struct {
+		email             string
+		targetUsers       []model.CourseUserReference
+		targetAssignments []string
+		locator           string
+		expected          map[string]map[string]string
+	}{
+		// Valid Permissions
```
eriq-augustine: I would use a different comment here.
Since most of these cases have valid permissions,
this comment is not super helpful.
Maybe something about a full query or full gradebook.

eriq-augustine (changes requested):
Almost there!

Remember that all conversations that do not require input from the reviewer should be resolved.
You left one conversation (that required no feedback) unresolved.

The next step after this PR is approved (but NOT merged),
is to write the Python interface for this endpoint.
We want the Python interface we maintain to be a complete reference,
so it needs functionality for all endpoints.
An example of this for another endpoint:
 - https://github.com/edulinq/autograder-server/pull/196
 - https://github.com/edulinq/autograder-py/pull/59

Once both are approved, then I will merge them both (and handle committing test data).

internal/api/courses/gradebook.go
pshayes06: I'm not able to directly replace the loop with that because GetAssignments() returns a map instead of slice so I can't unpack it. I'm only appending the keys of the map, so if I were to do a one-liner it looks like i can use Slices.collect() with maps.Keys() to create a slice of the keys. However those are only available on a later version of Go. Should I just stick with my loop then?

eriq-augustine: In that case, use the two-element variant of `range`.
You want to always use the two-element variant for maps to avoid confusion (like here).

```
for id, _ := range ...
```

pshayes06:
I noticed that my assignment filtering test may not be too thorough since the course101 test data only has a single assignment. Would you recommend I use the course-languages data instead since it has multiple assignments? I could include these changes along with your recent round of suggestions.

eriq-augustine:
Yeah, you could use course-languages and then insert some fake submissions.
We don't have too many tests that do that, but you can see an example in the analysis testing:
https://github.com/edulinq/autograder-server/blob/main/internal/analysis/test.go#L56

Then you can have submissions for multiple assignments and multiple users.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,276 @@
...
+	"github.com/edulinq/autograder/internal/api/core"
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+// Flatten a gradebook response into submission IDs rather than comparing whole structs.
+func flattenGradebook(content any) map[string]map[string]string {
+	var responseContent GradebookResponse
+	util.MustJSONFromString(util.MustToJSON(content), &responseContent)
```
eriq-augustine: Pass either the `GradebookResponse` or `map[string]map[string]*model.SubmissionHistoryItem` into this function.

Passing an `any` is not recommended here since it is hard for people to see then intentionality.
The function (and description) says that it flattens a gradebook, but it actually pulls out the typed gradebook and then flattens it.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,276 @@
...
+		actual := flattenGradebook(response.Content)
+
+		if !reflect.DeepEqual(testCase.expected, actual) {
+			test.Errorf("Case %d: Unexpected gradebook. Expected: '%s', actual: '%s'.",
+				i, util.MustToJSONIndent(testCase.expected), util.MustToJSONIndent(actual))
+		}
+	}
+}
+
+func TestGradebookAssignmentFiltering(test *testing.T) {
```
eriq-augustine: I'm not seeing much of a reason to split these tests into two different functions.
Like I mentioned a couple passes ago, tests have their own maintenance burden.
Maintaining a single test function (with slightly more test cases) is generally easier than maintaining two test functions.

Of course you are free to push back on this if I am missing something,
but I think these functions would be simple to merge.

internal/api/courses/gradebook/fetch.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	if len(request.TargetAssignments) == 0 {
+		for id, _ := range request.Course.GetAssignments() {
+			request.TargetAssignments = append(request.TargetAssignments, id)
+		}
+	}
+
+	gradebook := make(map[string]map[string]*model.SubmissionHistoryItem, len(request.TargetAssignments))
+	for _, id := range request.TargetAssignments {
```
eriq-augustine, line 46: For slices/arrays, you can use the single-receiver range variant.

(We want to always use the two-receiver variant for maps,
but lists we can use the single if we don't need the index.
In this way, it should be fairly obvious to the reader when something is a slice and when it is a map.)

eriq-augustine (changes requested):
I think the tests need a bit of massaging, but things are looking pretty good.

internal/api/courses/gradebook/fetch.go
pshayes06: I'm not completely sure but are you suggesting something like **for id := range request.TargetAssignments** ? I thought that would cause id to hold the index instead of the value.

eriq-augustine: Sorry about that, you are correct.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,251 @@
...
+			ShortID:      "1234567890",
+			CourseID:     "course-languages",
+			AssignmentID: "cpp",
+			User:         "course-student@test.edulinq.org",
+		}},
+	}
+
+	for _, submission := range testSubmissions {
+		assignment := db.MustGetAssignment(submission.Info.CourseID, submission.Info.AssignmentID)
+		if err := db.SaveSubmission(assignment, submission); err != nil {
```
eriq-augustine: We generally don't do multi-statement lines in this project.
This is especially true on error checks, so it is easier to spot an error check when reading.

internal/api/courses/gradebook_test.go
```diff
@@ -0,0 +1,251 @@
...
+			continue
+		}
+
+		var responseContent GradebookResponse
+		util.MustJSONFromString(util.MustToJSON(response.Content), &responseContent)
+
+		actual := flattenGradebook(responseContent.Gradebook)
+
+		if !reflect.DeepEqual(testCase.expected, actual) {
+			test.Errorf("Case %d: Unexpected gradebook. Expected: '%s', actual: '%s'.",
```
eriq-augustine: (nit)

s/actual/Actual/

(If you are unfamiliar, a "nit" is a small (usually style) thing that is so minor that on its own would not stop a PR from going through.
But if there are other things to fix, then you may as well also fix the nits.)

eriq-augustine (changes requested):
Pretty much there, only some super small things and one location change.

I want to move this endpoint a bit to protect us for when we want to add the ability to edit/upload a gradebook.
Let's move this from `courses/gradebook` to `courses/gradebook/fetch`.
When moving things, don't forget to update internal/api/courses/routes.go to include the routes for the new gradebook sub-package.

eriq-augustine (approved):
LGTM!

We'll get this merged in once the sibling Python PR gets approved.

Issue #28, opened 2023-10-31
LMS Score Comment Author


eriq-augustine:
Add an "author" field to LMS score comments.
This allows scores that are uploaded via an API call to get an author recorded.

pshayes06:
It looks like the current lms/upload/scores endpoint doesn't support comments, so I'm thinking of incorporating that so grade_data[<student_id>][text_comment] can be populated in updateAssignmentScores. To record an author, I was thinking of just embedding something like "author":"name" to the start of the score comment, where the name is derived from the user making the request. I'm assuming this author field is for students to view so this seemed like an option to deliver that information, but correct me if it's more for internal usage instead.

eriq-augustine:
I think we probably want to make some changes to this issue.

For uploading grades directly to Canvas (the LMS),
we will probably want to also switch to the LMS Toolkit.
Meaning the primary functionality here is mostly obviated by #210.

So, I don't think an author field will be useful (since should almost always be the submitting student).
However, I think it would make sense to include a proxy user (if any) in the submission comment.
A proxy submission is the only time a submission's score is not sent in by the student (from the autograder, at least).
So, I can see adding in the proxy user and proxy time to the score comment.

eriq-augustine:
The spirit of this was resolved in 17b08b2d8150e585dd9793c98072c83eb2763c12.

It has changes a bit over time and the rest of the functionality will come naturally with the LMS Toolkit.

PR #230, opened 2026-07-29
Added proxy user and time to submission comments

pshayes06:
Added the proxy user and proxy time fields to the ScoringInfo type to address issue #28. The scoring and upload process can now take the proxy user and time from submissions and include them in the uploaded comment. Added a test for the uploading of these fields in scoring/course_test.

### DIFF
--- a/internal/model/scoring.go
+++ b/internal/model/scoring.go
@@ -7,15 +7,17 @@ import (
 const SCORING_INFO_STRUCT_VERSION = "1.0.0"
 
 type ScoringInfo struct {
-	ID             string              `json:"id"`
-	SubmissionTime timestamp.Timestamp `json:"submission-time"`
-	UploadTime     timestamp.Timestamp `json:"upload-time"`
-	RawScore       float64             `json:"raw-score"`
-	Score          float64             `json:"score"`
-	Lock           bool                `json:"lock"`
-	LateDayUsage   int                 `json:"late-date-usage"`
-	NumDaysLate    int                 `json:"num-days-late"`
-	Reject         bool                `json:"reject"`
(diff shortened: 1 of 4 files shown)


internal/model/scoring.go
```diff
@@ -41,7 +43,17 @@ func (this *ScoringInfo) Equal(other *ScoringInfo) bool {
 		return false
 	}
 
-	return (this.ID == other.ID &&
+	if this.ProxyTime != other.ProxyTime {
+		if (this.ProxyTime == nil) || (other.ProxyTime == nil) {
+			return false
+		}
```
eriq-augustine, line 49: Put a newline after this conditional.

We almost never want to have two `if`s right next to each other without any spacing.
It makes it too easy for a reader to mistake it as an `else if`.

internal/scoring/course_test.go
```diff
@@ -44,10 +44,36 @@ func TestCourseScoringEmptyStudent(test *testing.T) {
...
+
+	proxyTime := timestamp.FromMSecs(1697500000000)
+
+	submission, err := db.GetSubmissionContents(assignment, "course-student@test.edulinq.org", "1697406272")
+	if err != nil {
+		test.Fatalf("Failed to load submission: '%v'.", err)
+	}
+
+	submission.Info.ProxyUser = "course-grader@test.edulinq.org"
+	submission.Info.ProxyStartTime = &proxyTime
```
eriq-augustine: I would just use `expectedProxyTime` here.

eriq-augustine (changes requested):
Looking pretty good, just some small things.

eriq-augustine (approved):
LGTM!

Issue #164, opened 2025-03-30
Course Start and End Dates


eriq-augustine:
It would be great to have course start and end dates.
Outside of this period, a course would not:
 - Update (unless explicitly forced)
 - Run Tasks
 - Grade (all submissions rejected, but proxy submissions will be accepted)

Tasks that were already queued up will have to know how to skip themselves and not re-queue.
(We don't have to explicitly remove them from the task queue, just make sure they don't run).

The information could also be synced from the course's LMS.
We would want options for ensuring the correct syncing behavior, like:
 - `lms-sync-missing-course-options` -- Take options that are not in the config.
 - `lms-sync-overwrite-course-options` -- Overwrite options that are in the config.

This is a pretty easy issue, but does require touching several components.
So it's easy, but for a dev with some autograder experience.

pshayes06:
I've made changes to address the points before the LMS syncing part, and was wondering if I should open a PR with these current changes or wait in case you wanted the LMS changes added through the toolkit integration.

eriq-augustine:
@pshayes06
Let's chat about this synchronously (see email).

Issue #172, opened 2025-04-12
Administrative Tasks


eriq-augustine:
We currently have infrastructure to regularly run course-based tasks.
But, it would be great if we could also run more administrative server-based tasks,
like pruning docker or cleaning up some other resources.

The course-based tasks should probably fall within this infrastructure as well.
So, there would be some sort of abstract task infrastructure that can run both course-based and server-based tasks.

eriq-augustine:
There should be some task (or group of tasks) that manages the maintenance for each course.
Things like:
 - backups
 - active tasks (add/remove active tasks based on a course's active dates)

These are things that should not be controllable from the owner's side, only the server side.

PR #231, opened 2026-09-03
Course Inactivity

pshayes06:
Implemented the ability for authorized (course admin +) users to mark a course as inactive. Addresses issue #164, and can serve as an alternative for a hard-deletion of a course.

Currently, an inactive course cannot run tasks (will be skipped), and normal assignment submissions will be rejected. A course's inactivity can be configured through a start and end date explicitly written into the course config. A course outside the set window will be inactive. The alternative, which has priority over the date window, is a map of statuses set by users. These statuses map owners to their status (1 per owner), and have dedicated storage. The process for deciding on a single status to use is in model/course_statuses. 

There are new endpoints to set or remove course statuses, and the existing get endpoint for the course can now also retrieve a course's active status and the map of statuses. Test files exist for all these additions, which stay consistent with surrounding conventions and structure if possible.


internal/model/course.go
```diff
@@ -248,3 +260,20 @@ func (this *Course) GetTemplatesDir() string {
 func (this *Course) GetSourceConfigPath() string {
 	return filepath.Join(this.GetBaseSourceDir(), COURSE_CONFIG_FILENAME)
 }
+
+func (this *Course) IsActive(now timestamp.Timestamp) bool {
```
eriq-augustine, line 264: I think there should be two variants of this method, one with and one without a timestamp argument.
The one without the timesatmp is just a convenience one that checks for now.
Maybe `IsActiveNow() bool`?

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+// Ties are broken by the SetTime, giving priority to the most recently set status.
+func (this *CourseStatus) hasPriorityOver(other *CourseStatus) bool {
+	if this.Source != other.Source {
+		return this.Source > other.Source
+	}
+	return this.SetTime > other.SetTime
+}
+
+// Iterates through every status to determine which should actually be used
+func determineStatus(statuses map[string]*CourseStatus) *CourseStatus {
```
eriq-augustine, line 80: Expose a method (GetActiveStatus()?) for courses.
You can probably just move this to there.

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+// StatusSourceServer is for users with admin/owner privileges in the server.
+// StatusSourceAutomated is for automated processes.
+// StatusSourceRoot is for users with root privileges.
+const (
+	StatusSourceUnknown   StatusSource = 0
+	StatusSourceCourse                 = 10
+	StatusSourceServer                 = 20
+	StatusSourceAutomated              = 30
+	StatusSourceRoot                   = 40
+)
```
eriq-augustine, line 35: Move this to above the definition of CourseStatus.
(Of course this is not technically necessary, just makes it cleaner to read.)

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"github.com/edulinq/autograder/internal/timestamp"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+type CourseStatus struct {
+	Active bool         `json:"active"`
+	Source StatusSource `json:"source"`
+
+	// User who set the status
```
eriq-augustine, line 12: We generally make comments either complete sentences (with punctuation), or Title Case.

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+	Owner string `json:"owner"`
+
+	// Optional message to describe the status reason
+	Message string `json:"message,omitempty"`
+
+	SetTime timestamp.Timestamp `json:"set-time"`
+}
+
+// Represents the source of a course status, which is either a user or automated process.
```
eriq-augustine, line 21: "Source" is ambiguous here, and you use is semi-recursively in the definition.
Maybe "initiator" would be better here.

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+type StatusSource int
+
+// StatusSourceUnknown is the zero value and is not a valid source.
+// StatusSourceCourse is for users with admin/owner privileges in the course.
+// StatusSourceServer is for users with admin/owner privileges in the server.
+// StatusSourceAutomated is for automated processes.
+// StatusSourceRoot is for users with root privileges.
```
eriq-augustine, line 28: For comments like this, go ahead and comment each constant specifically, e.g.,:
```
const (
    // The zero ...
	StatusSourceUnknown   StatusSource = 0

    // For users ...
	StatusSourceCourse                 = 10
)
```

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+	StatusSourceRoot:      "root",
+}
+
+var stringToStatusSource = map[string]StatusSource{
+	"unknown":   StatusSourceUnknown,
+	"course":    StatusSourceCourse,
+	"server":    StatusSourceServer,
+	"automated": StatusSourceAutomated,
+	"root":      StatusSourceRoot,
+}
```
eriq-augustine, line 51: I know you may have seen this pattern in our code somewhere,
but I think it would be great here if we instead had some map reversing function.
Then we wouldn't have to worry about it getting out of sync.
Maybe we could have something like:
```
var stringToStatusSource map[string]StatusSource = MapReverse(statusSourceToString)
```

Take a look and see if we can make it work with generics.

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+	if err == nil {
+		*this = *value
+	}
+
+	return err
+}
+
+// Checks if a status has priority over another status, which is determined through the StatusSource hierarchy.
+// Ties are broken by the SetTime, giving priority to the most recently set status.
+func (this *CourseStatus) hasPriorityOver(other *CourseStatus) bool {
```
eriq-augustine, line 72: I would prefer this method be a more standard [three way comparison](https://en.wikipedia.org/wiki/Three-way_comparison) function (negative, zero, positive).

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+	if this.Source != other.Source {
+		return this.Source > other.Source
+	}
```
eriq-augustine, line 75: Put a newline after this conditional to visually separate out the cases (compare with source and compare with time).
(This applies to other control structures like loops.)

internal/model/course_status.go
```diff
@@ -0,0 +1,88 @@
...
+	return this.SetTime > other.SetTime
+}
+
+// Iterates through every status to determine which should actually be used
+func determineStatus(statuses map[string]*CourseStatus) *CourseStatus {
+	var best *CourseStatus
+	for _, status := range statuses {
+		if best == nil || status.hasPriorityOver(best) {
```
eriq-augustine, line 83: In this project, we prefer to put parens around binary logical operands.
So here:
```
if (best == nil) || status.hasPriorityOver(best) {
```

eriq-augustine (commented):
(Sorry, accidentally sent this too early.
Combine these comments with the next batch.)


Note that we will typically comment on things only once,
but they may apply to the whole PR,
so make sure to check for multiple instances of a comment.

internal/model/course_status_test.go
```diff
@@ -0,0 +1,42 @@
+package model
+
+import (
+	"testing"
+
+	"github.com/edulinq/autograder/internal/timestamp"
+)
+
+func TestDetermineStatus(test *testing.T) {
```
eriq-augustine, line 9: You need more test cases here.

I would turn this into a "test case"-style test,
so you can quickly validate more cases, e.g., partial ties, full ties, other owners.

internal/tasks/core.go
```diff
@@ -86,6 +86,24 @@ func runNextTask() {
...
+			log.Error("Failed to get course for scheduled task.", task, err)
+			return
+		}
+
+		if (course != nil) && !course.IsActive(startTimestamp) {
+			task.AdvanceRunTimes()
+			err = db.UpsertActiveTask(task)
+			if err != nil {
+				log.Error("Failed to save task.", err)
+			}
```
eriq-augustine, line 102: I don't this that these tasks should get marked as run,
when they are not run.

internal/api/courses/status/remove.go
```diff
@@ -0,0 +1,79 @@
...
+	// Optional log message to include with status removal.
+	Message string `json:"message"`
+}
+
+type RemoveResponse struct {
+	Removed []string `json:"removed"`
+}
+
+func HandleRemove(request *RemoveRequest) (*RemoveResponse, *core.APIError) {
+
```
eriq-augustine, line 29: Extra newline.

internal/api/courses/status/remove.go
```diff
@@ -0,0 +1,79 @@
...
+	// Optional log message to include with status removal.
+	Message string `json:"message"`
+}
+
+type RemoveResponse struct {
+	Removed []string `json:"removed"`
+}
+
+func HandleRemove(request *RemoveRequest) (*RemoveResponse, *core.APIError) {
```
eriq-augustine, line 28: Missing comment.
Note that comments on the API endpoint handlers get automatically added to the API documentation (try it out).

internal/api/courses/status/remove.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+
+			deletionMap[request.TargetOwner] = nil
+		}
+	} else {
+		_, ok := courseStatuses[request.ServerUser.Email]
+		if ok {
+			deletionMap[request.ServerUser.Email] = nil
+		}
+	}
```
eriq-augustine, line 55: These seems like the same case to me.
If you set TargetOwner earlier, then are these not the same?

internal/api/courses/status/remove.go
```diff
@@ -0,0 +1,79 @@
...
+
+	if len(deletionMap) > 0 {
+		err := db.UpsertCourseStatuses(request.Course, deletionMap)
+		if err != nil {
+			return nil, core.NewInternalError("-649", request, "Failed to upsert status.").Err(err)
+		}
+	}
+
+	removed := []string{}
+
```
eriq-augustine, line 65: I would remove this newline.
Then, you have a single visual block that is "collect the removed entries".

internal/api/courses/status/remove.go
```diff
@@ -0,0 +1,79 @@
...
+	// If true, remove all statuses the caller has permission to remove.
+	Clear bool `json:"clear"`
+
+	// Optional log message to include with status removal.
+	Message string `json:"message"`
+}
+
+type RemoveResponse struct {
+	Removed []string `json:"removed"`
```
eriq-augustine, line 25: What about returning a `map[string]model.CourseStatus`?

internal/api/courses/status/remove_test.go
```diff
@@ -0,0 +1,167 @@
...
+			Active:  false,
+			Source:  model.StatusSourceServer,
+			Owner:   "server-owner@test.edulinq.org",
+			SetTime: timestamp.Now(),
+		},
+	}
+
+	db.MustUpsertCourseStatuses(db.MustGetCourse("course101"), statuses)
+
+	testCases := []struct {
```
eriq-augustine, line 47: I think it would make your test cases more clear if you put the statuses in the test cases themselves.
Then, you also have the flexibility to change the statuses on a per-case basis.

internal/api/courses/status/remove_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+		if !slices.Equal(removed, testCase.expected) {
+			test.Errorf("Case %d: Expected '%v', found '%v'.", i, testCase.expected, removed)
+		}
+	}
+
+	finalStatuses, err := db.GetCourseStatuses(db.MustGetCourse("course101"))
+	if err != nil {
+		test.Fatalf("Failed to get final statuses: '%v'.", err)
+	}
```
eriq-augustine, line 163: NEVER (99.95% of the time)
have two different conditionals right next to eachother.
It is soooo easy for a reader to accidentally see the second conditional as an `else if`.

internal/api/courses/status/set.go
```diff
@@ -0,0 +1,73 @@
...
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleAdmin
+
+	// Indicates whether course should be active or inactive.
+	Active bool `json:"active"`
+
+	// Optional message to include with status change.
+	Message string `json:"message"`
+
+	// If status already exists for user, allows overwrite for existing status.
```
eriq-augustine, line 20: s/user/the user/

internal/api/courses/status/set.go
```diff
@@ -0,0 +1,73 @@
...
+	Status *model.CourseStatus `json:"status"`
+}
+
+func HandleSet(request *SetRequest) (*SetResponse, *core.APIError) {
+
+	owner := request.ServerUser.Email
+
+	_, ok := request.Course.Statuses[owner]
+	if ok && !request.Force {
+		return nil, core.NewBadRequestError("-646", request, "Existing status for owner (must force to overwrite).")
```
eriq-augustine, line 34: I think the message needs a little work.
Note that the status owner can only be the caller,
so I don't think you need the "for owner" part.
Consider also returning some information about the existing status in this message.

internal/api/courses/status/set.go
```diff
@@ -0,0 +1,73 @@
...
+		request.ServerUser,
+		log.NewAttr("message", request.Message),
+		log.NewAttr("active", request.Active),
+		log.NewAttr("source", status.Source),
+	)
+
+	return &SetResponse{Status: status}, nil
+}
+
+// Finds the highest StatusSource using the user's server role
```
eriq-augustine, line 62: Note in this comment that the caller must be at least and admin because of the request's permissions.

internal/api/courses/status/set_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/edulinq/autograder/internal/db"
+	"github.com/edulinq/autograder/internal/model"
+	"github.com/edulinq/autograder/internal/util"
+)
+
+func TestSet(test *testing.T) {
+	db.ResetForTesting()
+	defer db.ResetForTesting()
+
+	testCases := []struct {
```
eriq-augustine, line 16: Having state-full test cases gets too confusing.
Reset the DB for each test and have each test case specify any statuses that should exist.

internal/model/course.go
```diff
@@ -26,10 +27,15 @@ type Course struct {
...
+	// Course Active Window
+	StartDate *timestamp.Timestamp `json:"start-date,omitempty"`
+	EndDate   *timestamp.Timestamp `json:"end-date,omitempty"`
+
 	Tasks []*UserTaskInfo `json:"tasks,omitempty"`
 
 	// Internal fields the autograder will set.
-	Assignments map[string]*Assignment `json:"-"`
+	Assignments map[string]*Assignment   `json:"-"`
+	Statuses    map[string]*CourseStatus `json:"-"`
```
eriq-augustine, line 38: I disagree with this modeling.

If we let the statuses stay with the course,
then it makes the DB work WAY simpler.

eriq-augustine (changes requested):
Looking pretty good.

Note that I messed up and let a batch of comments through before I was done reviewing.

The only core part I disagree with is the data modeling around courses and statuses.
I think we should keep statuses with courses, then the DB work becomes much simpler.
I can see arguments about how we don't want to load statuses from users (i.e., when a course is loaded from source),
but I think it would be better to handle that case than make our DB work more complicated.
(Of course, we can discuss this synchronously.)

internal/tasks/core.go
pshayes06: I could implement logic inside SaveCourse to remove or repopulate active tasks whenever the active state of a course switches. However, there is no trigger/call whenever the active state passively change via the date window. I could change this check in tasks/core.go to remove all active tasks for the course, handling courses that have gone inactive via date window. On the other hand, if a course becomes active through its window, it will have no tasks to process since they will have been removed. Unless we create some other way to automatically check and handle this case, I believe a user would have to manually repopulate the active course tasks. Do you have a suggestion on handling this case, or should I stick with just skipping over tasks?

eriq-augustine: In the future, we can just create a job to to maintenance on courses, including checking active status (based on timestamps).
So, do your work with the assumption that this will exist in the future.

See #172.

internal/model/course.go
```diff
@@ -109,6 +116,17 @@ func (this *Course) Validate() error {
 		}
 	}
 
+	if (this.StartDate != nil) && (this.EndDate != nil) {
+		if *this.EndDate < *this.StartDate {
+			return fmt.Errorf("Course end date is before start date.")
+		}
+	}
+
+	// Creating an empty map for safe write operations.
```
eriq-augustine: We usually use a more imperitive tone with comments, e.g.,
"Create an empty map for safe write operations.".

internal/model/course.go
```diff
@@ -248,3 +266,38 @@ func (this *Course) GetTemplatesDir() string {
...
+	for _, status := range this.Statuses {
+		if (best == nil) || (status.compareTo(best) > 0) {
+			best = status
+		}
+	}
+
+	return best
+}
+
+// Returns whether a course is active at a specific time, defaulting to true.
```
eriq-augustine: I woujld say that there is no "default" case here.
Yes, you return true at the very end of your function,
but you check all possible conditions first (so it is not a "default").

internal/model/course.go
```diff
@@ -248,3 +266,38 @@ func (this *Course) GetTemplatesDir() string {
...
+// If there are no statuses, the date window is checked.
+func (this *Course) IsActive(time timestamp.Timestamp) bool {
+	highestStatus := this.GetActiveStatus()
```
eriq-augustine: s/highestStatus/activeStatus/

This will give you symmetry with your GetActiveStatus() method.

internal/model/course_status.go
```diff
@@ -0,0 +1,89 @@
...
+	// For users with admin/owner privileges in the course.
+	StatusSourceCourse = 10
+
+	// For users with admin/owner privileges in the server.
+	StatusSourceServer = 20
+
+	// For automated processes.
+	StatusSourceAutomated = 30
+
+	// For users with root privileges.
```
eriq-augustine: There is only one root user, and no users are allowed to have that role.
So this comment should just be something like: "For the root user."

internal/model/course_status.go
```diff
@@ -0,0 +1,89 @@
...
+
+var statusSourceToString = map[StatusSource]string{
+	StatusSourceUnknown:   "unknown",
+	StatusSourceCourse:    "course",
+	StatusSourceServer:    "server",
+	StatusSourceAutomated: "automated",
+	StatusSourceRoot:      "root",
+}
+
+var stringToStatusSource = util.MapReverse(statusSourceToString)
```
eriq-augustine: Define the type here just to make it extra clear to readers.

internal/model/course_status.go
```diff
@@ -0,0 +1,89 @@
...
+	value, err := util.UnmarshalEnum(data, stringToStatusSource, true)
+	if err == nil {
+		*this = *value
+	}
+
+	return err
+}
+
+// Checks if a status has priority over another status, which is determined through the StatusSource hierarchy.
+// Ties are awarded to the most recent SetTime, and if those are equal, the unique owner string are compared lexicographically.
```
eriq-augustine: Note here that "larger" or "later" status have priority over earlier ones.

eriq-augustine: This sentence needs a little more consistency.
Note that you  use a field name (`SetTime`), but then also a description of another field ("the unique owner string").
We generally want to be consistent, and in this case we usually don't want to use field names (so a reader can understand it without needing to refer back to the struct).

internal/model/course_status.go
```diff
@@ -0,0 +1,89 @@
...
+func (this *CourseStatus) compareTo(other *CourseStatus) int {
+	if this.Source != other.Source {
+		if this.Source > other.Source {
+			return 1
+		}
```
eriq-augustine, line 78: Add a newline after this conditional.
We want to separate out these "visual blocks" into cases.
Case 1: This source is larger than other.
Case 2: This source is smaller than other.

internal/model/course_status_test.go
```diff
@@ -0,0 +1,124 @@
...
+				"server-admin@test.edulinq.org": {
+					Source:  StatusSourceServer,
+					Owner:   "server-admin@test.edulinq.org",
+					SetTime: timestamp.FromMSecs(100),
+				},
+			},
+			"server-admin@test.edulinq.org",
+		},
+
+		// Tie on StatusSource and SetTime (picks owner string that comes latest in lexicographical order).
```
eriq-augustine: s/latest/last/

internal/procedures/courses/upsert_test.go
```diff
@@ -336,6 +336,39 @@ func TestUpsertBase(test *testing.T) {
...
 }
 
+func TestUpsertPreserveStatuses(test *testing.T) {
+	db.ResetForTesting()
+	defer db.ResetForTesting()
+
+	// Save the initial status.
+	course := db.MustGetCourse("course101")
+	course.Statuses = map[string]*model.CourseStatus{
+		"fake-user@test.edulinq.org": {},
```
eriq-augustine: I would put in an actual status here.
It would be fairly reasonable for the DB backend to not return empty values.

internal/util/map.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+// Returns a new map where the original keys and values are swapped.
+// For map values that are not unique, only one pair using that value as a key will exist.
+func MapReverse[K comparable, V comparable](original map[K]V) map[V]K {
+	rev := make(map[V]K, len(original))
+	for k, v := range original {
```
eriq-augustine: s/k/key/
s/v/value/

We generally shy away from one letter variable names (`i` (and sometimes `j`) being the exceptions).
It is fine to keep your generic type variables as one letter, since that is a strong convention.

internal/api/courses/status/remove.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+type RemoveRequest struct {
+	core.APIRequestCourseUserContext
+	core.MinCourseRoleAdmin
+
+	// Email of status owner to remove. Defaults to the caller.
+	TargetOwner string `json:"target-owner"`
+
+	// If true, remove all statuses the caller has permission to remove.
```
eriq-augustine: Note that if this is true, the target owner is ignored.

internal/api/courses/status/remove.go
```diff
@@ -0,0 +1,73 @@
...
+	} else {
+		target := request.TargetOwner
+		if target == "" {
+			target = request.ServerUser.Email
+		}
+
+		targetStatus, ok := courseStatuses[target]
+		if ok {
+			if callerSource < targetStatus.Source {
+				return nil, core.NewPermissionsError("-648", request, targetStatus.Source, callerSource, "Cannot remove a status with a higher source.")
```
eriq-augustine: I think we should use `NewBadRequestError` instead.
This is a case where both work, but the difference comes does to visibility.
Permission errors are more secretive (so as to not leak sensitive information),
but bad request errors allow you to pass a message to the user.
(Note that we have already decided that it is fine to show a course's status to anyone that can get the course's information.)

And since this error now goes to the user, try to make it a little more clear to them.
For example, a caller probably won't know what a "status" is.

internal/api/courses/status/remove_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	testCases := []struct {
+		email            string
+		target           string
+		clear            bool
+		existingStatuses map[string]*model.CourseStatus
+		locator          string
+		expectedOwners   []string
+	}{
+		// Basic Deletion
```
eriq-augustine: s/Basic/Self/

internal/api/courses/status/set.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Set the status of a course, where each user can have at most one status.
+func HandleSet(request *SetRequest) (*SetResponse, *core.APIError) {
+	owner := request.ServerUser.Email
+
+	existingStatus, ok := request.Course.Statuses[owner]
+	if ok && !request.Force {
+		return nil, core.NewBadRequestError("-646", request,
+			fmt.Sprintf("Incoming status must be forced to overwrite existing (set at %s).",
+				existingStatus.SetTime.SafeString()))
```
eriq-augustine: This message may be a bit confusing for the caller (remember that they may not know about all the options off the top of their head).
Maybe something like:
"Course status already exists for this user (set at %s), use the force option to overwrite."

internal/db/course.go
```diff
@@ -123,18 +123,21 @@ func upsertActiveCourseTasks(course *model.Course) error {
...
-		}
-
-		if newTask == nil {
-			continue
+	// An inactive course has all its tasks removed, only repopulated once active again.
+	if course.IsActiveNow() {
+		for i, task := range course.Tasks {
+			newTask, err := task.ToFullCourseTask(course.GetID())
+			if err != nil {
+				return fmt.Errorf("Unable to upsert task at inded %d: '%w'.", i, err)
```
eriq-augustine: s/inded/index/

(I know this was not your typo.)

internal/db/task_test.go
```diff
@@ -183,6 +183,65 @@ func (this *DBTests) DBTestGetNextActiveTaskBase(test *testing.T) {
...
+		test.Fatalf("Failed to fetch case 1 tasks: '%v'.", err)
+	}
+
+	if len(tasks) != 1 {
+		test.Fatalf("Did not get expected number of case 1 tasks. Expected: %d, Actual: %d.", 1, len(tasks))
+	}
+
+	// Set status to inactive, removing all active tasks for the course.
+	course = MustGetTestCourse()
+	course.Tasks = []*model.UserTaskInfo{testTask}
```
eriq-augustine: Do you need this set here (and in the next case)?
Shouldn't the course already have the task, but it is just not active.

eriq-augustine (changes requested):
This is looking great!
Your tests are looking real good.

internal/api/courses/status/remove.go
```diff
@@ -14,7 +14,7 @@ type RemoveRequest struct {
 	// Email of status owner to remove. Defaults to the caller.
 	TargetOwner string `json:"target-owner"`
 
-	// If true, remove all statuses the caller has permission to remove.
+	// If true, ignores TargetOwner and removes all statuses the caller has permission to remove.
```
eriq-augustine: Similar to a comment in the previous pass,
try to not reference field names when you don't need to.
This case is a little different because we are in the same struct,
but these comments make it into the API documentation (see api.json in your PR),
which use the JSON names and not the field names.
So if you do something like s/TargetOwner/the target owner/ ,
you can satisfy both.

internal/api/courses/status/remove.go
```diff
@@ -48,7 +48,8 @@ func HandleRemove(request *RemoveRequest) (*RemoveResponse, *core.APIError) {
 		targetStatus, ok := courseStatuses[target]
 		if ok {
 			if callerSource < targetStatus.Source {
-				return nil, core.NewPermissionsError("-648", request, targetStatus.Source, callerSource, "Cannot remove a status with a higher source.")
+				return nil, core.NewBadRequestError("-648", request,
+					"Cannot remove this course's active/inactive status because it was set by a higher privileged source.")
```
eriq-augustine: Include the actual source in the message.
Like I mentioned in the previous comment, the caller already has access to the status information,
so it's fine if we give them additional details.
Remember that this project is open source,
so a static message provides about the same information as a NewPermissionsError (which we wanted to not use in this case).
(The caller could always just look into the source code.)

internal/model/course_status.go
```diff
@@ -67,13 +67,14 @@ func (this *StatusSource) UnmarshalJSON(data []byte) error {
 	return err
 }
 
-// Checks if a status has priority over another status, which is determined through the StatusSource hierarchy.
-// Ties are awarded to the most recent SetTime, and if those are equal, the unique owner string are compared lexicographically.
+// Checks if a status has priority over another status, determined by the one with a larger source value.
+// Statuses set later have priority over earlier ones, and if the time is equal, the unique owner strings are compared lexicographically.
```
eriq-augustine: This comment is off (and got further away since the last pass).

This is a standard comparison function first and foremost,
as indicated by the name.

Try something like this:
> Compare two status using status priority ordering.
> Status priority ordering is determined by ...
> Status sorted later (having a positive return) have higher priority.

eriq-augustine (changes requested):
Just some small comment things.
(Only looked at delta this pass.)

eriq-augustine (approved):
LGTM!

Same as before, now we are ready to get in the Python interface endpoints for this.
Like we do with a lot of our admin-level endpoints,
it should be fine to display pretty (indented) JSON as the stdout output.

Note that you overindexed a bit when not using field names in comments.
The general rule is that comments that could get presented to an outside user (not a developer) should try not to use field names.
So function and field comments (which automatically get included in go-style documentation) and API-level comments (which our project scoops up specially for our documentation) are the main places to be careful.

